PR #13, opened 2019-04-02
TunedRF: lower #trees during tuning and tune with ascending mtry

seb-h2o:
@PGijsbers please have a check and tell me if you agree with those changes.
I tried it with some of the small and medium datasets.
Haven't seen any significant change in picking `max_features` for  small ones, but it allows to build more models with different `mtry` in case of larger ones.


PGijsbers:
Thanks 👍

PR #14, opened 2019-04-10
Autosklearn experiments [not for merge]

seb-h2o:


frameworks/autosklearn/exec.py
```diff
@@ -4,6 +4,9 @@
 import warnings
 
 os.environ['JOBLIB_TEMP_FOLDER'] = tmp.gettempdir()
+os.environ['OMP_NUM_THREADS'] = '1'
+os.environ['OPENBLAS_NUM_THREADS'] = '1'
+os.environ['MKL_NUM_THREADS'] = '1'
```
seb-h2o: adding those environment variables to prevent numpy from using multiple threads per process

seb-h2o: will need to apply this for TunedRF and TPOT as well

seb-h2o: https://github.com/joblib/threadpoolctl provide a generic fix for those configs, but it's too recent (not available in pypi yet)

PR #19, opened 2019-06-11
updated scipy version constrant in requirements.txt

nandini269:
Updated constraint to avoid error with arff library. Scipy 1.3 was not using the liac-arff library as required by the code.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,6 @@ numpy>=1.15,<2.0
 pandas>=0.23,<1.0
 psutil>=5.4,<6.0
 ruamel.yaml>=0.15,<1.0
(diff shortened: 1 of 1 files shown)


PGijsbers:
Sorry to revive this topic, but I don't exactly understand what you mean by this. Could you elaborate? (looping @sebhrusen in case he knows)

sebhrusen:
@PGijsbers I don't know either, I don't even know why we have `scipy` in the requirements as the app is not using it directly so it should just be a transitive dependency.
So, imo, the best way to uprade `scipy` is to remove it from `requirements.txt` and let `numpy`, `openml` or whatever resolve their required version...

Issue #42, opened 2019-07-17
Add support for AutoXGBoost


sebhrusen:
We can now add @ja-thomas's toolkit to our framework list:
https://github.com/ja-thomas/autoxgboost

and run existing benchmarks against it once integration is stable.

PR #50, opened 2019-08-19
New Framework: autoxgboost

Coorsaa:
see #42 .


PGijsbers:
To clarify; I only had a look at the code. I am having some difficulties running the benchmark right now, so I can't test it yet which is why it isn't merged.

Coorsaa:
```
python3 runbenchmark.py autoxgboost test -m docker -Xdocker.force_branch=False
```
worked for me in a virtual environment. Could you provide some info on the problems you're facing?

PGijsbers:
My problems are not specific to the autoxgboost PR, but due to a fresh windows install (and windows not officially being supported). Sorry for the delay.

frameworks/autoxgboost/setup.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+HERE=$(dirname "$0")
+. $HERE/../shared/setup.sh
+if [[ -x "$(command -v apt-get)" ]]; then
+    SUDO apt-get update
+    SUDO apt-get install -y software-properties-common apt-transport-https libxml2-dev
+    SUDO apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9
```
seb-h2o: looks like this is importing the wrong public key:
```
root@a5753493cd7e:/bench# apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9
Executing: /tmp/apt-key-gpghome.uiTd5mx21z/gpg.1.sh --keyserver keyserver.ubuntu.com --recv-keys E084DAB9
gpg: key 4359ED62E084DAB9: public key "Totally Legit Signing Key <mallory@example.org>" imported
gpg: Total number processed: 1
gpg:               imported: 1
```
this causes later:
```
W: GPG error: https://cran.rstudio.com/bin/linux/ubuntu bionic-cran35/ InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 51716619E084DAB9
E: The repository 'https://cran.rstudio.com/bin/linux/ubuntu bionic-cran35/ InRelease' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

seb-h2o: probably a conflict between 2 keys with same suffix.
importing the key explicitly works better apparently:
```
root@a5753493cd7e:/bench# apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 51716619E084DAB9
Executing: /tmp/apt-key-gpghome.L0VQc8DfC6/gpg.1.sh --keyserver keyserver.ubuntu.com --recv-keys 51716619E084DAB9
gpg: key 51716619E084DAB9: public key "Michael Rutter <marutter@gmail.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1
```

seb-h2o: @Coorsaa please change the key in your PR.
We must have the same issue with one or two other framework integrations, I'll take care of those.

seb-h2o: after this I have issues installing `r-cran-rjava`: see https://stackoverflow.com/questions/51256462/r-cannot-install-rjava-what-is-r-api-3-4

do you mind taking a look?

seb-h2o:
@Coorsaa I directly updated your `setup.sh`. With the changes in https://github.com/openml/automlbenchmark/pull/50/commits/815883c8efac9e94044a3368f39770bb6ca1b6b8, I could build the docker image and run the test dataset.

Will run validation set and confirm if everything ok: I'm still unsure about the predictions format for some corner case datasets, as this is usually handled by python `results` utility module.

seb-h2o:
@Coorsaa feel free to rebase on top of `master` if you want to retry building a docker image.
This integration convinced me to implement live output display when building the image as we were missing a lot of information there when building it through the Py command: cf. https://github.com/openml/automlbenchmark/pull/52

frameworks/autoxgboost/exec.py
```diff
@@ -0,0 +1,29 @@
...
+    here = dir_of(__file__)
+    run_cmd(r"""Rscript --vanilla -e "source('{script}'); run('{train}', '{test}', '{output}', {cores}, time.budget = {time_budget})" """.format(
+      script = os.path.join(here, 'exec.R'),
+      train = dataset.train.path,
+      test = dataset.test.path,
+      output = config.output_predictions_file,
+      cores = config.cores,
+      time_budget = config.max_runtime_seconds
+    ))
+
```
seb-h2o: small improvement: I understand that it's not simple to obtain the number of models trained from R, but can you please at least return `training_duration`.
See AutoWEKA integration for an example on how it can be easily done, basically:
```
    from automl.utils import Timer
    ...
    with Timer() as training:
        run_cmd(...)

    return dict(
        training_duration=training.duration
    )
```

seb-h2o:
@Coorsaa After extending it (https://github.com/openml/automlbenchmark/pull/54), I ran the validation benchmark for a few minutes for each task on one fold only.
The first 5 tasks ran perfectly.
But the last one (`APSFailure`) has a good name indeed:
```
Error in makeClassifTask(data = train, target = target) :
  Assertion on 'eg_000' failed: Must be of type 'factor', not 'double'.
Calls: run ... makeClassifTask -> assertFactor -> makeAssertion -> mstop
Execution halted

Command 'Rscript --vanilla -e "source('/bench/frameworks/autoxgboost/exec.R'); run('/input/org/openml/www/datasets/41138/dataset_train_0.arff', '/input/org/openml/www/datasets/41138/dataset_test_0.arff', '/output/docker_autoxgboost_validation_20190902T210539/predictions/autoxgboost_APSFailure_0.csv', 2, time.budget = 3600)" ' returned non-zero exit status 1.
  File "/bench/automl/benchmark.py", line 371, in run
    meta_result = framework.run(self._dataset, task_config)
  File "/bench/frameworks/autoxgboost/__init__.py", line 10, in run
    return run(*args, **kwargs)
  File "/bench/frameworks/autoxgboost/exec.py", line 26, in run
    time_budget = config.max_runtime_seconds
  File "/bench/automl/utils.py", line 698, in run_cmd
    raise e
  File "/bench/automl/utils.py", line 685, in run_cmd
    universal_newlines=True)
  File "/bench/automl/utils.py", line 631, in run_subprocess
    raise subprocess.CalledProcessError(retcode, process.args, output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command 'Rscript --vanilla -e "source('/bench/frameworks/autoxgboost/exec.R'); run('/input/org/openml/www/datasets/41138/dataset_train_0.arff', '/input/org/openml/www/datasets/41138/dataset_test_0.arff', '/output/docker_autoxgboost_validation_20190902T210539/predictions/autoxgboost_APSFailure_0.csv', 2, time.budget = 3600)" ' returned non-zero exit status 1.
```

the `exec.R` makes the assumption that the target is the last column, but it's not the case for all datasets.
However, in `exec.Py`, you can easily obtain the target col index using `dataset.target.index`: you would need to pass this info to `exec.R` then.

frameworks/autoxgboost/exec.R
```diff
@@ -0,0 +1,31 @@
+library(mlr)
+library(autoxgboost)
+library(farff)
+
+run = function(train_file, test_file, output_predictions_file, cores, time.budget) {
+  train = farff::readARFF(train_file)
+  colnames(train) = make.names(colnames(train))
+  target = colnames(train)[ncol(train)]
```
seb-h2o: as explained in https://github.com/openml/automlbenchmark/pull/50#issuecomment-527243835
you need to find a way to use `dataset.target.index` here instead of `ncol(train)`.

seb-h2o: @Coorsaa any update on this?

Coorsaa:
Thanks for the great feedback. Working on it right now.

PR #52, opened 2019-09-02
Live display for some subprocesses when using `run_cmd`

seb-h2o:
adding live console display for long-running commands when using `run_cmd`.
The main utility is when building docker images as we're blind otherwise and can't see issues that may block docker.


seb-h2o:
tested, works like a charm now when building docker images: very useful when setting up R libraries during `setup.sh` phase as a lot of compilation errors may occur then.

Issue #17, opened 2019-05-03
Decouple resources from tasks in the benchmark


PGijsbers:
I think coupling the resources and tasks into a single file is not optimal.
This leads to a lot of duplicate benchmarks where only the resources specifications differ (see our current resources/benchmarks folder).
At the very least I think we should have separate files specifying resources and collections of tasks.
Perhaps we should only have files specifying the resources (if at all, since that would be easy to do from the command line too), and use [OpenML Studies](https://docs.openml.org/#studies-under-construction) to specify tasks.

The upsides would be:
 - easier to add a dataset: modify only one file if not using OpenML study, or none if we are.
 - rerunning a set of tasks with different resources would not require a new file.

Downsides:
 - no automatic public history of tasks added (if through an OpenML study).
 - not possibly to add arbitrary evaluation metrics (not possible in the current setup either, but would be easier to add as it would not rely on an OpenML implementation for it). I wouldn't be sure how to, if ever, allow arbitrary evaluation metrics on OpenML (as you would need to be able to run them in any of the supported languages).

@joaquinvanschoren are there any plans for something like versions for studies? That could negate one of the downsides.

seb-h2o:
Closing this issue, created https://github.com/openml/automlbenchmark/issues/61 for the OpenML studies part.

PR #57, opened 2019-10-09
Decoupling benchmark constraints (folds, max_runtime, resources...) from the tasks definitions

seb-h2o:
cf. https://github.com/openml/automlbenchmark/issues/17

- removed the `benchmarks/medium-4c8h.yaml` files and the likes.
- tasks are now only defined in `benchmarks/small.yaml`, `medium.yaml`, `large.yaml` without any redundancies in those files.
- some predefined constraint definitions are available by default in `resources/constraints.yaml`, for example:
```
4h8c:
  folds: 10
  max_runtime_seconds: 14400
  cores: 8
  min_vol_size_mb: 65536
```
- metrics are defined by default as global config and are applied depending on the problem type (the first one is used for ranking). See in `config.yaml`:
```
benchmarks:
  metrics:    # default metrics by dataset type (as listed by amlb.data.DatasetType)
    binary: ['auc', 'acc']
    multiclass: ['logloss', 'acc']
    regression: ['rmse', 'r2']
```
- user specifies the constraint config using the command (the default constraint definition is `test`):
```python runbenchmark.py {framework} {benchmark} {constraint}```
```python runbenchmark.py autosklearn medium 4h8c```
- legacy way of specifying constraints in the benchmark file is still available, but should become useless (may be removed later).
- support for `OpenML Studies` is a next step: won't be done in this PR though. It probably makes more sense to have a separate script converting the `OpenML study` to our `yaml` format: should be fairly simple.

### Note
this PR also includes changes in the `reports` notebook which extracts selected dataset metadata from OpenML, exports them into a csv file accessible to everyone on GitHub, and allows sorting of datasets according to those metadata in current visualizations.

### Tests
- tested changes in local, docker and AWS modes.
- updated README.md


PGijsbers:
For some reason I can't see the file changes right now. So I assume that all checks out. But I agree with the changes as described in the PR comment. :+1: Nice step forward

PR #65, opened 2019-11-07
Gama

### LINKED ISSUE #63: Add support for GAMA
New AutoML framework: https://pgijsbers.github.io/gama/

PGijsbers:
Fixes #63

Added GAMA as AutoML framework. Looking for feedback on the integration. I have to perform more tests myself. Currently only tested locally on the test benchmark with 4 cores and 1 minute of runtime.


frameworks/GAMA/exec.py
```diff
@@ -0,0 +1,82 @@
...
+    scoring_metric = metrics_mapping[config.metric] if config.metric in metrics_mapping else None
+    if scoring_metric is None:
+        raise ValueError("Performance metric {} not supported.".format(config.metric))
+
+    training_params = {k: v for k, v in config.framework_params.items() if not k.startswith('_')}
+    n_jobs = config.framework_params.get('_n_jobs', config.cores)  # useful to disable multicore, regardless of the dataset config
+
+    dir, file = os.path.split(config.output_predictions_file)
+    file_name, file_ext = os.path.splitext(file)
+    log_file = os.path.join(dir, file_name + '.gamalog')
```
seb-h2o: This pattern is used multiple times in integration code each time we need to generate predictions in different format, for logs or whatever.
Wondering if I should add a property like `config.output_file_pattern` or sth similar:
```log_file = config.output_file_pattern.format(ext="gamalog")```

Note that the output dir is already directly available as `config.output_dir` otherwise.

File: frameworks/GAMA/setup.sh [hunk 2]
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+shopt -s expand_aliases
```
seb-h2o: not sure this is still useful, not using aliases anymore (they're not reliable)

frameworks/GAMA/exec.py
PGijsbers: I missed the `config.output_dir` and will use that instead.

frameworks/GAMA/setup.sh
PGijsbers: Will remove, I had this copied over from a the RandomForest `setup.sh`

PGijsbers:
I don't know why Github decided my replies were a review :D

PGijsbers:
It ran to completion on all (7) validation tasks under a 5 minute time budget (one fold per task).

Just a small note, I have yet to pick good defaults for search strategy (evolution vs successive halving) and post-processing (mostly how many pipelines to include in the ensemble). The version of GAMA on PyPI currently uses evolution and no post-processing (it does not build an ensemble, but fits the best found pipeline).

Finding good default behavior is high on my list, but before then I don't know that I would recommend running it extensively (e.g. across all tasks with 10-fold 8 hour runs).

PGijsbers:
Hi Seb, sorry to make this PR go stale. I think it's in good shape for a merge, but would appreciate it if you could do some validation on aws since I don't have the credits to test that.

sebhrusen:
@PGijsbers I'll build a docker image + run validation benchmark in the next few days.

Note, you should be able to run it on your private AWS account with a free or very cheap image.
I can provide you with a local `config.yaml` for this if you want.
Actually, it's a good occasion: I realize that the doc doesn't clearly specify how to do this.

PGijsbers:
I _should_ be able to figure that out. But I agree it's a good opportunity to test this. If you can provide the config file and instructions I'll follow it step by step and troubleshoot if necessary.

PR #68, opened 2019-12-11
Provide the ability to extract artifacts from the benchmark runs

seb-h2o:
Provide the ability to extract artifacts from the benchmark runs.

Here, artifacts mean data, files generated by the tested framework that can help in the interpretation or investigation of results. They are specific to each framework and can't easily be processed automatically across all of them.

Example of such artifacts could be logs, generated pipeline/models...

Documentation available at https://github.com/openml/automlbenchmark/blob/artifacts/docs/HOWTO.md#extract-more-information


seb-h2o:
@PGijsbers This is a branch that I created months ago to investigate results in depth, especially for H2O, but to compare generated models with `autosklearn` and `TPOT`.
I just rebased it now so that it's up-to-date.

At user's level, it just exposes a framework param where user can list the artifacts he wants to be extracted during the benchmark.
For example:
```
autosklearn:
  version: '0.5.1'
  project: https://automl.github.io/auto-sklearn/stable/
  params:
    _save_artifacts: ['models']

H2OAutoML:
  version: '3.24.0.1'
  project: http://docs.h2o.ai/h2o/latest-stable/h2o-docs/automl.html
  params:
    _save_artifacts: ['leaderboard', 'logs', 'models']
```

Just tell me if you feel like this could/should go to main branch. 
Otherwise, I'll leave it where it is.

PGijsbers:
I think it's very useful to be able to save the logs of AutoML frameworks easily. I would ask to add some documentation as to how the system works (e.g. how the AutoML framework script knows which artifacts to store, and where to store them). It can be seen from referencing auto-sklearn/h2o code, but I think it's probably better captured also in a paragraph or two of documentation.

Issue #75, opened 2020-02-27
Object 'Benchmark' has no attribute 'SetupMode'


Rong-Inspur:
Hi,

I run h2oautoml in docker mode
`python3 runbenchmark.py h2oautoml test -m docker`
and get the following error
```
Running `h2oautoml` on `test` benchmarks in `docker` mode.
Loading frameworks definitions from /home/ubuntu/automlbenchmark/resources/frameworks.yaml.
Loading benchmark constraint definitions from /home/ubuntu/automlbenchmark/resources/constraints.yaml.
Loading benchmark definitions from /home/ubuntu/automlbenchmark/resources/benchmarks/test.yaml.
type object 'Benchmark' has no attribute 'SetupMode'
Traceback (most recent call last):
  File "runbenchmark.py", line 126, in <module>
    bench.setup(amlb.SetupMode[args.setup])
  File "/home/ubuntu/automlbenchmark/amlb/docker.py", line 54, in setup
    if mode == Benchmark.SetupMode.skip:
AttributeError: type object 'Benchmark' has no attribute 'SetupMode'
```
Seems in `docker.py` there is a bug. I tried to import `SetupMode` and using `SetupMode` instead of `Benchmark.SetupMode` in the code and it works. 
Just want to mention this bug here. If help is needed, I can do a pr related to this.

sebhrusen:
@Rong-Inspur thanks so much raising this.
I'm really sorry, looks like I broke in my last PR.
Fixing this now.

Rong-Inspur:
Cool!

PR #76, opened 2020-02-27
fix SetupMode namespace

sebhrusen:
https://github.com/openml/automlbenchmark/issues/75


sebhrusen:
@Rong-Inspur should be similar to what you did to fix it locally. Can you please confirm?

Rong-Inspur:
> @Rong-Inspur should be similar to what you did to fix it locally. Can you please confirm?

Yes, that's what I did locally.

PR #83, opened 2020-03-27
Fix wrong autosklearn requirements v0.5.1

franchuterivera:
Current framework install dependencies from:
curl https://raw.githubusercontent.com/automl/auto-sklearn/master/requirements.txt | sed '/^$/d' | while read -r i; do PIP install "$i"; done

Above dependencies are meant for auto-sklearn v0.6.0, not v0.5.1 -- in this case, the framework is broken as xgboost support is missing 

Can you consider updating the requirement files with the explicit requirements to fix auto-sklearn framework?


PGijsbers:
Looks good to me. I think this should be preferred as we want to make sure versions are clearly defined when running the benchmark. @sebhrusen thoughts?

PGijsbers:
Thank you @franchuterivera! 🎉

PR #85, opened 2020-03-31
Singularity support

franchuterivera:
This is a pull request to provide singularity container support to the automlbenchmark.

The support has been created as a replica of docker support, which can be seen in the amlb/singularity.py

There is one pending Item in resources/config.yaml  . There is a new keyword called library. In this keyword, one must add the organization library in which we would upload the images to be pull. In docker, one just do: docker push to put into the registry, in singularity one needs to create a library as detailed in here https://sylabs.io/guides/3.2/user-guide/cloud_library.html.

That is we need to change the support from:
singularity push <local image name> shub://**francisco.rivera.valverde/default**/<image name>:latest
to:
singularity push <local image name> shub://**ORGANIZATION NAME**/<image name>:latest

I am sorry to put this on your plate, but we have to create this organization support from where to download from https://sylabs.io/guides/3.2/user-guide/cloud_library.html. I hope to use this pull request as a way to start this conversation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,10 @@ _site/
 *.bz2
 *.png
 
+# Singurality files
+Singularityfile
+*.sif
+
 #whitelist
 !reports/**/*.csv
 !reports/**/*.png
--- a/amlb/__init__.py
+++ b/amlb/__init__.py
@@ -7,6 +7,7 @@ from .errors import AutoMLError
 from .resources import Resources
 from .benchmark import Benchmark, SetupMode
 from .docker import DockerBenchmark
+from .singularity import SingularityBenchmark
 from .aws import AWSBenchmark, AWSRemoteBenchmark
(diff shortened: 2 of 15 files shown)


sebhrusen:
@franchuterivera I'm not familiar with Singularity, and it may be a useful addition, I'm open to it: I'm mainly concerned by the effort to maintain this.
May I ask you what is your motivation behind adding Singularity vs Docker?

Otherwise, if the main difference between both implementations is the image creation script + upload, then I'd like to abstract those implementations to avoid duplicate code (those are annoying when we have to extend the app).
Can I suggest sth like `container.py` as a sort of abstract module, with `docker.py` and `singularity.py` as implementations?

franchuterivera:
Wait, I think I started with the wrong foot on this one. Let me create an issue where I give the details and link my new pull request with above recommendations.

Issue #84, opened 2020-03-30
Upgrade frameworks to their latest stable version


sebhrusen:
- autosklean: 0.6.0
- TPOT: 0.11.1
- H2O: 3.30.x
...

franchuterivera:
Hello, I am working to create a pull request to move to autosklean==0.6.0 (working with the autosklearn developer to do so).
I guess it makes sense not to duplicate effort...

sebhrusen:
@franchuterivera sure, thanks for that. Nothing started on my side. Please refer to this ticket once you create your PR.

PR #86, opened 2020-03-31
Moving sklearn to v0.6.0

franchuterivera:
This pull request is related to #84 

It moves autosklearn version to v0.6.0.

I also took advantage and move autosklearn to the venv feature, following the recommendation given in above issue.

It is important to note that I had to add some procs to frameworks/shared/callee.py that might benefit other frameworks, yet added some custom memory management inside auto-sklearn (that required psutil).

This has been tested in autosklearn -m local and -m docker. I also ran RandomForest to make sure nothing is broken.


sebhrusen:
tested on macOS: 
- docker mode is working well.
- local mode fails on macOS: I've seen it working in the past but was not stable, not really important for benchmarks anyway.
 merging.

Issue #90, opened 2020-04-07
Problem With Singularity Pull Command


franchuterivera:
There is an issue with amlb/singularity.py, likely produced by a name change.

I think when doing the merge, you guys saw fit to rename:

def _container_image_name(self, branch=None, **return_docker_name**=False):

to be:

image_name(cls, framework_def, branch=None, **as_docker_image**=False, **kwargs):

We need to replace all occurrences of return_docker_name to be as_docker_image. For instance 

https://github.com/openml/automlbenchmark/blob/2842df7ada191d0a0e9a7d7ac9096439714b029c/amlb/singularity.py#L130

As of now, pulling from a singularity command fails and is masked by the exception catching.

PGijsbers:
Looks like it probably got introduced by #89 @sebhrusen

sebhrusen:
@franchuterivera I'm really sorry, I had to make a quick fix yesterday because the singularity changes broke the AWS support when using docker image on EC2.
I should have retested singularity properly.
Fixing asap.

sebhrusen:
fixed.
@franchuterivera if you have any idea about my comment https://github.com/openml/automlbenchmark/pull/91#discussion_r404910131
It would be a useful improvement, thx

PR #91, opened 2020-04-07
Issue 90: fix Singularity support

sebhrusen:
https://github.com/openml/automlbenchmark/issues/90

### DIFF
--- a/amlb/singularity.py
+++ b/amlb/singularity.py
@@ -127,7 +127,7 @@ class SingularityBenchmark(ContainerBenchmark):
         try:
             # We pull from docker as there are not yet singularity org accounts
             run_cmd("singularity pull {output_file} docker://{image}".format(
-                image=self._container_image_name(return_docker_name=True),
+                image=self._container_image_name(as_docker_image=True),
                 output_file=self._image_name,
             ), _live_output_=True)
             return True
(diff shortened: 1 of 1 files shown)


sebhrusen:
testing it now

amlb/singularity.py
```diff
@@ -127,15 +127,15 @@ def _image_exists(self):
         try:
             # We pull from docker as there are not yet singularity org accounts
             run_cmd("singularity pull {output_file} docker://{image}".format(
-                image=self._container_image_name(return_docker_name=True),
+                image=self._container_image_name(as_docker_image=True),
                 output_file=self._image_name,
             ), _live_output_=True)
```
sebhrusen, line 132: @franchuterivera any idea why the live output doesn't seem to work with `singularity pull`?
My logic has been working with all commands I've been using until now, including when building or pulling docker image, but here I don't see any output/feedback when pulling singularity image, which is very annoying because it's hard to know if it's really doing anything or if it just got stuck...

sebhrusen: looks like singularity pull is buffering the output, idk..

franchuterivera: Yeah, same behavior on my end... there is an output, but after the command has completed, which takes like 5 mins... If I found a solution I would let you know! thanks for looking into this...

Issue #102, opened 2020-04-21
Add new framework: mljar-supervised


pplonski:
I'm working on AutoML python package: [mljar-supervised](https://github.com/mljar/mljar-supervised). I would like to add it to `automlbenchmark`. 

I successfully run through 1-9 points from [HOWTO::Add default framework](https://github.com/pplonski/automlbenchmark/blob/master/docs/HOWTO.md#add-a-default-framework)

I stuck at 10 (docker integration) and 11 (aws).

For docker I susspect I had problem with privilages:

```
python3 runbenchmark.py supervised -m docker
Running `supervised` on `test` benchmarks in `docker` mode.
Loading frameworks definitions from /home/piotr/sandbox/automlbenchmark/resources/frameworks.yaml.
Loading benchmark constraint definitions from /home/piotr/sandbox/automlbenchmark/resources/constraints.yaml.
Loading benchmark definitions from /home/piotr/sandbox/automlbenchmark/resources/benchmarks/test.yaml.
Running cmd `docker images -q automlbenchmark/supervised:0.2.6-stable`
WARNING: Error loading config file: /home/piotr/.docker/config.json: stat /home/piotr/.docker/config.json: permission denied
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.38/images/json?filters=%7B%22reference%22%3A%7B%22automlbenchmark%2Fsupervised%3A0.2.6-stable%22%3Atrue%7D%7D: dial unix /var/run/docker.sock: connect: permission denied
```

For aws I got a lot of loga and then errors:
```
[  119.518648] cloud-init[1475]:     Running setup.py install for openml: finished with status 'done'
[...]

PR #105, opened 2020-04-27
add mljar-supervised framework

pplonski:
I've added [mljar-supervised](https://github.com/mljar/mljar-supervised) framework.

It should work in all modes: locally, docker, and aws.

The mljar-supervised is in a 0.2.8 version.


frameworks/supervised/exec.py
```diff
@@ -0,0 +1,96 @@
...
+    algorithms = ["Baseline", "Decision Tree", "Random Forest", "Xgboost", "LightGBM"]
+    if config.max_runtime_seconds <= 60:
+        algorithms = ["Baseline"]
+    results_path = "AutoML_results"
+    shutil.rmtree(results_path, ignore_errors=True) # clear just in case
+    automl = AutoML(
+        results_path=results_path,
+        algorithms=algorithms,
+        total_time_limit=config.max_runtime_seconds,
+        seed=12,
```
sebhrusen, line 65: please use
```
seed=config.seed
```

frameworks/supervised/exec.py
```diff
@@ -0,0 +1,96 @@
...
+    # however, in case of multiclass classification, we will set the ml_task manually.
+    # AutoML assumes that multiclass calssification is when there is from 2 to 20 unique values.
+    # In this benchmark, there can be casses when it is much more classes than 20. That's why we set manually.
+    if is_classification and len(np.unique(y_train)) > 2:
+        ml_task = "multiclass_classification"
+
+    algorithms = ["Baseline", "Decision Tree", "Random Forest", "Xgboost", "LightGBM"]
+    if config.max_runtime_seconds <= 60:
+        algorithms = ["Baseline"]
+    results_path = "AutoML_results"
```
sebhrusen, line 59: is it relative to pwd?
can you please ensure it goes with the other outputs from this framework? 
Also when running multiple tasks/folds on the same instance, we usually want to distinguish all those outputs, that's why for other frameworks, we use the `make_subdir` function below:
```
results_path = make_subdir("results", config)
...

def make_subdir(name, config):
    subdir = os.path.join(config.output_dir, name, config.name, str(config.fold))
    touch(subdir, as_dir=True)
    return subdir
```

frameworks/supervised/exec.py
```diff
@@ -0,0 +1,96 @@
...
+    ml_task = None  # the AutoML will guess about ML task
+    # however, in case of multiclass classification, we will set the ml_task manually.
+    # AutoML assumes that multiclass calssification is when there is from 2 to 20 unique values.
+    # In this benchmark, there can be casses when it is much more classes than 20. That's why we set manually.
+    if is_classification and len(np.unique(y_train)) > 2:
+        ml_task = "multiclass_classification"
+
+    algorithms = ["Baseline", "Decision Tree", "Random Forest", "Xgboost", "LightGBM"]
+    if config.max_runtime_seconds <= 60:
+        algorithms = ["Baseline"]
```
sebhrusen, line 58: what happens if you don't specify the `algorithms` param?
One of the requirement for the integration in this app is that frameworks should run with their defaults (outside time, memory constraints, seed and other params that don't influence the training, like your `results_path`).

In your case, it would be preferable to call AutoML as follow:
```
   training_params = {k: v for k, v in config.framework_params.items() if not k.startswith('_')}

    automl = AutoML(
        results_path=results_path,
        ml_task=ml_task,
        total_time_limit=config.max_runtime_seconds,
        seed=config.seed,
        **training_params
    )
```

and let the framework use the default algorithms: user can always then create a custom definition if he wants to specify those params (see https://github.com/openml/automlbenchmark/blob/master/docs/HOWTO.md#run-a-framework-with-different-hyper-parameters)

sebhrusen: I just added our requirements, as I realize it was missing: https://github.com/openml/automlbenchmark/blob/master/docs/HOWTO.md#add-a-default-framework

frameworks/supervised/exec.py
```diff
@@ -0,0 +1,96 @@
...
+    predictions, probabilities = None, None
+    if is_classification:
+        predictions = preds["label"].values
+        probabilities = preds[preds.columns[:-1]].values
+    else:
+        predictions = preds["prediction"].values
+
+    # clean the results 
+    # if you want to inspect results, please comment below line
+    shutil.rmtree(results_path, ignore_errors=True)
```
sebhrusen, line 83: also see `_save_artifacts` support in `autosklearn`, `H2OAutoML` and `TPOT` for example: this allows the user to define at runtime (using the framework definition) if he wants to keep the generated artifacts or not.

pplonski: @sebhrusen thank you for the review! I will try to fix all issues soon.

sebhrusen:
@pplonski thanks for this new addition!
Please have a look at my comment to put this to the same standard as other frameworks.
When the PR looks good, I'll test it against the `validation` dataset on  docker to confirm that everything works fine.

frameworks/supervised/exec.py
pplonski: thanks, I corrected this

frameworks/supervised/exec.py
pplonski: I've used `make_subdir` method.

frameworks/supervised/exec.py
pplonski: As suggested, I have used `training_params`.

frameworks/supervised/exec.py
pplonski: I have used `_save_artifacts` for deciding whether to keep results. By default, my package is always storing all results in the hard drive.

pplonski:
@sebhrusen Thank you for the review! I've corrected all suggestions.

I'm running tests right now, I would like to check `docker` and `aws`. I will let you know after all is ready.

frameworks/supervised/exec.py
pplonski: I was getting error messages in xgboost with `config.seed` values and it looks like a problem with xgboost package https://github.com/dmlc/xgboost/issues/5637

frameworks/supervised/exec.py
pplonski: @sebhrusen I have one question. By default the `mljar-supervised` is producing explanations for models. The computation of explanation is very time-consuming. For normal usage,  explanations are needed. For benchmark reasons they should be switched off.

Should I switch-off explanations in `exec.py` or add param in `frameworks.yaml` ? I'd like to keep the switched-on in the `mljar-supervised` package.

pplonski:
@sebhrusen I have no idea why, but when I'm running benchmarks in `aws` mode, the script is installing an old version of `mljar-supervised` even if a newer version is specified in requirements.txt file. Any idea why?

Docker is using the correct version of the package.

pplonski:
@sebhrusen the `docker` mode on validation data is working:
```
                    id                     task   framework fold    result    mode version params                  utc  duration models       seed       acc       auc   logloss
0    openml.org/t/9910              bioresponse  supervised    0  0.882951  docker   0.3.4         2020-05-07T11:15:08    3538.4     11  564956983  0.803191  0.882951       NaN
1  openml.org/t/125920            dresses-sales  supervised    0  0.564860  docker   0.3.4         2020-05-07T11:55:44    2432.3     95  564956983  0.540000  0.564860       NaN
2    openml.org/t/2079               eucalyptus  supervised    0  0.680128  docker   0.3.4         2020-05-07T12:53:10    3442.5     58  564956983  0.662162       NaN  0.680128
3  openml.org/t/167125  internet-advertisements  supervised    0  0.933241  docker   0.3.4         2020-05-07T13:45:56    3086.0     11  564956983  0.951220  0.933241       NaN
4    openml.org/t/9950               micro-mass  supervised    0  0.370797  docker   0.3.4         2020-05-07T14:30:51    2684.2     14  564956983  0.862069       NaN  0.370797
5    openml.org/t/3917                      kc1  supervised    0  0.817650  docker   0.3.4         2020-05-07T15:33:19    3741.8     89  564956983  0.848341  0.817650       NaN
6  openml.org/t/168868               APSFailure  supervised    0  0.991915  docker   0.3.4         2020-05-07T16:28:45    3207.3     14  564956983  0.980132  0.991915       NaN
Job docker_validation_1h4c_all_0_supervised executed in 22552.541 seconds.
```

The code is ready for review, please take a look. Thank you!

frameworks/supervised/exec.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+    # cast columns to have float types
+    for col in X_train.columns:
+        try: 
+            x1 = X_train[col].astype(float)
+            X_train[col] = x1
+            x2 = X_test[col].astype(float)
+            X_test[col] = x2
+        except Exception as e:
```
sebhrusen, line 43: does it mean you're ignoring all categorical columns?

sebhrusen: if so, why no using encoded data instead, like for `autosklearn` for example.

frameworks/supervised/exec.py
```diff
@@ -0,0 +1,91 @@
...
+def run(dataset, config):
+    log.info("\n**** mljar-supervised ****\n")
+
+    is_classification = config.type == "classification"
+
+    X_train, X_test = dataset.train.X, dataset.test.X
+    y_train, y_test = dataset.train.y, dataset.test.y
+
+    X_train = pd.DataFrame(X_train, columns=[f"f{i}" for i in range(X_train.shape[1])])
+    X_test = pd.DataFrame(X_test, columns=[f"f{i}" for i in range(X_test.shape[1])])
```
sebhrusen, line 30: if, for some reason, you want the real column names, see the integration for AutoGluon here: https://github.com/openml/automlbenchmark/pull/109/files#diff-afee22ac53339a4a6051bba646ad17c3

sebhrusen:
> @sebhrusen I have no idea why, but when I'm running benchmarks in `aws` mode, the script is installing an old version of `mljar-supervised` even if a newer version is specified in requirements.txt file. Any idea why?
> 
> Docker is using the correct version of the package.

the most likely explanation is that your `aws` run was using a different branch...
In your custom `config.yaml`, were you always setting?
```
project_repository: https://github.com/mljar/automlbenchmark#master
```

resources/frameworks.yaml
```diff
@@ -94,3 +94,11 @@ GAMA:
 autoxgboost:
   version: '0.0.0.9000'
   project: https://github.com/ja-thomas/autoxgboost
+
+supervised:
+  version: '0.3.4'
+  project: https://github.com/mljar/mljar-supervised
+#  params:
+#    explain_level: 0
```
sebhrusen, line 102: given your comment regarding explanation, you can reasonable disable this feature here and explain why:
```
params:
   explain_level: 0   #unnecessary for benchmarks and time-consuming

File: resources/frameworks.yaml [hunk 8]
```diff
@@ -94,3 +94,11 @@ GAMA:
 autoxgboost:
   version: '0.0.0.9000'
   project: https://github.com/ja-thomas/autoxgboost
+
+supervised:
```
sebhrusen, line 98: didn't notice this before, but can you give a more explicit name to your framework (same concern for the directory), I'd recommand `mljarsupervised`.
`mljar_supervised` should also work, but I'm only 95% sure (there may be some name resolution/regexp issues when used with a dataset that also has an underscore in its name).

frameworks/supervised/exec.py
sebhrusen: see https://github.com/openml/automlbenchmark/pull/105#discussion_r423230748

frameworks/supervised/exec.py
sebhrusen: that's weird, our random seeds are also limited to int32 for the same reason that there was issues with larger seeds for some frameworks.
in `amlb/resources.py`:
``` 
            return random.randint(1, (1 << 32) - 1)  # limiting seed to int32
```

frameworks/supervised/exec.py
pplonski: Thank you for review! :) 

The problem is that all columns have `object` type - all looks like categoricals. That's why I'm trying to cast into `float` type. If they can't be cast then they are left as `object`. My framework distinguish categoricals and numeric, it is using pandas types for this.

resources/frameworks.yaml
pplonski: OK!

resources/frameworks.yaml
pplonski: I liked `supervised` :) I will change to `mljarsupervised`.

frameworks/supervised/exec.py
pplonski: xgboost is using signed 32-bit integer value as seed, it will be changed in next version

frameworks/supervised/exec.py
sebhrusen: gotcha!
if you want a bit cleaner way to do this, you can previously list all columns that are categoricals, see https://github.com/openml/automlbenchmark/blob/master/frameworks/autosklearn/__init__.py

pplonski: from what I remember all columns are `object` type that's why loop over all of them

sebhrusen: `object` type is at `numpy` level, but when parsing the data file, especially if it's in ARFF format, we can retrieve the "real" type, this is why the snippet above can distinguish categoricals from numericals.

sebhrusen: typically this would be more explicit on your side:
in `__init__`:
```
   data = dict(
        train=dict(
            X=dataset.train.X,
            y=dataset.train.y
        ),
        ...
        column_names = [p.name for p in dataset.predictors]
        column_type={p.name: ('object' if p.is_categorical() else 'float') for p in dataset.predictors}
    )
```
in `exec`:
```
X_train = pd.DataFrame(X_train, columns=data.column_names).astype(column_types)
```

pplonski: @sebhrusen thank you! I've used your example, code looks much better.

frameworks/supervised/exec.py
pplonski: Yep, changed!

pplonski:
@sebhrusen thank you for review! All comments from review added. Code is ready

Blow output for `docker` mode on validation sets:
```
                    id                     task        framework fold    result    mode version                params                  utc  duration models        seed       acc       auc   logloss
0    openml.org/t/9910              bioresponse  mljarsupervised    0  0.834929  docker   0.3.5  {'explain_level': 0}  2020-05-14T06:31:16     902.8     70  2108295964  0.587766  0.834929       NaN
1  openml.org/t/125920            dresses-sales  mljarsupervised    0  0.579639  docker   0.3.5  {'explain_level': 0}  2020-05-14T06:42:16     656.9    102  2108295964  0.500000  0.579639       NaN
2    openml.org/t/2079               eucalyptus  mljarsupervised    0  0.688925  docker   0.3.5  {'explain_level': 0}  2020-05-14T06:53:45     685.5    100  2108295964  0.635135       NaN  0.688925
3  openml.org/t/167125  internet-advertisements  mljarsupervised    0  0.948890  docker   0.3.5  {'explain_level': 0}  2020-05-14T07:12:26    1021.7      6  2108295964  0.960366  0.948890       NaN
4    openml.org/t/9950               micro-mass  mljarsupervised    0  0.409021  docker   0.3.5  {'explain_level': 0}  2020-05-14T07:26:39     847.6     84  2108295964  0.896552       NaN  0.409021
5    openml.org/t/3917                      kc1  mljarsupervised    0  0.820094  docker   0.3.5  {'explain_level': 0}  2020-05-14T07:38:52     729.4     88  2108295964  0.848341  0.820094       NaN
6  openml.org/t/168868               APSFailure  mljarsupervised    0  0.990533  docker   0.3.5  {'explain_level': 0}  2020-05-14T07:54:55     910.0     22  2108295964  0.984342  0.990533       NaN

```

sebhrusen:
Hey @pplonski , sorry was very busy lately.
I'll try to make a final review + test very soon.
If you don't hear from me before end of next week, don't hesitate to ping me back!

pplonski:
Thank you @sebhrusen!

frameworks/mljarsupervised/exec.py
```diff
@@ -0,0 +1,77 @@
...
+import numpy as np
+
+from frameworks.shared.callee import call_run, result, Timer, touch
+
+log = logging.getLogger(os.path.basename(__file__))
+
+import supervised
+from supervised.automl import AutoML
+import shutil
+
```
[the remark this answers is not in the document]
pplonski: I will fix it with newer version of mljar-supervised (should be soon).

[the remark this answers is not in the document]
sebhrusen: Hi @pplonski. Are you still interested in adding mljar to the  list of supported frameworks? We plan to run a series of benchmarks soon (maybe August).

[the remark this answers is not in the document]
pplonski: @sebhrusen thanks for reaching out! Yes, I'm interested. I will all issues in July.

pplonski:
@sebhrusen, changes:
- imports order fixed
- newer version of the `mljar-supervised` (0.5.5) set
- `exec.py` file formated

frameworks/mljarsupervised/setup.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+HERE=$(dirname "$0")
+#AMLB_DIR="$1"
+# by passing the module directory to `setup.sh`, it tells it to automatically create a virtual env under the current module.
+# this virtual env is then used to run the exec.py only, and can be configured here using `PIP` and `PY` commands.
+. $HERE/../shared/setup.sh $HERE
+#. $AMLB_DIR/frameworks/shared/setup.sh $HERE
+PIP install -r $HERE/requirements.txt
```
sebhrusen, line 8: I just pushed changes for `setup` recently to simplify version upgrade and to make it easier to benchmark multiple versions of the same framework.
Can you please change your `setup.sh` following the `autosklearn` example? https://github.com/openml/automlbenchmark/blob/master/frameworks/autosklearn/setup.sh

You should also be able to remove the `requirements.txt` then.

frameworks/mljarsupervised/exec.py
```diff
@@ -0,0 +1,84 @@
...
+from frameworks.shared.callee import call_run, result, Timer, touch
+
+matplotlib.use("agg")  # no need for tk
+log = logging.getLogger(os.path.basename(__file__))
+
+def make_subdir(name, config):
+    subdir = os.path.join(config.output_dir, name, config.name, str(config.fold))
+    touch(subdir, as_dir=True)
+    return subdir
+
```
sebhrusen, line 18: I added this function to `frameworks.shared.callee` as it is now used in almost every integration:
```
from frameworks.shared.callee import output_subdir
```
this will save you a few lines.

pplonski: Thanks! Changed.

frameworks/mljarsupervised/setup.sh
pplonski: Sure! `setup.sh` is changed

resources/frameworks.yaml
```diff
@@ -109,3 +109,11 @@ GAMA:
 autoxgboost:
   version: '0.0.0.9000'
   project: https://github.com/ja-thomas/autoxgboost
+
+mljarsupervised:
+  version: '0.5.5'
+  project: https://github.com/mljar/mljar-supervised
+  params:
+    mode: "Compete"   # set mode for Compete, default mode is Explain
```
sebhrusen, line 117: @pplonski for the "official" openml benchmark, we currently require that all AutoML frameworks use their default configuration outside the time/cores/metrics constraints imposed by benchmark app, and the parameters defining the ML problem (+dataset).
This `mode` parameter should then be commented out.

@PGijsbers what do you think about this kind of parameter? Given the evolution of focus of many ML frameworks, we see more and more presets/high-level configs like that (AutoGluon has a similar approach, I'm thinking about sth similar for H2OAutoML...).
I sometimes find it weird that we refuse this kind of param when we benchmark only on prediction performance.

pplonski:
@sebhrusen I understand that it will be the most elegant way to check all frameworks with defaults. The default mode of work for `mljar-supervised` is `Explain` which produces extensive Markdown reports. It is not even full cross-validation, just 80/20 split, and not all available algorithms are used. I've used this set-up to be user-friendly. The mode `Compete` will run with a full 10-fold CV and all available algorithms, ensembling+stacking. This mode was designed for ML competitions. In my opinion, the AutoML goal is more than only competing on the score.

sebhrusen:
> This mode was designed for ML competitions. In my opinion, the AutoML goal is more than only competing on the score.

@pplonski I completely agree with you, I just want the OpenML members involved in the comparison study to be aware of this concern. 
To be clear, it doesn't affect only your framework, and any user is free to add as many definitions for the same framework in its local `frameworks.yaml` (see https://github.com/openml/automlbenchmark/blob/master/docs/HOWTO.md#framework-definition).
But for simplicity — and simplicity may be unfair sometimes —, the definitions in the default `resources/frameworks.yaml` have no params, and these are the definitions we currently use for the OpenML benchmarks. This may change in the future, but there's no agreement yet.

pplonski:
Thanks @sebhrusen, please let me know if you will know how it will be handled in the comparison. I can change the default in the package.

pplonski:
Hi @sebhrusen, do you know when this PR will be merged? 

How the process of package update looks like after merge? Let's say, I have a new major version update of my package, then I should create new PR and update the version?

sebhrusen:
@pplonski let's merge it: we're having a separate discussion to decide if we'll allow the `mode` param or not.
If we decide not to, I'll create a PR and keep you informed.

Regarding package update, ideally it should be done as you suggest if you want it to be always up-to-date.
On our side, we're just checking framework versions only once a while on a very irregular basis, like once every 3 months, and the more frameworks get supported, the more we should rely on the framework maintainer as you suggest.

pplonski:
Thank you @sebhrusen you are doing amazing work for ML community!

sebhrusen:
@pplonski Thanks for your contribution and for making this project more and more useful!

This completes https://github.com/openml/automlbenchmark/issues/102

Issue #107, opened 2020-04-29
ConnectionResetError while doing multiple H2O runs


franchuterivera:
When H2O is initialized, it is done so in the same hostname/port.

If multiple runs for this framework are launched in parallel it causes a crash. For instance, ConnectionResetError(104, 'Connection reset by peer'))  happens on an H2OAutoML job called B, when a job A that was launched few mins before job B finishes and orders hostname:port to close the connection.

This issue tracks the proposal to isolate the H2OAutoML ports per run using the PID

PR #108, opened 2020-04-29
Fix port configuration in H2O for multiple runs

franchuterivera:
This request is related to #107 and it only makes sure that the h2o runs are isolated.

To do so we condition on the PID to build the port number.


frameworks/H2OAutoML/exec.py
```diff
@@ -36,7 +36,11 @@ def run(dataset: Dataset, config: TaskConfig):
         nthreads = config.framework_params.get('_nthreads', config.cores)
 
         log.info("Starting H2O cluster with %s cores, %sMB memory.", nthreads, config.max_mem_size_mb)
+        max_port_range = 49151
+        min_port_range = 1000
```
sebhrusen, line 40: I guess you wanted to limit to registered ports
```suggestion
        min_port_range = 1024
```

Issue #93, opened 2020-04-07
Add AutoGluon framework


sebhrusen:
https://github.com/awslabs/autogluon
Last stable version seems to be v0.0.6.

@Innixma do you have any concern if we want to make Autogluon publicly available in the benchmark application?

Innixma:
No concern, I think this would be great!

We actually used automlbenchmark for our paper: https://arxiv.org/pdf/2003.06505.pdf

I have a working version of AutoGluon in my fork of the repo: https://github.com/Innixma/automlbenchmark

I've only tested it on the local version, and its based on a fairly outdated version of automlbenchmark, so you may have to rework parts of it.

One note: At least in local, this fix needs to be added for AG to work, as AG imports numba and numba must be imported prior to some of the imports that occur in automlbenchmark: https://github.com/Innixma/automlbenchmark/blob/master/amlb/logger.py#L9

sebhrusen:
@Innixma great!
I'll use your fork to make the integration easier, will only create the default `autogluon`, it's then possible to create additional frameworks purely declaratively in a local `framework.yaml`.

Same for the `auto_stack=True`: our policy for AutoML tools is to integrate then and provide a default definition in `resources/frameworks.yaml` with only the default parameters defined by the AutoML tool itself.
The integration only set parameters related with time limit and resources (memory, cpu...) if any to ensure that the toolkit provides a result after the time limit.
But as mentioned above, we also integrate those in such a way that user can activate feature declaratively in a custom `frameworks.yaml`:
```
autogluon_autostack:
   extends: autogluon
   params:
       auto_stack: true

autogluon_noknn:
    extends: autogluon
    params:
         #your noknn params here
```
Those params are then automatically included in the benchmark results file, which also helps to analyze results.

Also, there have been some improvements to the benchmark app since, and AutoGluon will be able to run in its own process in a dedicated virtualenv, so it should make the setup easier and the `numba` fix unnecessary.

Innixma:
Please use https://github.com/Innixma/automlbenchmark/tree/autogluon-test for the latest version, I made some recent changes to reduce the amount of internal AutoGluon functions being used in the wrapper.

The declarative functionality sounds great, definitely easier than what I was doing!

PR #109, opened 2020-04-29
add AutoGluon framework

sebhrusen:
https://github.com/openml/automlbenchmark/issues/93


frameworks/AutoGluon/exec.py
```diff
@@ -0,0 +1,114 @@
...
+    with Timer() as predict:
+        predictions = predictor.predict(X_test)
+
+    probabilities = predictor.predict_proba(X_test, as_pandas=True) if is_classification else None
+    if is_classification and len(probabilities.shape) == 1:
+        # for binary, AutoGluon returns only one probability and it's not systematically the smaller or greater label.
+        # Here, we're trying to identify to which label correspond the probabilities
+        # and from there we can build the probabilities for all labels.
+        classes = dataset.target.classes
+        prob_class = next((predictions[i] if p > 0.5 else next(c for c in classes if c != predictions[i])
```
Innixma, line 66: Soon in latest AutoGluon mainline (will be in 0.0.7 release, which will likely occur by EOW), you can do this:

```
probabilities = predictor.predict_proba(dataset=X_test, as_pandas=True, as_multiclass=True) if is_classification else None
```

This should avoid the need for the entire if statement code. You will only need to check and confirm that dataset.target.classes == predictor.class_labels (if not, you may have to reorder output columns)

PR with fix: https://github.com/awslabs/autogluon/pull/447

frameworks/AutoGluon/exec.py
```diff
@@ -0,0 +1,114 @@
+import logging
```
Innixma, line 1: Regarding CatBoost files polluting the working directory, this has been fixed and will be included in the upcoming 0.0.7 release.

PR with fix: https://github.com/awslabs/autogluon/pull/446

frameworks/AutoGluon/exec.py
```diff
@@ -0,0 +1,114 @@
...
+        log.warning("Performance metric %s not supported.", config.metric)
+
+    is_classification = config.type == 'classification'
+
+    train = pd.DataFrame(dataset.train.data, columns=dataset.columns)
+    label = dataset.target.name
+
+    output_dir = make_subdir("models", config)
+    with Timer() as training:
+        predictor = task.fit(
```
Innixma, line 45: No usage of config to specify AutoGluon fit args?

frameworks/AutoGluon/requirements.txt
```diff
@@ -0,0 +1,2 @@
+mxnet
+autogluon==0.0.6
```
Innixma, line 2: autogluon==0.0.7 will release very soon, I would recommend updating this PR to support 0.0.7 before merging.

frameworks/AutoGluon/setup.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+HERE=$(dirname "$0")
+
+# creating local venv
```
Innixma, line 4: Regarding libomp, this is the setup code I use for internal testing:

```
if [[ "$OSTYPE" == "darwin"* ]]; then
    echo "Operating System is Mac, running homebrew install of libomp to enable LightGBM"
    # Install libomp to support LightGBM package on mac
    if brew ls --versions libomp > /dev/null; then
        echo "libomp already installed..."
    else
        brew install libomp
    fi
fi
```

I don't believe libomp installation is required on other operating systems, and I'm not sure how `SUDO apt-get install -y libomp-dev` differs from `brew install libomp`. I'm curious to know if there is a difference, since I would be interested in adding it to our setup documentation to assist users that don't have brew installed.

frameworks/AutoGluon/exec.py
```diff
@@ -0,0 +1,114 @@
...
+
+def save_artifacts(predictor, leaderboard, config):
+    artifacts = config.framework_params.get('_save_artifacts', ['leaderboard'])
+    try:
+        if 'leaderboard' in artifacts:
+            models_dir = make_subdir("models", config)
+            save_pd.save(path=os.path.join(models_dir, "leaderboard.csv"), df=leaderboard)
+
+        if 'info' in artifacts:
+            ag_info = predictor._learner.get_info()
```
Innixma, line 106: Recommended to change to predictor.info() in 0.0.7

frameworks/AutoGluon/exec.py
sebhrusen: good catch, thx! 
I was also surprised to not see differences when running `AutoGluon` definition vs. `AutoGluon_autostack`.

frameworks/AutoGluon/requirements.txt
sebhrusen: great, given the list of improvements you mentioned in your comments, I'll definitely wait.

frameworks/AutoGluon/setup.sh
sebhrusen: I'm just making the assumption that macOS users will have brew installed (others have to install missing dependencies manually).
The `apt-get install` is mainly for debian-based distros: this is what we're using to build the docker images, and/or to run the benchmarks on AWS.

frameworks/AutoGluon/exec.py
```diff
@@ -0,0 +1,114 @@
...
+                         else np.array([[row, 1-row] for row in probabilities]))
+
+    leaderboard = predictor._learner.leaderboard(X_test, y_test, silent=True)
+    with pd.option_context('display.max_rows', None, 'display.max_columns', None, 'display.width', 1000):
+        print(leaderboard)
+
+    save_artifacts(predictor, leaderboard, config)
+
+    num_models_trained = len(leaderboard)
+    num_models_ensemble = len(leaderboard[leaderboard['stack_level'] > 0])
```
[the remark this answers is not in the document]
sebhrusen: I probably misunderstood your original intent.
I'll switch it to your proposal if that's the information you're interested in.

frameworks/AutoGluon/setup.sh
Innixma: You are saying that debian-based distros don't work out of the box with LightGBM? I have only tested with Ubuntu which does not require any installation. If so I will add that to AutoGluon's [installation instructions](https://autogluon.mxnet.io/).

sebhrusen: to be honest, I didn't try without the current state of this `setup.sh` file.
I'm working on macOS and as I had to install `libomp` there, I made the assumption that it was also needed on Ubuntu, but if you're saying that it's not necessary, then I'll try to remove it.

Innixma:
0.0.7 has been released, so I believe my comments can now be addressed.

One additional recommendation:

For AutoGluon, I'd recommend having a preset which is `{presets='best_quality'}` and also potentially `{presets='good_quality_faster_inference_only_refit'}`.

'best_quality' is shorthand for auto_stack=True for new users, but if we find a better preset than simply auto_stack=True in future, it will be reflected in presets='best_quality', to ensure that automlbenchmark utilizes the most competitive configuration for AutoGluon regardless of version.

Additionally, presets='good_quality_faster_inference_only_refit' is a middle ground between best quality and fast inference. In my tests it results in inference speeds similar to the other frameworks, so it may be useful to have if we ever consider inference time as a factor.

frameworks/AutoGluon/exec.py
Innixma, line 45: We should pass problem_type here as some problems are ambiguous between being regression and multiclass.

Innixma:
Regarding 0.0.7 release, we have a PyPi issue and are trying to resolve it, at the moment 0.0.7 will not install properly from git.

Innixma:
AutoGluon 0.0.8 is released and addresses the PyPi issue, please use 0.0.8 for this PR.

Best,
Nick

sebhrusen:
Thanks @Innixma, I will apply the changes today.

resources/frameworks.yaml
```diff
@@ -46,6 +46,16 @@ TunedRandomForest:
 #      n_estimators: 500
 
 ### AutoML frameworks
+AutoGluon:
+  version: "0.0.8"
```
Innixma, line 50: Just released autogluon==0.0.9 as an emergency fix to sklearn==0.23 breaking skopt, which resulted in autogluon failing to import. Would recommend 0.0.9 to avoid having to specify sklearn version

frameworks/AutoGluon/requirements.txt
```diff
@@ -0,0 +1,3 @@
+mxnet
+scikit-learn==0.22.2
```
sebhrusen, line 2: @Innixma for some reason, I had to enforce scikit-learn version.
When I first updated autogluon version, everything was working fine, but after trying on a fresh environment, the process was hanging when loading autogluon.
Apparently, `scikit-optimize` doesn't support scikit-learn 0.23.0 yet, hence the issue. You may also want to enforce this in your setup.py as well.

frameworks/AutoGluon/exec.py
```diff
@@ -57,17 +60,8 @@ def run(dataset, config):
...
-        # Here, we're trying to identify to which label correspond the probabilities
-        # and from there we can build the probabilities for all labels.
-        classes = dataset.target.classes
-        prob_class = next((predictions[i] if p > 0.5 else next(c for c in classes if c != predictions[i])
-                           for i, p in enumerate(probabilities) if p != 0.5),
-                          classes[1])
-        probabilities = (np.array([[1-row, row] for row in probabilities]) if prob_class == classes[1]
-                         else np.array([[row, 1-row] for row in probabilities]))
+    probabilities = predictor.predict_proba(dataset=X_test, as_pandas=True, as_multiclass=True) if is_classification else None
+    prob_labels = probabilities.columns.values.tolist()
```
Innixma, line 63: This will break if is_classification == False, since probabilities will be None

sebhrusen:
running the validation benchmark on AWS, if everything's fine, I'm ready to merge.

frameworks/AutoGluon/requirements.txt
Innixma: Yep, I realized a bit ago. Just updated autogluon to 0.0.9 to fix.

Innixma:
> running the validation benchmark on AWS, if everything's fine, I'm ready to merge.

Sounds good, just had a few things I mentioned in my latest review that should be updated but shouldn't change the results from the validation benchmark run.

One thing I'm curious about is if you noticed any difference or issues with AutoGluon using the raw dataset.train.data compared to saving the data as csv and loading with task.Dataset()? I originally saved and reloaded to ensure types were not being set in any strange ways, but I think it should be fine considering its a numpy array originally. I mention this because the H2O exec actually does load from file: 
```
train = h2o.import_file(dataset.train.path, destination_frame=frame_name('train', config))
```

frameworks/AutoGluon/exec.py
```diff
@@ -0,0 +1,108 @@
...
+
+    perf_metric = metrics_mapping[config.metric] if config.metric in metrics_mapping else None
+    if perf_metric is None:
+        # TODO: figure out if we are going to blindly pass metrics through, or if we use a strict mapping
+        log.warning("Performance metric %s not supported.", config.metric)
+
+    is_classification = config.type == 'classification'
+    training_params = {k: v for k, v in config.framework_params.items() if not k.startswith('_')}
+
+    train = pd.DataFrame(dataset.train.data, columns=dataset.columns)
```
Innixma, line 40: You mentioned [here](https://github.com/openml/automlbenchmark/pull/105#discussion_r423299073) that
```
pd.DataFrame(dataset.train.data, columns=dataset.columns)
```
Returns a DataFrame of all object types, but that we can get the true types through an edit to `__init__` and then:

```
train = pd.DataFrame(dataset.train.data, columns=data.columns).astype(column_types)
```

Where now train has types float and object. This would be ideal for AutoGluon.

Otherwise, AutoGluon will treat the floats as objects (It will not override the types given to it), dramatically reducing overall performance. This was why I originally would save the DataFrame to csv and reload it with AutoGluon, as that infers the types automatically and converts the floats to float types instead of being objects.

frameworks/AutoGluon/exec.py
sebhrusen: good point, I should enable the regression dataset in the test benchmark...

frameworks/AutoGluon/exec.py
sebhrusen: hmm, there's possibly a flaw in the benchmark app then, that penalizes many sklearn-based frameworks...
The app only manipulates `numpy` arrays, so I can't set the type per column, and frameworks that can load files directly have this possibility.
If `amlb.data` module was exposing `pd.DataFrame` instead of `ndarray`, maybe this would globally tackle this concern for everyone..

sebhrusen: OK, I got a bit worried, but it's not that bad for sklearn-based frameworks (autosklearn, TPOT...): those are using encoded data which are typed as float.
So it's only a concern for frameworks that support categoricals, in which case I'll use the approach I recommended for mljar for now.

amlb/data.py
```diff
@@ -35,16 +35,16 @@ def __init__(self, index, name, data_type, values=None, has_missing_values=False
...
-        self.data_type = data_type
+        self.data_type = data_type.lower() if data_type is not None else None
         self.values = values
         self.has_missing_values = has_missing_values
         self.is_target = is_target
 
     def is_categorical(self, strict=True):
         if strict:
-            return self.data_type is not None and self.data_type.lower() in ['categorical', 'nominal', 'enum']
+            return self.data_type in ['categorical', 'nominal', 'enum']
```
Innixma, line 45: What about 'object' and 'str' types?

frameworks/AutoGluon/requirements.txt
sebhrusen: done now, thx

amlb/data.py
sebhrusen: we don't use 'object' type here: added a comment + creating a PR for those data types as they lack consistency and it's getting really confusing. Will probably try to use pandas DataFrame everywhere instead of numpy arrays.

For the current state of AutoGluon integration, I'll only convert numerical columns, the others will remain typed as `object`.
Is it fine for you?

Innixma: Seems fine to me, we don't do anything special beyond object, float, integer, and bool as input types.

sebhrusen: https://github.com/openml/automlbenchmark/issues/116

sebhrusen:
Results look good, I'm ready to merge.

## Test Results

### test benchmark
This includes one dataset for each problem type (binary, multiclass, regression), ran locally (macOS).

|    | id                | task        | framework   |   fold |    result | mode   | version   |   params | tag                  | utc                 |   duration |   models |   seed |   info |        acc |        auc |    logloss |   models_ensemble_count |   predict_duration |          r2 |    rmse |
|----|-------------------|-------------|-------------|--------|-----------|--------|-----------|----------|----------------------|---------------------|------------|----------|--------|--------|------------|------------|------------|-------------------------|--------------------|-------------|---------|
|  0 | openml.org/t/3913 | kc2         | AutoGluon   |      0 |  0.861472 | local  | 0.0.8     |      nan | feature/93/autogluon | 2020-05-13T17:40:42 |        9.5 |       11 |      1 |    nan |   0.792453 |   0.861472 | nan        |                       4 |          0.178664  | nan         | nan     |
|  1 | openml.org/t/59   | iris        | AutoGluon   |      0 |  0.570048 | local  | 0.0.8     |      nan | feature/93/autogluon | 2020-05-13T17:40:59 |       11.7 |       11 |      1 |    nan |   0.933333 | nan        |   0.570048 |                       2 |          0.0270348 | nan         | nan     |
|  2 | openml.org/t/2295 | cholesterol | AutoGluon   |      0 | 43.644    | local  | 0.0.8     |      nan | feature/93/autogluon | 2020-05-13T17:41:10 |        5.9 |        9 |      1 |    nan | nan        | nan        | nan        |                       4 |          0.107596  |   0.0167699 |  43.644 |

### validation benchmark

This includes various small or medium datasets with various corner cases.
Executed on a Ubuntu EC2 instance.

|    | id                  | task                    | framework   |   fold |   result | mode   | version   |   params | tag                  | utc                 |   duration |   models |       seed |   info |      acc |        auc |    logloss |   models_ensemble_count |   predict_duration |
|----|---------------------|-------------------------|-------------|--------|----------|--------|-----------|----------|----------------------|---------------------|------------|----------|------------|--------|----------|------------|------------|-------------------------|--------------------|
|  0 | openml.org/t/9910   | bioresponse             | AutoGluon   |      0 | 0.878819 | aws    | 0.0.9     |      nan | feature/93/autogluon | 2020-05-13T18:33:51 |      144.5 |       11 | 3227291290 |    nan | 0.808511 |   0.878819 | nan        |                       8 |           4.4397   |
|  1 | openml.org/t/125920 | dresses-sales           | AutoGluon   |      0 | 0.55665  | aws    | 0.0.9     |      nan | feature/93/autogluon | 2020-05-13T18:34:06 |        9.2 |       11 | 3227291290 |    nan | 0.62     |   0.55665  | nan        |                       7 |           0.304143 |
|  2 | openml.org/t/2079   | eucalyptus              | AutoGluon   |      0 | 0.718791 | aws    | 0.0.9     |      nan | feature/93/autogluon | 2020-05-13T18:34:30 |       18.1 |       11 | 3227291290 |    nan | 0.635135 | nan        |   0.718791 |                       6 |           0.42092  |
|  3 | openml.org/t/167125 | internet-advertisements | AutoGluon   |      0 | 0.986509 | aws    | 0.0.9     |      nan | feature/93/autogluon | 2020-05-13T18:38:59 |      161.2 |       11 | 3227291290 |    nan | 0.97561  |   0.986509 | nan        |                       5 |          17.5924   |
|  4 | openml.org/t/9950   | micro-mass              | AutoGluon   |      0 | 0.179956 | aws    | 0.0.9     |      nan | feature/93/autogluon | 2020-05-13T18:49:20 |      610.5 |        9 | 3227291290 |    nan | 0.965517 | nan        |   0.179956 |                       3 |           0.683948 |
|  5 | openml.org/t/3917   | kc1                     | AutoGluon   |      0 | 0.849075 | aws    | 0.0.9     |      nan | feature/93/autogluon | 2020-05-13T18:49:43 |       11.1 |       11 | 3227291290 |    nan | 0.890995 |   0.849075 | nan        |                       5 |           0.392232 |
|  6 | openml.org/t/168868 | APSFailure              | AutoGluon   |      0 | 0.994876 | aws    | 0.0.9     |      nan | feature/93/autogluon | 2020-05-13T18:57:35 |      394.1 |       11 | 3227291290 |    nan | 0.996184 |   0.994876 | nan        |                       4 |           0.915555 |

Innixma:
Looks good, thanks for all the work. Excited to start using this!

Issue #62, opened 2019-11-05
OpenML library upgrade


sebhrusen:

PGijsbers:
Does this mean raising the dependency from 0.7.0 to 0.10x?

sebhrusen:
last version I tried months ago was `0.9.0`, and there were some issues, so if `0.10.x` is available, we can try to switch directly to that one.

franchuterivera:
Hello! Is there any plan to move to a new opeml version alongside a new sklearn version? 
I ask because there are some new metrics in sklearn it would be nice to enable (Balanced accuracy).
I tried 0.10.2 and did not see any issues while moving, but please let me know!

sebhrusen:
Hey @franchuterivera, I'm definitely interested if you have a working branch (working with all current frameworks) with upgraded openml + sklearn.
Thanks!

franchuterivera:
ok, I have been running smoothly using:
https://github.com/openml/automlbenchmark/pull/129  --> based on 0.10.2

There are some frameworks  I don't run (like constant predictor, etc) so let me try them out and make sure... 

But at least with 0.10.2 I don't see a problem with the majority of frameworks. I will confirm shortly...

franchuterivera:
Please take a look at #129. I did not experience any problem while moving to the new openml benchmark version. Maybe some 0.9 problems that are now fixed.

All frameworks (except oboe) run smoothly. I need to debug (why oboe doesn't run) it on our end, but doesn't seem a problem with this transition.

Do notice that I am also fixing some typos I saw in the code and I am enabling balacc. I did not add it to the config but I would like to ask your opinion on that. Personally, i think balanced accuracy should be enabled. This means we have to tweak the frameworks for support on that. Let me know your thoughts

PR #129, opened 2020-05-23
Moving to new openml=0.10.2 version

franchuterivera:
Realted to OpenML library upgrade #62

### DIFF
--- a/amlb/results.py
+++ b/amlb/results.py
@@ -14,7 +14,7 @@ import numpy as np
 from numpy import nan, sort
 
 from .data import Dataset, DatasetType, Feature
-from .datautils import accuracy_score, confusion_matrix, f1_score, log_loss, mean_absolute_error, mean_squared_error, mean_squared_log_error, r2_score, roc_auc_score, read_csv, write_csv, is_data_frame, to_data_frame
+from .datautils import accuracy_score, confusion_matrix, f1_score, log_loss, balanced_accuracy_score, mean_absolute_error, mean_squared_error, mean_squared_log_error, r2_score, roc_auc_score, read_csv, write_csv, is_data_frame, to_data_frame
 from .resources import get as rget, config as rconfig, output_dirs
 from .utils import Namespace, backup_file, cached, datetime_iso, memoize, profile
 
@@ -319,7 +319,7 @@ class Result:
         if hasattr(self, metric):
             return getattr(self, metric)()
         # raise ValueError("Metric {metric} is not supported for {type}.".format(metric=metric, type=self.type))
-        log.warning("Metric %s is not supported for %s!", metric, type=self.type)
+        log.warning("Metric {} is not supported for {}!".format( metric, self.type))
         return nan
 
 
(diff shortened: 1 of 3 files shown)


amlb/results.py
```diff
@@ -319,7 +319,7 @@ def evaluate(self, metric):
         if hasattr(self, metric):
             return getattr(self, metric)()
         # raise ValueError("Metric {metric} is not supported for {type}.".format(metric=metric, type=self.type))
-        log.warning("Metric %s is not supported for %s!", metric, type=self.type)
+        log.warning("Metric {} is not supported for {}!".format( metric, self.type))
```
sebhrusen, line 322: FYI, this was not a typo, just a common practice to use the log formatter (here default printf-style) to avoid unnecessary string interpolation (e.g. if log level is not enabled).
I agree that the benefit is close to 0 here, but I mainly do this for consistency in all log statements.

amlb/results.py
```diff
@@ -357,8 +357,7 @@ def acc(self):
         return float(accuracy_score(self.truth, self.predictions))
 
     def balacc(self):
-        # return float(balanced_accuracy_score(self.truth, self.predictions))
-        pass
+        return float(balanced_accuracy_score(self.truth, self.predictions))
```
sebhrusen, line 360: yes, you can enable now that the `sklearn` version is supporting this.

> I did not add it to the config but I would like to ask your opinion on that.

I don't think it's necessary to add it to the default `config.yaml`.
What I should add there though, is the list of all supported metrics, as a comment to each `benchmark.metrics` entry, so that users can decide which metrics they can use:
1. as an objective function  for the automl frameworks.
2. as a result metric to compare the predictions.
I also need to improve documentation regarding those metrics, here is how it works:
```
metrics: 
    binary: ['auc', 'logloss', 'acc']`
    multiclass: ['logloss', 'acc']
    regression: ['rmse', 'r2']
```
means that the objective functions given to the frameworks are respectively `auc`, `logloss` and `rmse` (the first of each list).
Those objective metrics are also the metric used for the `result` metric column.
But ALL listed metrics (for a given problem) are computed from the predictions and available as additional information to compare the results.

> Personally, i think balanced accuracy should be enabled. This means we have to tweak the frameworks for support on that.

I don't know if many frameworks will support `balanced accuracy` as an objective function, most likely some of them would need to fallback to another metric, probably `accuracy`.
I just know that H2O's equivalent as an objective function (leading to same behaviour at least) would be `mean_per_class_error`.

amlb/results.py
franchuterivera: Sorry for the misunderstanding, but I am talking about the type argument. Specifically:
>>> import logging
>>> log = logging.getLogger()
>>> metric = 'bacloss'
>>> log.warning("Metric %s is not supported for %s!", metric, type='example')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/logging/__init__.py", line 1320, in warning
    self._log(WARNING, msg, args, **kwargs)
**TypeError: _log() got an unexpected keyword argument 'type'**

amlb/results.py
franchuterivera: understood, here I just enable it so that the users can overwrite the metrics through there config file (and balanced accuracy is an option, not I don't think it is the case)!

amlb/results.py
sebhrusen: oh right, my bad! Didn't even notice.  Good catch, thx!

sebhrusen:
merging it, I'll upgrade some reference framework versions (using sklearn directly) in a separate PR.
thx again

PR #138, opened 2020-05-28
KeyboardInterrupt should interrupt the entire benchmark

sebhrusen:
reviewed too broad usages of except (catching BaseException only to ensure that subprocesses can be cleaned).

ensure that one single  KBI allows to safely stop all running instances in AWS mode.

### DIFF
--- a/amlb/benchmark.py
+++ b/amlb/benchmark.py
@@ -420,7 +420,8 @@ class BenchmarkTask:
             result = ErrorResult(e)
         finally:
             self._dataset.release()
-            meta_result = meta_result or {}
-            meta_result['params'] = task_config.framework_params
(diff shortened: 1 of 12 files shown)


amlb/benchmark.py
```diff
@@ -420,7 +420,8 @@ def run(self, framework, framework_name):
             result = ErrorResult(e)
         finally:
             self._dataset.release()
-            meta_result = meta_result or {}
-            meta_result['params'] = task_config.framework_params
-            return results.compute_scores(framework_name, task_config.metrics, result=result, meta_result=meta_result)
```
sebhrusen, line 425: this nasty `return` in `finally` block was swallowing KBI

PR #141, opened 2020-05-29
Custom functions: e.g. custom metrics

sebhrusen:
adding generic support for frameworks extensions with an example of integration for custom metrics.

Extensions are loaded by default from `{user}/extensions.py`.
User can define as many extension files as he wants in his `config.yaml`, e.g.
```
extensions_files: 
   -  '{user}/autosklearn_extensions.py'
   -  'user}/autogluon_extensions.py'
   -  '{user}/tpot_extensions.py'
   - '{user}/extensions.py'
```

a specific variable, function or class defined in those files can be loaded by name from the framework integration using:
```
from frameworks.shared.callee import get_extension
...
ext = get_extension(config.extensions, 'foo')
```

The first name that could be loaded succesfully will be returned.
For example, if 'foo' is defined on all the extensions files above, and we're running `TPOT`, then the first 2 files may fail loading in TPOT integration if  they import `autogluon` or `autosklearn` modules, however loading `{user}/tpot_extensions.py` should succeed, so if there's a variable/function named "foo" there, it will be returned, otherwise it will look into the last file.

This PR shows how this can be used to inject custom metrics in frameworks like `autogluon`, `autosklearn` and `TPOT`.

### DIFF
--- a/amlb/benchmark.py
+++ b/amlb/benchmark.py
@@ -279,7 +279,7 @@ class TaskConfig:
 
     def __init__(self, name, fold, metrics, seed,
                  max_runtime_seconds, cores, max_mem_size_mb, min_vol_size_mb,
-                 input_dir, output_dir):
+                 input_dir, output_dir, extensions):
         self.framework = None
         self.framework_params = None
         self.type = None
@@ -295,6 +295,7 @@ class TaskConfig:
         self.input_dir = input_dir
         self.output_dir = output_dir
         self.output_predictions_file = os.path.join(output_dir, "predictions.csv")
+        self.extensions = extensions
 
     def __json__(self):
         return self.__dict__
@@ -350,6 +351,7 @@ class BenchmarkTask:
             min_vol_size_mb=task_def.min_vol_size_mb,
             input_dir=rconfig().input_dir,
             output_dir=benchmark.output_dirs.session,
+            extensions=rconfig().extensions_files,
         )
         # allowing to override some task parameters through command line, e.g.: -Xt.max_runtime_seconds=60
         if rconfig()['t'] is not None:
--- a/frameworks/AutoGluon/exec.py
+++ b/frameworks/AutoGluon/exec.py
@@ -12,7 +12,7 @@ from autogluon.task.tabular_prediction.tabular_prediction import TabularPredicti
 from autogluon.utils.tabular.utils.savers import save_pd, save_pkl
 import autogluon.utils.tabular.metrics as metrics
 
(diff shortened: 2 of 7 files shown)


sebhrusen:
TODO: add description to HOWTO documentation.

PGijsbers:
It looks like the metric is not reported in the results, so I set up #173. 
Other than that, as far as I can tell it looks good (only tested with TPOT though).
The documentation (the message in the PR) was good enough for me to set up my custom metric*.

\* I did run into a small issue where the amlb framework provides the `y` as (N,1)-vector, which the [TPOT custom metric example](https://epistasislab.github.io/tpot/using/#scoring-functions) could not deal with. The subsequent error from tpot (incorrect data) was then a bit puzzling considering the baked-in scikit-learn metrics work (presumably because they strip the second dimension automatically if required.

PGijsbers:
comment by seb from #173:
> There's a difficulty here. If we take TPOT as an example, then the training and the result processing are not done on the same process/venv, this means that if the tpot_extensions.py has a dependency on tpot module, then we will be able to load the extension in the training process, but it will fail in the one processing results...
> 
> The lookup logic suggested in #141 is then probably too simple.
> Maybe we need something like:
> 
> extensions_files: 
>    autosklearn:  '{user}/autosklearn_extensions.py'
>    autogluon:  '{user}/autogluon_extensions.py'
>    tpot: ['{user}/tpot_extensions.py', '{user}/sklearn_extensions.py']
>    default:  '{user}/extensions.py'
> 
> Now let's imagine that the tpot_extensions.py can be only be loaded in the tpot venv. Then, we can still provide extensions for the results that only require sklearn: to report the custom metrics in the results, the requirement is that it relies only on sklearn, numpy, pandas, scipy..., that is the dependencies available in the main process.
> 
> I think you can still merge this PR, I will probably consider this to improve #141 though.

Is this a thing we expect to worry about in practice? I suppose automl packages could have their own metrics predefined to import, but if that is the case a scikit-learn fallback would not be enough either.

sebhrusen:
@PGijsbers 
> I suppose automl packages could have their own metrics predefined to import, but if that is the case a scikit-learn fallback would not be enough either.

I was trying to make the distinction between the custom metric used by the framework for optimization (this one may be provided by the automl package), and the one that we compute ex post from the `predictions.csv`, and this one we should be able to implement it only with the core libraries, it's just a math formula after all.

PGijsbers:
> for this one the user should be able to implement it only with the core libraries, it's just a math formula after all.

Exactly. And this is what the current implementation already facilitates as far as I am aware?

PR #152, opened 2020-07-27
fix matplotlib config to run with basic python on macOS

sebhrusen:
@pplonski a small fix for macOS. I had the change locally, forgot to mention.


frameworks/mljarsupervised/exec.py
```diff
@@ -4,11 +4,12 @@
 
 import pandas as pd
 import matplotlib
+matplotlib.use("agg")  # no need for tk
```
sebhrusen, line 7: needs to be called before importing supervised module

Issue #43, opened 2019-07-17
Add support for AILib (aka. ML-Plan)


sebhrusen:
New OS framework to be added:
https://github.com/fmohr/AILibs

looking for contributors.

mwever:
Hi,

I now integrated ML-Plan with the AutoML Benchmark. More specifically, I added two default frameworks for ML-Plan:

- **MLPlanWEKA**: ML-Plan using WEKA as a basis for configuring ML algorithms. It is capable of tackling classification as well as regression tasks. However, we did not benchmark it yet for regression. For internal validation, MLPlanWEKA offers acc, auc, logloss, and f1 for classification, and rmse, mse, mae, rmsle, and r2 for regression.
- **MLPlanSKLearn**: ML-Plan using scikit-learn as a basis for configuring ML algorithms. It is also capable of dealing with classification and regression tasks. Metrics are the same for classification and regression as for MLPlanWEKA.

I have only one question left: How is the value of the positive class provided for AUC and F1 in case of binary classification?

PR #153, opened 2020-07-28
Integration of ML-Plan as a default framework

mwever:
With this PR, ML-Plan is introduced as a new default framework. More specifically, there are two versions of ML-Plan added:
- MLPlanWEKA which denotes the version of ML-Plan working with WEKA as a backend
- MLPlanSKLearn which denotes the version of ML-Plan working with scikit-learn as a backend.

Both versions should be capable of working with classification and regression tasks for various metrics (acc, logloss, auc, f1, rmse, mse, rmsle, mae, mse) depending on the respective task, i.e. it is for example not possible to run ML-Plan on a multi-class classification dataset with auc nor f1.

Furthermore, I increased the version number of liac-arff in the root requirements.txt file since the already listed openml dependency required the newer version of liac-arff.

Sumary of Changes:
- Add default framework MLPlanWEKA
- Add default framework MLPlanSKLearn
- Edited documentation for ML-Plan
- Increased version number of liac-arff dependency


mwever:
#43

resources/frameworks.yaml
```diff
@@ -69,6 +69,14 @@ AutoWEKA:
   version: '2.6'
   project: https://www.cs.ubc.ca/labs/beta/Projects/autoweka/
 
+MLPlanWEKA:
+  version: 'latest'
+  project: https://mlplan.org
+
+MLPlanSKLearn:
```
sebhrusen, line 76: Just wondering, I didn't compare the 2 integrations line by line, but they look very similar to me. The main difference I see in the `setup` is that `sklearn` version requires of course Python dependencies. 
Wouldn't be easier for future maintenance to have only one integration?
They would then share the same `setup.sh` (downloading sklearn even in Weka mode should be harmless).
And you could just switch between modes in the framework definition, for example:
```yaml
MLPlan:
  module: frameworks.MLPlan
  version: latest
  project: https://mlplan.org

MLPlanWEKA:
    extends: MLPlan
    params:
        _mode: weka

MLPlanSklearn:
    extends: MLPlan
    params:
        _mode: sklearn
```
in the shared `exec.py`, you can then use:
```python
mode = config.framework_params.get('_mode')  # the leading underscore in the param is by convention for params not passed directly to the framework itself
```
to switch some logic between both implementations.

Again, this makes sense only if you think it would simplify integration and avoid duplicate logic.

frameworks/MLPlanSKLearn/exec.py
```diff
@@ -0,0 +1,67 @@
...
+    train_file = dataset.train.path
+    test_file = dataset.test.path
+    # Weka requires target as the last attribute
+    if dataset.target.index != len(dataset.predictors):
+        train_file = reorder_dataset(dataset.train.path, target_src=dataset.target.index)
+        test_file = reorder_dataset(dataset.test.path, target_src=dataset.target.index)
+
+    training_params = {k: v for k, v in config.framework_params.items() if not k.startswith('_')}
+
+    cmd_root = "java -jar {here}/lib/mlplan/mlplan-cli*.jar ".format(here=dir_of(__file__))
```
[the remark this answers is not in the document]
sebhrusen: Is it that you don't care because the Java process then delegates most work to sklearn?

requirements.txt
```diff
@@ -1,5 +1,5 @@
 boto3>=1.9,<2.0
-liac-arff==2.3
+liac-arff==2.4
```
sebhrusen, line 2: thx for that, if openml requires it as well, maybe I should think about relaxing the version constraint.

docs/automl_overview.md
```diff
@@ -214,15 +214,15 @@ Komer, Brent, James Bergstra, and Chris Eliasmith (2014).
...
+[documentation](https://starlibs.github.io/AILibs/projects/mlplan/) |
 Java |
-Optimization: Hierachical Task Planning | -
+Optimization: Best-First Search on a search graph induced through Hierachical Task Network Planning | -
 
 > a new approach to AutoML based on hierarchical planning
 
-ML-Plan is a new AutoML framework based on Hierarchical Task Planning.
-It can optimize with both a WEKA and scikit-learn backend.
+ML-Plan organizes the search space of possible AutoML solution candidates via Hierarchical Task Network (HTN) planning.
```
PGijsbers, line 224: Thanks for updating the docs! But I think it should be ML instead of AutoML?

docs/automl_overview.md
```diff
@@ -214,15 +214,15 @@ Komer, Brent, James Bergstra, and Chris Eliasmith (2014).
 *ICML workshop on AutoML 2014*.
 
 ## ML-Plan
-[source](https://github.com/fmohr/AILibs/tree/master/softwareconfiguration/mlplan) |
-[documentation](https://github.com/fmohr/AILibs/tree/dev/softwareconfiguration/mlplan/docs/javadoc) |
+[source](https://github.com/starlibs/AILibs) |
+[documentation](https://starlibs.github.io/AILibs/projects/mlplan/) |
 Java |
-Optimization: Hierachical Task Planning | -
+Optimization: Best-First Search on a search graph induced through Hierachical Task Network Planning | -
```
PGijsbers, line 220: The '-' field should hold the license, the repo lists  AGPL-3.0. If that is correct, could you update that here?

frameworks/MLPlanWEKA/exec.py
```diff
@@ -0,0 +1,67 @@
...
+def run(dataset: Dataset, config: TaskConfig):
+    log.info("\n**** ML-Plan for WEKA ****\n")
+
+    # Mapping of benchmark metrics to Weka metrics
+    metrics_mapping = dict(
+        acc='ERRORRATE',
+        auc='AUC',
+        logloss='LOGLOSS',
+	f1='F1',
+	rmse="ROOT_MEAN_SQUARED_ERROR",
```
PGijsbers, line 23: Can you keep consistent indentation both here and below when defining `cmd_params`?

PGijsbers:
@sebhrusen Marcel told me he was not able to evaluate on aws. Would you be able to run a validation test?

frameworks/MLPlanSKLearn/exec.py
```diff
@@ -0,0 +1,67 @@
...
+    cmd_params = dict(
+        f='"{}"'.format(train_file),
+        p='"{}"'.format(test_file),
+        t=config.max_runtime_seconds,
+        ncpus=config.cores,
+        l=metric,
+	    s=config.seed,   # weka accepts only int16 as seeds
+	    ooab=config.output_predictions_file,
```
sebhrusen, line 50: is this predictions file already in the expected format?
Usually, we ask the frameworks to provide the predictions and probabilities + various flags to generate the predictions file in a standard way. 
see for example:
```python
    return result(output_file=config.output_predictions_file,
                  predictions=predictions,
                  truth=y_test,
                  probabilities=probabilities,
                  probabilities_labels=prob_labels,
                  target_is_encoded=False,
                  models_count=num_models_trained,
                  models_ensemble_count=num_models_ensemble,
                  training_duration=training.duration,
                  predict_duration=predict.duration)
```
If you provide `probabilities_labels` like in the snippet above, the probabilities columns will be automatically reordered to meet sklearn expectations when computing metrics like AUC.

frameworks/MLPlanSKLearn/exec.py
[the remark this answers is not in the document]
mwever: Yes ML-Plan will spawn additional processes for running python pipelines which are not subsumed under the memory restrictions of Java. Nontheless, I added now some constraints to allow the Java part read in larger datasets as well.

resources/frameworks.yaml
mwever: Thanks for the pointer. I will re-arrange the include of the framework definition

frameworks/MLPlanWEKA/exec.py
mwever: sorry, of course I will make it consistent.

frameworks/MLPlanSKLearn/setup.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
mwever: I am re-arranging the setup making sure to use a virtual environment for ML-Plan

frameworks/MLPlanSKLearn/requirements.txt
```diff
@@ -0,0 +1,7 @@
+liac-arff==2.4
+numpy>=1.15,<2.0
+scipy>=1.5,<1.6
+pandas>=0.23,<1.0
+tpot>=0.11.0,<0.12
+xgboost>=1.1.0,<1.2
+scikit-learn>=0.22.2
```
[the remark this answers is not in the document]
mwever: How can I then reorder the dataset (as it is also done for AutoWEKA) such that the target is in the last column?

frameworks/MLPlanSKLearn/exec.py
mwever: I extended the CLI of ML-Plan to conform the output format as required by the benchmark which will produce a csv file where the first columns contain the probabilities of the respective class labels in case of classification, the predictions, and the truth (the latter two are written out for both tasks classification and regression). However, there was no information about the number of models trained etc. I am currently extending the output so that we can use the result function.

frameworks/MLPlanSKLearn/requirements.txt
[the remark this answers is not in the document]
sebhrusen: > How can I then reorder the dataset (as it is also done for AutoWEKA) such that the target is in the last column?

Oh I see, we've got a problem here.
When running in a venv, your code can use any function from `amlb.utils` module because this one has minimmal dependencies (only yaml), but by it can't use `reorder_dataset` because this function is in `amlb.datautils`, this last module depending on `pandas`, `sklearn` and so on, so I didn't want to have all those dependencies by default in all venvs of course. 
It's possible to do the reordering in the `__init__.py` though, before calling the subprocess.
Let me try a couple a things, I'll come back to you with a solution (I hope).

frameworks/MLPlanSKLearn/exec.py
sebhrusen: great, let me verify that you can call the `result` function even with `predictions=None` for example, otherwise I'll fix it.

mwever: You dont need to check that. I already added another output format and called the result in the ordinary way. So no worries.

frameworks/MLPlanSKLearn/requirements.txt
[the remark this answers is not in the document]
mwever: That'd be great!

frameworks/MLPlan/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+
+    data = dict(
+        train=dict(path=dataset.train.path),
+        test=dict(path=dataset.test.path),
+        target=dict(index=dataset.target.index),
+        predictors_type=['Numerical' if p.is_numerical() else 'Categorical' for p in dataset.predictors]
+    )
+
+    return run_in_venv(__file__, "exec.py",
+                       input_data=data, dataset=dataset, config=config)
```
sebhrusen, line 22: This should solve your reordering issue:
```suggestion
from amlb.benchmark import TaskConfig
from amlb.data import Dataset
from amlb.resources import config as rconfig
from amlb.datautils import reorder_dataset
from amlb.utils import call_script_in_same_dir


def setup(*args, **kwargs):
    call_script_in_same_dir(__file__, "setup.sh", rconfig().root_dir, *args, **kwargs)


def run(dataset: Dataset, config: TaskConfig):
    from frameworks.shared.caller import run_in_venv

    train_path = dataset.train.path
    test_path = dataset.test.path
    mode = config.framework_params.get('_mode')
    # Weka requires target as the last attribute
    if mode == 'weka' and dataset.target.index != len(dataset.predictors):
        train_path = reorder_dataset(dataset.train.path, target_src=dataset.target.index)
        test_path = reorder_dataset(dataset.test.path, target_src=dataset.target.index)

    data = dict(
        train=dict(path=train_path),
        test=dict(path=test_path),
        target=dict(index=dataset.target.index),
    )

    return run_in_venv(__file__, "exec.py",
                       input_data=data, dataset=dataset, config=config)
```

frameworks/MLPlan/exec.py
```diff
@@ -0,0 +1,87 @@
...
+    metric = metrics_mapping[config.metric] if config.metric in metrics_mapping else None
+    if metric is None:
+	    raise ValueError('Performance metric {} is not supported.'.format(config.metric))
+    
+    train_file = dataset.train.path
+    test_file = dataset.test.path
+    # Weka requires target as the last attribute
+    if dataset.target.index != len(dataset.predictors):
+        train_file = reorder_dataset(dataset.train.path, target_src=dataset.target.index)
+        test_file = reorder_dataset(dataset.test.path, target_src=dataset.target.index)
```
sebhrusen, line 36: you can now remove this here as the reordering has been done in the parent process
```suggestion
 ```

File: frameworks/MLPlan/exec.py [hunk 12]
```diff
@@ -0,0 +1,87 @@
...
+    return result(
+	    output_file=config.output_predictions_file,
+		predictions=predictions,
+		truth=y_test,
+		probabilities=probabilities,
+		probabilities_labels=probabilities_labels,
+		target_is_encoded=is_classification,
+		models_count=models_count,
+        training_duration=training.duration
+    )
```
sebhrusen, line 84: created fix (https://github.com/openml/automlbenchmark/pull/154), you'll need to rebase.
```suggestion
    return result(
        models_count=models_count,
        training_duration=training.duration
    )
```
If you can't provide `models_count`, no problem, just put `1`: it's purely informative currently, afaik we don't use it for analysis.

frameworks/MLPlan/setup.sh
```diff
@@ -0,0 +1,24 @@
...
+pip3 install --no-cache-dir -r $HERE/requirements.txt
+
+MLPLAN_ARC="mlplan.zip"
+DOWNLOAD_DIR="$HERE/lib"
+TARGET_DIR="$DOWNLOAD_DIR/mlplan"
+
+if [[ ! -e "$TARGET_DIR" ]]; then
+    mkdir -p $DOWNLOAD_DIR
+    echo "Download ML-Plan from extern"
+    wget http://127.0.0.1/mlplan/version/$VERSION -O $DOWNLOAD_DIR/$MLPLAN_ARC
```
sebhrusen, line 21: `127.0.0.1` is not going to work for everyone ;)

PGijsbers: On the contrary, it works even without a network connection!

sebhrusen: premise: I have a local build
proposition: all local builds can be downloaded
conclusion: I can download the build

mwever: Sorry, on my working laptop I have Windows as an OS and thus needed to transfer changes to a Linux machine. I will change the URL asap again.

PGijsbers: Don't worry, it was some lighthearted fun :) these things slip in occasionally

sebhrusen:
@mwever I have another fix in `master` coming for you soon: had to change the PATH for the subprocess so that it can call `tpot` installed in the virtual env.

mwever:
> @mwever I have another fix in `master` coming for you soon: had to change the PATH for the subprocess so that it can call `tpot` installed in the virtual env (in case you're already investigating this)

I am currently working on a maybe related issue. So currently we spawn new python processes calling `python3`. As far as I can see we would need to call the python binary within the venv folder (what we currently do not do, which is probably also the reason why the required dependencies are not available).  Am I on the right track regarding this issue? And is your fix related to this problem?

sebhrusen:
I think so it's related, I got `MLPlanSklearn` working with my fix.
I basically dynamically had `venv/bin` at the head the PATH environement variable that is used to spawn the first subprocess.
Then, all the subprocesses you will spawn will inherit this PATH by default, so it should resolve all your `python3` + `tpot` issues.

mwever:
Okay this would be awesome. But I fear that its not solely a python3 + tpot issue. If one tries to run it in docker mode, the dependencies are also missing.

sebhrusen:
will also try docker mode soon and confirm

sebhrusen:
one point: I noticed that MLPlan is generating a lot of stuff under `./tmp` folder.
Is it possible for you to change this destination?
1. use `os.TmpDir`
2. use `output_subdir('tmp')`

note that the app allows to download those generated artifacts if you want when running in AWS mode, see autogluon/autosklearn/h2oautoml for examples.

mwever:
I can make it possible to change the destination. But we dont need to make the artifacts downloadable since these artifacts are only temporary files necessary to run the scikit-learn candidates.
I think I will also make the python binary configurable so that we have a backup if the PATH strategy is not working. Is this information available in the utils (in exec.py)?

sebhrusen:
if you don't plan to download anything, please use `os.TmpDir` then, as everything under a folder created by `output_subdir` in downloaded by default...
I'm testing `docker` mode now, had to apply minor changes that I will add as suggestios, but otherwise the `PATH` fix seems to work.
so with the PATH fix in master, you should be good after a rebase and the small remaining changes.

Note, I also got a few of those:
```
Error class: java.util.concurrent.ExecutionException
    Error message: org.api4.java.ai.graphsearch.problem.pathsearch.pathevaluation.PathEvaluationException: Error in evaluating node!
    Error trace:
        ai.libs.jaicore.timing.TimedComputation.compute(TimedComputation.java:105)
        ai.libs.jaicore.search.algorithms.standard.bestfirst.nodeevaluation.RandomCompletionBasedNodeEvaluator.evaluateTimeouted(RandomCompletionBasedNodeEvaluator.java:259)
        ai.libs.jaicore.search.algorithms.standard.bestfirst.nodeevaluation.TimeAwareNodeEvaluator.lambda$evaluate$1(TimeAwareNodeEvaluator.java:66)
```
but they don't fail the process, so they may just be warnings.
If there are issues with multithreading, you may want disable parallelism, some frameworks use the `_njobs` param for that to ignore the default `ncpus` enforcement.

frameworks/MLPlan/exec.py
```diff
@@ -0,0 +1,100 @@
...
+    train_file = dataset.train.path
+    test_file = dataset.test.path
+
+    training_params = {k: v for k, v in config.framework_params.items() if not k.startswith('_')}
+    backend = config.framework_params.get('_backend', 'weka')
+
+    if backend == "weka":
+        mem_limit = str(max(config.max_mem_size_mb-1024,2048))
+    else:
+        mem_limit = str(max((config.max_mem_size_mb-1024) / config.cores,2048))
```
sebhrusen, line 41: in `-Xmx{val}M` , val  must be an integer
```suggestion
    if backend == "weka":
        mem_limit = str(max(config.max_mem_size_mb-1024, 2048))
    else:
        mem_limit = str(max(round((config.max_mem_size_mb-1024) / config.cores), 2048))
```

frameworks/MLPlan/setup.sh
```diff
@@ -0,0 +1,28 @@
...
+HERE=$(dirname "$0")
+VERSION=${2:-"latest"}
+
+echo "Setup ML-Plan for version $VERSION"
+
+# creating local venv
+. $HERE/../shared/setup.sh $HERE
+
+if [[ -x "$(command -v apt-get)" ]]; then
+    echo "setup system packages"
```
sebhrusen, line 11: for some reason, it failed downloading openjdk without this update, which I don't completely understand because there's already an update before installing python, I think it's a docker thing, each `RUN` command being somehow isolated. Note that we do the same for H2OAutoML and AutoWeka, which also require a jdk.
```suggestion
    echo "setup system packages"
    SUDO apt-get update
```

mwever:
> if you don't plan to download anything, please use `os.TmpDir` then, as everything under a folder created by `output_subdir` in downloaded by default...
> I'm testing `docker` mode now, had to apply minor changes that I will add as suggestions, but otherwise the `PATH` fix seems to work.
> so with the PATH fix in master, you should be good after a rebase and the small remaining changes.
> 
> Note, I also got a few of those:
> 
> ```
> Error class: java.util.concurrent.ExecutionException
>     Error message: org.api4.java.ai.graphsearch.problem.pathsearch.pathevaluation.PathEvaluationException: Error in evaluating node!
>     Error trace:
>         ai.libs.jaicore.timing.TimedComputation.compute(TimedComputation.java:105)
>         ai.libs.jaicore.search.algorithms.standard.bestfirst.nodeevaluation.RandomCompletionBasedNodeEvaluator.evaluateTimeouted(RandomCompletionBasedNodeEvaluator.java:259)
>         ai.libs.jaicore.search.algorithms.standard.bestfirst.nodeevaluation.TimeAwareNodeEvaluator.lambda$evaluate$1(TimeAwareNodeEvaluator.java:66)
> ```
> 
> but they don't fail the process, so they may just be warnings.
> If there are issues with multithreading, you may want disable parallelism, some frameworks use the `_njobs` param for that to ignore the default `ncpus` enforcement.
> 
> docker results:
> 
> ```
>                   id         task      framework constraint fold     result   metric    mode version                   params                  utc  duration models        seed       acc       auc   logloss        r2     rmse
> 0  openml.org/t/3913          kc2  MLPlanSKLearn       test    0   0.852814      auc  docker   0.2.3  {'_backend': 'sklearn'}  2020-07-29T20:46:36      53.1     10  2882908699  0.773585  0.852814  0.361292       NaN      NaN
> 1  openml.org/t/3913          kc2  MLPlanSKLearn       test    1   0.882035      auc  docker   0.2.3  {'_backend': 'sklearn'}  2020-07-29T20:47:42      65.4     10  2882908700  0.886792  0.882035  0.319335       NaN      NaN
> 2    openml.org/t/59         iris  MLPlanSKLearn       test    0   1.098610  logloss  docker   0.2.3  {'_backend': 'sklearn'}  2020-07-29T20:49:09      86.3     10  2882908699  0.333333       NaN  1.098610       NaN      NaN
> 3    openml.org/t/59         iris  MLPlanSKLearn       test    1   1.147280  logloss  docker   0.2.3  {'_backend': 'sklearn'}  2020-07-29T20:50:34      84.1     10  2882908700  0.000000       NaN  1.147280       NaN      NaN
> 4  openml.org/t/2295  cholesterol  MLPlanSKLearn       test    0  43.475800     rmse  docker   0.2.3  {'_backend': 'sklearn'}  2020-07-29T20:51:17      42.8     10  2882908699       NaN       NaN       NaN  0.024334  43.4758
> 5  openml.org/t/2295  cholesterol  MLPlanSKLearn       test    1  56.160900     rmse  docker   0.2.3  {'_backend': 'sklearn'}  2020-07-29T20:52:04      46.6     10  2882908700       NaN       NaN       NaN -0.093724  56.1609
> ```

Some ExecutionExceptions are expected to occur. So without a more detailed exception stacktrace it is hard to tell what the problem is there. So in ML-Plan for example we do not make sure that the learning algorithm will converge with the chosen hyperparameter values. This is one potential reason for this fail. But there might also be other misconfigurations. For some configurations it might be the case that it is an invalid configuration. However, these fails should be not that expensive as these configurations usually crash immediately.

mwever:
Status update from my side:
- I configured ML-Plan such that it will now use os.TmpDir for storing temporary files.
- I merged the original repository's master into the fork's master to include the new requirements.txt and the PATH variable fix
- I merged your suggested changes for the apt-get update as well as ensuring that configuring the RAM for the JVM is done with an integer value by rounding

frameworks/MLPlan/exec.py
```diff
@@ -0,0 +1,101 @@
...
+        f='"{}"'.format(train_file),
+        p='"{}"'.format(test_file),
+        t=config.max_runtime_seconds,
+        ncpus=config.cores,
+        l=metric,
+        m=mode,
+        s=config.seed,   # weka accepts only int16 as seeds
+        ooab=predictions_file,
+        os=statistics_file,
+        tmp=os.TmpDir,
```
sebhrusen, line 65: sorry, I must have been confusing (and confused myself at the same time): `os.TmpDir` is not  in core `os` module, but in `utils.os` module...
And this is a context manager: compared with the core `tempfile` module, this context cleans after itself, which is mainly useful when running frameworks generating gazillions of artifacts locally:
```python
with utils.os.TmpDir as tmp_dir:
    cmd_params = dict(
        ...
        tmp=tmp_dir,
        **training_params
    )
    ...
    with utils.Timer() as training:
        utils.run_cmd(cmd, _live_output_=True)

```

mwever: So when I try to use utils.os.TmpDir I obtain the error message that os has no attribute 'TmpDir'. I am not sure whether the system's os module is maybe overriding the one of the utils package...

sebhrusen: My bad, I should have tried it before mentioning it: `utils` doesn't expose submodules themselves, only their content.
So the right syntax is `utils.TmpDir`.
I tried it locally and seems to work when used as a context manager as above, let me add a suggestion inline.

frameworks/MLPlan/exec.py
```diff
@@ -0,0 +1,101 @@
...
+        ooab=predictions_file,
+        os=statistics_file,
+        tmp=utils.os.TmpDir,
+        **training_params
+    )
+
+    cmd = cmd_root + ' '.join(["-{} {}".format(k, v) for k, v in cmd_params.items()])
+
+    with utils.Timer() as training:
+        utils.run_cmd(cmd, _live_output_=True)
```
sebhrusen, line 72: tried this locally, should work:
```suggestion
   with utils.TmpDir() as tmp_dir:
        cmd_root = "java -jar -Xmx{mem_mb}M {here}/lib/mlplan/mlplan-cli*.jar ".format(here=os.path.dirname(__file__), mem_mb=mem_limit)
        cmd_params = dict(
            f='"{}"'.format(train_file),
            p='"{}"'.format(test_file),
            t=config.max_runtime_seconds,
            ncpus=config.cores,
            l=metric,
            m=mode,
            s=config.seed,   # weka accepts only int16 as seeds
            ooab=predictions_file,
            os=statistics_file,
            tmp=tmp_dir,
            **training_params
        )

        cmd = cmd_root + ' '.join(["-{} {}".format(k, v) for k, v in cmd_params.items()])

        with utils.Timer() as training:
            utils.run_cmd(cmd, _live_output_=True)
```

sebhrusen: I ran the above after reinstalling ML plan, see the `-tmp` param:
```
Called ML-Plan CLI with the following params: >[-f, /Users/seb/.openml/cache/org/openml/www/datasets/61/dataset_train_0.arff, -p, /Users/seb/.openml/cache/org/openml/www/datasets/61/dataset_test_0.arff, -t, 60, -ncpus, 2, -l, LOGLOSS, -m, sklearn, -s, 312602755, -ooab, /Users/seb/repos/ml/automlbenchmark/results/mlplansklearn_test_test_local_20200807T145352/mlplan_out/iris/0/predictions.csv, -os, /Users/seb/repos/ml/automlbenchmark/results/mlplansklearn_test_test_local_20200807T145352/mlplan_out/iris/0/statistics.json, -tmp, /tmp/tmp2j03htfv]<
```
The only small annoyance is that the good old `./tmp/model_dumps` folder is still created, although remains empty.

sebhrusen: I reliazed the core library already provides the same feature (I reinvented the wheel, and will remove it in a next PR): https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory
So, you can use `with tempfile.TemporaryDirectory() as tmp_dir:` instead...

mwever: Adapted the exec.py to work with the tempfile. The /tmp/model_dumps issue is also fixed now.

mwever:
So I think from my side the integration of ML-Plan is all done by now. I resolved the issue with the tmp folder still popping up and also adapted the tmpfolder in the python script to work with python's native method of retrieving a tmp folder.

resources/frameworks.yaml
```diff
@@ -69,6 +69,21 @@ AutoWEKA:
   version: '2.6'
   project: https://www.cs.ubc.ca/labs/beta/Projects/autoweka/
 
+MLPlan:
+  module: frameworks.MLPlan
+  version: '0.2.3'
+  project: https://mlplan.org
```
sebhrusen, line 75: just noticed that this URL currently points to the Reisebuch from the author :)
Apparently, it's not the case with the `http` version which redirects correctly.
```suggestion
  project: http://mlplan.org
```

sebhrusen:
Tested all modes, and the validation benchmark, integration is working fine.
Just found an issue (#161) that I fixed separately as it's on `amlb` side.
Just adding a proper test option before merging that issue first, and then I'll merge this one.

mwever:
Hugh...difficult question. Here is why:

Since WEKA is in Java too, it is the more natural backend for ML-Plan: Better control over resources allocated to single candidates easier management since we do not need to interface with Python.

scikit-learn also has its advantages. Mainly, because it is sometimes much faster to evaluate even though we have an extra mile to go for the evaluation itself. So eventually, ML-Plan can explore much more in scikit-learn than compared to WEKA. However, evaluating the single candidates is not always as stable as it should be, so there is a higher danger for invalid runs (not adhering to the candidate evaluation timeouts; not adhering to resource limits, since killing sub-processes may fail due to unknown reasons).

To keep Auto-WEKA a little company in the big table of AutoML tools, it might be nice to go for the WEKA flavor.

Issue #61, opened 2019-11-05
Support for OpenML Studies


sebhrusen:
As mentioned in https://github.com/openml/automlbenchmark/issues/17 and https://github.com/openml/automlbenchmark/pull/57, now that the benchmark definition file format is very simple, we could create a small script converting [OpenML Studies](https://docs.openml.org/#studies-under-construction) into a benchmark definition to make it easier to manage definitions and maybe as well to upload results.

PGijsbers:
@ja-thomas did you ever manage to upload the full list of tasks we considered previously? If so, what is the study id?

The studies I created are 218 (study - collection of tasks), and 226 (suite - a collection of runs).

PR #155, opened 2020-07-29
Allow an OpenML suite to define tasks of benchmark

PGijsbers:
When specifying the benchmark to run any integer name is interpreted as an openml study.
E.g. `python runbenchmark.py constantpredictor openml/s/218`.

An unnecessary amount of meta-data is still downloaded, which is (mostly) due to openml tasks not having the associated dataset name. 

I'm looking for feedback. For example:
 - Allowing a study id instead of benchmark name seems natural to me, but do you foresee problems? The only thing I can think of is 
 - I think I found the best place to do the conversion, but is there a better spot?
 - I am not entirely sure why the `Namespace` are used, but I simply adopted it. Does this seem reasonable?

Related to #61. However uses the study directly instead of writing to intermediate file. 

If approved I'll add/update docs.

### DIFF
--- a/amlb/benchmarks/file.py
+++ b/amlb/benchmarks/file.py
@@ -1,9 +1,8 @@
 import logging
 import os
-from typing import List
+from typing import List, Tuple, Optional
 
-from amlb.utils import config_load
-
+from amlb.utils import config_load, Namespace
 
 log = logging.getLogger(__name__)
 
@@ -26,9 +25,10 @@ def _find_local_benchmark_definition(name: str, benchmark_definition_dirs: List[
             name, benchmark_definition_dirs))
 
 
-def load_file_benchmark(name: str, benchmark_definition_dirs: List[str]):
+def load_file_benchmark(name: str, benchmark_definition_dirs: List[str]) -> Tuple[str, Optional[str], List[Namespace]]:
+    """ Loads benchmark from a local file. """
     benchmark_file = _find_local_benchmark_definition(name, benchmark_definition_dirs)
     log.info("Loading benchmark definitions from %s.", benchmark_file)
     tasks = config_load(benchmark_file)
     benchmark_name, _ = os.path.splitext(os.path.basename(benchmark_file))
-    return benchmark_name, tasks
+    return benchmark_name, benchmark_file, tasks
--- a/amlb/benchmarks/file.py
+++ b/amlb/benchmarks/file.py
@@ -14,15 +14,13 @@ def _find_local_benchmark_definition(name: str, benchmark_definition_dirs: List[
         return name
 
     for bd in benchmark_definition_dirs:
-        bf = os.path.join(bd, "{}.yaml".format(name))
+        bf = os.path.join(bd, f"{name}.yaml")
         if os.path.exists(bf):
             # We don't account for duplicate definitions (yet).
             return bf
 
     # should we support s3 and check for s3 path before raising error?
-    raise ValueError(
-        "Incorrect benchmark name or path `{}`, name not available in {}.".format(
-            name, benchmark_definition_dirs))
+    raise ValueError(f"Incorrect benchmark name or path `{name}`, name not available in {benchmark_definition_dirs}.")
 
 
 def load_file_benchmark(name: str, benchmark_definition_dirs: List[str]) -> Tuple[str, Optional[str], List[Namespace]]:
--- a/amlb/benchmarks/file.py
+++ b/amlb/benchmarks/file.py
@@ -1,0 +1,34 @@
+import logging
+import os
+from typing import List
+
+from amlb.utils import config_load
+
+
+log = logging.getLogger(__name__)
+
+
+def _find_local_benchmark_definition(name: str, benchmark_definition_dirs: List[str]) -> str:
+    # 'name' should be either a full path to the benchmark,
+    # or a filename (without extension) in the benchmark directory.
+    if os.path.exists(name):
+        return name
+
+    for bd in benchmark_definition_dirs:
+        bf = os.path.join(bd, "{}.yaml".format(name))
+        if os.path.exists(bf):
+            # We don't account for duplicate definitions (yet).
+            return bf
+
+    # should we support s3 and check for s3 path before raising error?
+    raise ValueError(
+        "Incorrect benchmark name or path `{}`, name not available in {}.".format(
+            name, benchmark_definition_dirs))
+
+
+def file_benchmark_to_tasks(name: str, benchmark_definition_dirs: List[str]):
+    benchmark_file = _find_local_benchmark_definition(name, benchmark_definition_dirs)
+    log.info("Loading benchmark definitions from %s.", benchmark_file)
+    tasks = config_load(benchmark_file)
+    benchmark_name, _ = os.path.splitext(os.path.basename(benchmark_file))
+    return tasks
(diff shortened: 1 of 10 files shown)


PGijsbers:
Should I also overload the task specification parameter `-t` to directly run on a task? (Obviously both the benchmark framework and AutoML system must support the task type for a successful run)

runbenchmark.py
```diff
@@ -15,8 +15,8 @@
 parser.add_argument('framework', type=str,
                     help="The framework to evaluate as defined by default in resources/frameworks.yaml.")
 parser.add_argument('benchmark', type=str, nargs='?', default='test',
-                    help="The benchmark type to run as defined by default in resources/benchmarks/{benchmark}.yaml "
-                         "or the path to a benchmark description file. Defaults to `%(default)s`.")
+                    help="The benchmark type to run as defined by default in resources/benchmarks/{benchmark}.yaml, "
+                         "a path to a benchmark description file, or an openml suite ID. Defaults to `%(default)s`.")
```
sebhrusen, line 19: suggestion what about using sth similar to the nifty openml urls:
```bash
# to run the entire openml suite:
python runbenchmark.py gama openml/s/42 

# to run a single openml task (without the need for a yaml definition)
python runbenchmark.py gama openml/t/42 
```
this leaves this parameter open for future extensions, for example to provide easy support for kaggle datasets:
```
python runbenchmark.py gama kaggle/some/dataset
```

sebhrusen: ideally, in future versions, for consistency we could do the same for the benchmark definitions.
Instead of 
```
- name: Airlines
  openml_task_id: 189354
```
have
```
- name: Airlines
   dataset: openml/t/189354

- name: CreditCardFraud
   dataset: kaggle/mlg-ulb/creditcardfraud
```
to be aligned with current support for external files, cf. https://github.com/openml/automlbenchmark/blob/master/examples/custom/benchmarks/filedatasets.yaml

amlb/resources.py
```diff
@@ -165,34 +166,48 @@ def _constraints(self):
...
+            study = openml.study.get_suite(study_id)
+            oml_tasks = [openml.tasks.get_task(tid, download_data=False) for tid in study.tasks]
+            oml_datasets = [openml.datasets.get_dataset(t.dataset_id, download_data=False) for t in oml_tasks]
+            tasks = [Namespace(name=d.name, description=d.description, openml_task_id=t.id) for t, d in zip(oml_tasks, oml_datasets)]
+            hard_defaults = Namespace()
+            benchmark_file = "study_{}".format(study_id)
+        else:
+            benchmark_dir = self.config.benchmarks.definition_dir
+            if not isinstance(benchmark_dir, list):
+                benchmark_dir = [benchmark_dir]
```
sebhrusen, line 190: see my suggestion below, what about decoupling the parsing/task extraction logic in a separate module, so that we can extend easily to support other datasources in the future?

runbenchmark.py
PGijsbers: Just pushed some changes that works towards this refactoring. It's a bit slower than I'm used to since I'm still getting the hang of the infrastructure.

sebhrusen: no one here is in a hurry, are you?
As you started to refactor this, could you please move your new files into a submodule, in a way similar to the `datasets` submodule. I'd suggest sth like `definitions` or `argsparser`,  exposing only your `benchmark_load` for now.

sebhrusen: > Should I also overload the task specification parameter -t to directly run on a task? (Obviously both the benchmark framework and AutoML system must support the task type for a successful run)

Just noticed this comment.
If you follow my previous suggestion, you shouldn't to override this because `-t` doesn't make any sense when calling:
```
python runbenchmark.py gama openml/t/42 
```

If you were asking if we should allow
```
python runbenchmark.py gama openml/s/226 -t helena
```
I think it's necessary anyway for parallelism.
User triggers the entire suite with `python runbenchmark.py gama openml/s/226`, but each ec2 instance for example will run only something like `python runbenchmark.py gama openml/s/226 -t helena -f 3`.

PGijsbers: Both of these options work already (automatically 👍 )

amlb/benchmarks/openml.py
```diff
@@ -0,0 +1,51 @@
...
+    if len(benchmark.split('/')) == 3:
+        domain, oml_type, oml_id = benchmark.split('/')
+        supported_types = ['s', 't']
+
+        valid_id = False
+        try:
+            _ = int(oml_id)
+            valid_id = True
+        except ValueError:
+            pass
```
sebhrusen, line 23:
```suggestion
valid_id = oml_id.isdecimal()
```
should do the job

amlb/benchmarks/openml.py
```diff
@@ -0,0 +1,51 @@
...
+
+        return domain == "openml" or oml_type in supported_types and valid_id
```
sebhrusen, line 25: why do you have a `or` here?
should it not be all `and`?

amlb/benchmarks/file.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def _find_local_benchmark_definition(name: str, benchmark_definition_dirs: List[str]) -> str:
+    # 'name' should be either a full path to the benchmark,
+    # or a filename (without extension) in the benchmark directory.
+    if os.path.exists(name):
+        return name
+
+    for bd in benchmark_definition_dirs:
+        bf = os.path.join(bd, "{}.yaml".format(name))
```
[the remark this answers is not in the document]
PGijsbers: Ah, I saw this used before and just assumed we were at Py3.5+. Good to know!

docs/README.md
```diff
@@ -57,12 +57,18 @@ Optional: create a Python3 virtual environment.
...
 _Therefore, we rather suggest one of the method below:_ 
 
-using venv:
+using venv on Linux/macOS:
 ```bash
 python3 -m venv ./venv
 source venv/bin/activate
 # remember to call `deactivate` once you're done using the application
 ```
+using venv on Windows:
```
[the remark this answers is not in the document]
PGijsbers: Local runs generally don't work (though e.g. constant predictor does).

docs/HOWTO.md
```diff
@@ -244,6 +244,19 @@ Then the datasets can be declared in the benchmark definition file as follow:
 - the `folds` attribute is also optional but recommended for those datasets as the default value is `folds=10` (default amount of folds in openml datasets), so if you don't have that many folds for your custom datasets, it is better to declare it explicitly here.
 - Remote files are downloaded to the `input_dir` folder and archives are decompressed there as well, so you may want to change the value of this folder in your [custom config.yaml file](#custom-configuration) or specify it at the command line with the `-i` or `--indir` argument (by default, it points to the `~/.openml/cache` folder).
 
+#### OpenML suites
```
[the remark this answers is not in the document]
PGijsbers: They were renamed, a suite is a list of tasks (i.e. defines a benchmark, but does not contain results), while a study is the collection of results from running flows (i.e. some machine learning script/pipeline) on a suite. I'll double check this with the openml team and then verify I don't use those terms interchangeably here.

amlb/benchmarks/openml.py
PGijsbers: I wasn't sure - the numbers need to be integer, otherwise the openml-python api will raise an error later. I suppose this is wrong too since it does not enforce non-zero positive integers. Will rewrite.

amlb/benchmarks/openml.py
PGijsbers: good catch

PGijsbers:
Thanks for the help!

PR #144, opened 2020-06-09
Version agnostic framework setup

sebhrusen:
Trying to make the framework setup as agnostic as possible regarding versions and rely only on the version declared in `frameworks.yaml`.

First, this avoids mistakes when upgrading libraries.
Second, users can then easily benchmark different versions of the same framework simply by creating multiple entries in their local `frameworks.yaml` and then either create a docker image for each of them or use AWS mode, or force the setup (`--setup=force`) before each run.

Example:
```yaml
TPOT:
  version: '0.11.5'
  description: the current version defined in resources/framworks.yaml

TPOT_10:
   version: '0.10.2'
   description: an older version

TPOT_latest:
   description: no version specified, will use master branch

TPOT_mybranch:
   version: 'mybranch'
   description: will test mybranch  
```

no version or `version: latest` points to git `master` branch for most Python frameworks.
A notable exception is H2OAutoML (as it would require to compile the entire project), for which `latest` points to latest stable release, and `nightly` points to latest nightly build.

Also use this occasion to upgrade all major frameworks.

PR #159, opened 2020-07-31
update mljar-supervised version to 0.6.0

pplonski:
Updating mljar-supervised version to 0.6.0. Is there an option to use the latest version from github or pypi?

### DIFF
--- a/resources/frameworks.yaml
+++ b/resources/frameworks.yaml
@@ -114,7 +114,7 @@ TPOT:
 #    verbosity: 2
 
 mljarsupervised:
-  version: '0.5.5'
+  version: '0.6.0'
   project: https://github.com/mljar/mljar-supervised
   params:
     mode: "Compete"   # set mode for Compete, default mode is Explain


PGijsbers:
Yes, it is possible to specify the version as "latest" ([docs](https://github.com/openml/automlbenchmark/blob/master/docs/HOWTO.md#framework-definition)). But we prefer a fixed version by default so that it's clear on which version of the packages the benchmark is run and we can ensure it is stable and compatible with the current version of AMLB. This way results are reproducible when running a specific commit of the AMLB. 

It should be easy to run the benchmark against the latest version of a package but I can't find it in the docs (presumably through #144 @sebhrusen)?

We could alternatively also have the frameworks at "latest", and just generate a snapshot benchmark definition every few months with fixed versions. What would you think about that @sebhrusen?

sebhrusen:
@pplonski @PGijsbers for now, it's more reasonable to require a fixed version, user can define a local definition using the latest version as described informally with `autosklearn` in the Framework definitions example: https://github.com/openml/automlbenchmark/blob/master/docs/HOWTO.md#framework-definition

> We could alternatively also have the frameworks at "latest", and just generate a snapshot benchmark definition every few months with fixed versions.

I like the idea, it would also allow users to (re)run older benchmarks easily as we would add a version of `frameworks.yaml` per quarter let's say. Although, for a minimum of stability and as the frameworks are always read from `frameworks.yaml` by default, I would keep the default `frameworks.yaml` with stable versions. Would you be ok with that?
```
frameworks.yaml  # current stable versions
frameworks_latest.yaml  # all frameworks set to latest (may be pulled directly from `master` branch)
frameworks_2020_Q1.yaml
frameworks_2020_Q2.yaml
...
```

So to use latest for example, user would have to run for example
```
python runbenchmark.py resources/frameworks_latest.yaml#mljarsupervised
```
but the default
```
python runbenchmark.py mljarsupervised
```
would still use the default and stable version from `frameworks.yaml`.
I can create a ticket for that, as you can see, if we want to make the switch easy, it requires some changes in how the `framework` parameter is handled to select the alternative `yaml` definition: either by overloading the syntax on the current param (as we do now for benchmarks), or by adding a new param. Personally, I tend to prefer overloading in this situation...

Well, just realized, it may already work using:
```
python runbenchmark.py mljarsupervised -Xframeworks.definition_file=resources/frameworks_latest.yaml
```
I would need to try it.

sebhrusen:
@pplonski no API change with `0.6.0` right? I guess you tried it and I can trust you on this, this would same me some time, thx!

pplonski:
@sebhrusen yes, no API changes

sebhrusen:
Also moved the discussion regarding `latest` versions and version history to https://github.com/openml/automlbenchmark/issues/160

Issue #157, opened 2020-07-29
Error when using shared/setup.sh


paxcema:
Hi,

I am trying to add a framework to the benchmark suite. However, when doing `python runbenchmark.py framework validation` my setup.sh calls the shared/setup.sh and the latter fails at line 42 with the error `.../venv/bin/pip: No such file or directory`. I already tried using a shorter (54 characters, without white spaces) path, as it seems there can be errors related to that as well.

I also tried running `python runbenchmark.py TPOT validation`, and got the same error, which suggests the issue is not related to my setup.sh file.

Any ideas as to what it might be?

sebhrusen:
Hi @paxcema, do you really get 3 dots in `.../venv/bin/pip: No such file or directory.`
On which OS are you?
from the `automlbenchmark` app folder, can you please run the following command and tell me the output?
```
cd frameworks/shared; SHARED_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd -P)"; APP_ROOT=$(dirname $(dirname "$SHARED_DIR")); echo "shared=$SHARED_DIR"; echo "root=$APP_ROOT"; cd ../..
```

sebhrusen:
sorry, wrong command, this one is closer to what is effectively done:
```
SHARED_DIR="$(cd $(dirname "frameworks/shared/setup.sh") && pwd -P)"; APP_ROOT=$(dirname $(dirname "$SHARED_DIR")); echo "shared=$SHARED_DIR"; echo "root=$APP_ROOT"
```

paxcema:
Hi @sebhrusen!

No, to be precise the exact error in the .sh is: `/home/pcerdam/mdb/amlb/frameworks/TPOT/../shared/setup.sh: line 42: /home/pcerdam/mdb/amlb/frameworks/TPOT/venv/bin/pip: No such file or directory`. 

The .py error is `subprocess.CalledProcessError: Command '/home/pcerdam/mdb/amlb/frameworks/TPOT/setup.sh /home/pcerdam/mdb/amlb 0.11.5' returned non-zero exit status 127.`

The output to that second command is:

```
shared=/home/pcerdam/mdb/amlb/frameworks/shared
root=/home/pcerdam/mdb/amlb
```

sebhrusen:
> I am trying to add a framework to the benchmark suite

Ok, I guess you're not trying to add `TPOT` again, so I don't see why it's calling `amlb/frameworks/TPOT/setup.sh`.

[...]

PR #161, opened 2020-08-17
Reencode truth and predictions to match labels lexicographic order

sebhrusen:
When framework was providing the combination:
```python
prediction_labels=["b", "a"]
target_encoded=True
```
with therefore "b" encoded as 0, "a" encoded as 1.

Then the `amlb` app was reordering prediction columns to get ["a", "b"], but it was not reencoding predictions and truth values, so that 1 was then meaning "b" instead of "a" after the reordering.

This PR fixes this issues, and ensures that values are re-encoded if needed.

### DIFF
--- a/docs/HOWTO.md
+++ b/docs/HOWTO.md
@@ -611,11 +611,11 @@ Using the instructions above:
  1. try to setup the framework: 
     > python runbenchmark.py myframework -s only
  1. fixes the framework setup until it works: the setup being usually a simple `setup.sh` script, you should be able to test it directly without using the application.
- 1. try to run simple test against one fold using defaults (`test` benchmark and `test` constraints):
-    > python runbenchmark.py myframework -f 0
+ 1. try to run simple test against one fold using defaults (`test` benchmark and `test` constraints) with the `-Xtest_mode` that will trigger additional validations:
+    > python runbenchmark.py myframework -f 0 -Xtest_mode
  1. fix the module integration code until the test produce all results with no error (if the integration generated an error, it is visible in the results).
  1. if this works, validate it against the `validation` dataset using one fold:
-    > python runbenchmark.py myframework validation 1h4c -f 0
+    > python runbenchmark.py myframework validation 1h4c -f 0 -Xtest_mode
  1. if this works, try to run it in docker to validate the docker image setup: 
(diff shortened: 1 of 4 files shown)


sebhrusen:
discovered when running `validation` benchmarks on https://github.com/openml/automlbenchmark/pull/153

This was clearly visible on tasks like `APSFailure` (AUC < 0.5), or `micro-mass` (highest probability was not matching prediction).
And then also visible on fold 1 for `iris` on `test` benchmark.

Will create an Issue so that we have a proper test or a validation option that would verify that the `predictions.csv` are correct, which means for example that for classification, we should always have 
prediction value = column with highest probability.

PGijsbers:
Why was it that in the original paper the reported AUC for APSFailure was `>0.5`? ([figure](https://openml.github.io/automlbenchmark/results.html))

sebhrusen:
@PGijsbers, just updated my comment: this was occuring only for MLPlan framework, this is because MLPlan was using a combination of parameters (the ones in this PR description) that was not used by other frameworks until now.

docs/HOWTO.md
```diff
@@ -909,13 +909,27 @@ _Examples of method duration info when using this custom profiling_:
...
+The setup is created for Debian-based linux environments, and macOS (most frameworks can be installed on macOS, ideally with `brew` installed, but there may be a few exceptions), so it may work with other Linux environments not listed above (e.g. Debian, Ubuntu 20.04, ...).
+The best way to run benchmarks on non-supported OS, is to use the docker mode.
+
+If the setup fails on a supported environment, please try the following:
+- force the setup: see above.
+- ensure that the same framework is not set up multiple times in parallel on the same machine:
+  - first use `python runbenchmark.py MyFramework -s only` on one terminal.
+  - then you can trigger multiple `python runbenchmark.py MyFramework ...` (without `-s` option) in parallel.
+- delete the `lib` and `venv` folders, if present, under the given framework folder (e.g. `frameworks/MyFramework`), and try the setup again.
+
```
sebhrusen, line 931: not related with this PR, but I was discussing https://github.com/openml/automlbenchmark/issues/157 at the same time, to including this here.

PGijsbers (approved):

PR #174, opened 2020-10-13
proposal for test layout

sebhrusen:
- added namespace to differentiate unit tests from future integration/functional tests.
- grouping unit tests into packages matching the tested module.
- moved dedicated resources in package using them.
- moved `pytest.ini` to the root + add testpaths to allow running tests directly from the root dir using `python -m pytest`


PGijsbers:
> added namespace to differentiate unit tests from future integration/functional tests.

I agree this would probably otherwise happen in the future anyway 👍  thanks

> grouping unit tests into packages matching the tested module.

That was originally the case... until I extracted the `framework_definitions` module from the `resource` module 😓 

>  moved dedicated resources in package using them.

👍 

> moved pytest.ini to the root + add testpaths to allow running tests directly from the root dir using python -m pytest

👍

Issue #175, opened 2020-10-22
Potential "datasets" module collision with HuggingFace's Transformers


paxcema:
Hello,

I'm trying to integrate the MindsDB framework to the benchmark suite. 

One of MindsDB's dependencies is HuggingFace's Transformer library, which seems to cause a namespace collision. When I try to import MindsDB in `exec.py` this happens:

```
File "/automlbenchmark/frameworks/MindsDB/exec.py", line 5, in <module>

    from mindsdb import Predictor

  File "/automlbenchmark/frameworks/MindsDB/venv/lib/python3.7/site-packages/mindsdb/__init__.py", line 33, in <module>

    from mindsdb_native import *

  File "/automlbenchmark/frameworks/MindsDB/venv/lib/python3.7/site-packages/mindsdb_native/__init__.py", line 7, in <module>

    import lightwood

  File "/automlbenchmark/frameworks/MindsDB/venv/lib/python3.7/site-packages/lightwood/__init__.py", line 11, in <module>

    from lightwood.api.predictor import Predictor

  File "/automlbenchmark/frameworks/MindsDB/venv/lib/python3.7/site-packages/lightwood/api/predictor.py", line 10, in <module>

    from lightwood.api.data_source import DataSource

  File "/automlbenchmark/frameworks/MindsDB/venv/lib/python3.7/site-packages/lightwood/api/data_source.py", line 15, in <module>

    from lightwood.encoders import (

  File "/automlbenchmark/frameworks/MindsDB/venv/lib/python3.7/site-packages/lightwood/encoders/__init__.py", line 6, in <module>

    from lightwood.encoders.text.infersent import InferSentEncoder

  File "/automlbenchmark/frameworks/MindsDB/venv/lib/python3.7/site-packages/lightwood/encoders/text/__init__.py", line 3, in <module>

    from lightwood.encoders.text.distilbert import DistilBertEncoder

  File "/automlbenchmark/frameworks/MindsDB/venv/lib/python3.7/site-packages/lightwood/encoders/text/distilbert.py", line 5, in <module>

    from transformers import (

  File "/automlbenchmark/frameworks/MindsDB/venv/lib/python3.7/site-packages/transformers/__init__.py", line 22, in <module>

    from .integrations import (  # isort:skip

  File "/automlbenchmark/frameworks/MindsDB/venv/lib/python3.7/site-packages/transformers/integrations.py", line 58, in <module>

    from .file_utils import is_torch_tpu_available

[...]

PR #176, opened 2020-10-23
Remove relative or unused imports, extra newlines

PGijsbers:
It allows submodules to be imported, which avoids #175 behavior.

Also removed unused imports and any superfluous newlines at the end of files.

@sebhrusen  I was a bit confused about `amlb.dateutils` importing [sklearn metrics](https://github.com/openml/automlbenchmark/blob/74e532211ef119e279ce6453bb0f5daa5c2f7910/amlb/datautils.py#L17) which were unused in the module itself. They were then subsequently imported from `amlb.dateutils` by [`amlb.results`](https://github.com/openml/automlbenchmark/blob/74e532211ef119e279ce6453bb0f5daa5c2f7910/amlb/results.py#L18). I assume just importing from `sklearn.metrics` in the `amlb.results` module itself doesn't hurt anything? And that the old import is just some legacy stuff.

### DIFF
--- a/amlb/__init__.py
+++ b/amlb/__init__.py
@@ -2,14 +2,14 @@
 amlb entrypoint package.
 """
 
-from .logger import app_logger as log
-from .errors import AutoMLError
-from .resources import Resources
-from .benchmark import Benchmark, SetupMode
-from .docker import DockerBenchmark
-from .singularity import SingularityBenchmark
-from .aws import AWSBenchmark, AWSRemoteBenchmark
-from .results import TaskResult
+from amlb.logger import app_logger as log
+from amlb.errors import AutoMLError
+from amlb.resources import Resources
+from amlb.benchmark import Benchmark, SetupMode
+from amlb.docker import DockerBenchmark
+from amlb.singularity import SingularityBenchmark
+from amlb.aws import AWSBenchmark, AWSRemoteBenchmark
+from amlb.results import TaskResult
 
 __all__ = [
     "log",
--- a/amlb/aws.py
+++ b/amlb/aws.py
@@ -31,13 +31,13 @@ from urllib.parse import quote_plus as uenc
 import boto3
 import botocore.exceptions
 
-from .benchmark import Benchmark, SetupMode
-from .datautils import read_csv, write_csv
-from .docker import DockerBenchmark
-from .job import Job
-from .resources import config as rconfig, get as rget
-from .results import ErrorResult, Scoreboard, TaskResult
-from .utils import Namespace as ns, datetime_iso, file_filter, flatten, list_all_files, normalize_path, str_def, tail, touch
+from amlb.benchmark import Benchmark, SetupMode
+from amlb.datautils import read_csv, write_csv
+from amlb.docker import DockerBenchmark
+from amlb.job import Job
+from amlb.resources import config as rconfig, get as rget
+from amlb.results import ErrorResult, Scoreboard, TaskResult
+from amlb.utils import Namespace as ns, datetime_iso, file_filter, flatten, list_all_files, normalize_path, str_def, tail, touch
 
 
 log = logging.getLogger(__name__)
@@ -1051,4 +1051,3 @@ class AWSRemoteBenchmark(Benchmark):
 
     def _upload_results(self):
         pass
-
--- a/amlb/benchmark.py
+++ b/amlb/benchmark.py
@@ -15,12 +15,12 @@ import logging
 import math
 import os
 
-from .job import Job, SimpleJobRunner, MultiThreadingJobRunner, ThreadPoolExecutorJobRunner, ProcessPoolExecutorJobRunner
-from .datasets import DataLoader, DataSourceType
-from .data import DatasetType
-from .resources import get as rget, config as rconfig, output_dirs as routput_dirs
-from .results import ErrorResult, Scoreboard, TaskResult
-from .utils import Namespace as ns, OSMonitoring, as_list, datetime_iso, flatten, lazy_property, profile, repr_def, \
+from amlb.job import Job, SimpleJobRunner, MultiThreadingJobRunner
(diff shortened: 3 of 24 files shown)


sebhrusen:
@PGijsbers  did you verify that it's fixing #175 ? I'm not convinced that switching from absolute to relative imports will fix anything.
I rather think that the issue here lies in the fact that `frameworks/shared/caller.py` is modifying  the `PYTHONPATH` for the subprocess which runs in its own `venv`. This was done to allow the subprocess to use the `amlb.utils` (see `frameworks/shared/callee.py`) without duplicating code and without trying to load all the dependencies (amlb.utils requires minimal dependencies).

By changing this PYTHONPATH, it makes all `amlb` submodules + subpackages available to the subprocess, creating potential collision, that's most likely what's happening in #175.

Probably a better approach would be to remove the `amlb` package from the `PYTHONPATH`, but load the `amlb.utils` package in `callee.py` manually (see https://docs.python.org/3.8/library/importlib.html#importing-a-source-file-directly).
If we keep relative imports in all `utils` submodules, it should be able to load them directly (not sure about this, it may require a custom importer, idk) when `utils` is imported in `callee.py`.

sebhrusen:
> @sebhrusen I was a bit confused about amlb.dateutils importing sklearn metrics which were unused in the module itself. They were then subsequently imported from amlb.dateutils by amlb.results. I assume just importing from sklearn.metrics in the amlb.results module itself doesn't hurt anything? And that the old import is just some legacy stuff.

Yeah, I just wanted to have only one module depending on `sklearn` to see  clearly how much `amlb` depends on it, hence the aliases there used exclusively in `results`. It is coming from a time where framework's `venv` were not supported yet, and I was not sure I would not get rid  of dependency on `sklearn` completely.

PGijsbers:
> @PGijsbers did you verify that it's fixing #175 ? 

As far as I am aware the current work for @paxcema is not public, so I was not able to run the exact code. I did test that `datasets` can be imported as long as it is in the path, regardless of where you execute from.
i.e. on `master`:
```python
>>> import sys; sys.path.extend(["automlbenchmark", "automlbenchmark/amlb"]); import datasets
  File "<stdin>", line 1, in <module>
  File "automlbenchmark/amlb\datasets\__init__.py", line 3, in <module>
    from .file import FileLoader
  File "automlbenchmark/amlb\datasets\file.py", line 11, in <module>
    from ..data import Dataset, DatasetType, Datasplit, Feature
ValueError: attempted relative import beyond top-level package
```
but on `fix_#175`:
```python
>>> import sys; sys.path.extend(["automlbenchmark", "automlbenchmark/amlb"]); import datasets
>>>
```

The second part of the bug arises from the fact that the Transformer package only checks for `ImportError`, while the previous import failed with a `ValueError`. Because the import should no longer fail, this should be avoided. The [Transformer package then checks](https://github.com/huggingface/transformers/blob/023f0f3708f73e4fdffb92505296cd7d3d928aef/src/transformers/file_utils.py#L87-L98) for specific attributes of the `datasets` packages (specifically `__version__`  and `load_datasets`) which our dataset module does not have. The Transformer package will then ignore the imported datasets package correctly. At least, that's how I imagine it will work out, but if @paxcema can verify this, that would be great.

> Probably a better approach would be to remove the amlb package from the PYTHONPATH, but load the amlb.utils package in callee.py manually (see https://docs.python.org/3.8/library/importlib.html#importing-a-source-file-directly).
If we keep relative imports in all utils submodules, it should be able to load them directly (not sure about this, it may require a custom importer, idk).

I am not convinced. Specifically the functions which are imported to `callee` do not rely on any relative imports as far as I can tell. Perhaps moving them to modules which do not rely on other parts of the framework (at least `Namespace` probably warrants it own file?) would be even better. (I am also a bit scared of introducing the added complexity of non-standard imports, especially if it entails a custom importer.)

paxcema:
I just verified @PGijsbers, it does work! The framework is now successfully imported.

sebhrusen:
@PGijsbers 
>  I did test that datasets can be imported as long as it is in the path, regardless of where you execute from.

I'm not sure we should allow that, somehow, I'm happy that your example failed:
```>>> import sys; sys.path.extend(["automlbenchmark", "automlbenchmark/amlb"]); import datasets```
the path manipulation is a hack that should usually be avoided, and allowing such imports doesn't prevent future collisions, worse it may make them harder to detect as shown with `Transformer` logic.

PGijsbers:
Wasn't path manipulation what was used in the first place to allow for the imports (by [`caller.py`](https://github.com/openml/automlbenchmark/blob/74e532211ef119e279ce6453bb0f5daa5c2f7910/frameworks/shared/caller.py#L62))? The think the real issue here is that `automlbenchmark/amlb` is in the path, which is internal to the directory itself, and thus allows for naming conflicts with any (first-level) submodule of `amlb`.

It should be removed and only [this](https://github.com/openml/automlbenchmark/blob/74e532211ef119e279ce6453bb0f5daa5c2f7910/frameworks/shared/callee.py#L13) import statement should remain (minus the fact that perhaps e.g. namespace should be in its own submodule). This requires some 'outdated' frameworks (`oboe`, `RandomForest`, `TunedRandomForest`, `GAMA`, `MLJar` and custom stacking example - they all use `Timer`). If that's fixed, `import datasets` should from the callee should raise an `ImportError`.

sebhrusen:
> Wasn't path manipulation what was used in the first place to allow for the imports (by caller.py)? The think the real issue here is that automlbenchmark/amlb is in the path, which is internal to the framework itself, and thus allows for naming conflicts with any (first-level) submodule of amlb.

yes, that's what I was mentioning in my first comment.
Suggested fix: https://github.com/openml/automlbenchmark/pull/178

PGijsbers:
I'll go ahead and close this. It's going to be easier incorporate the formatting fixes into a separate pr.

PR #177, opened 2020-10-23
Fix module loading

sebhrusen:
module name was forced to lowercase by default, preventing correct loading

```
python runbenchmark.py autogluon
Running `autogluon` on `test` benchmarks in `local` mode.
...
No module named 'frameworks.autogluon'
  File "runbenchmark.py", line 111, in <module>
    bench = amlb.Benchmark(args.framework, args.benchmark, args.constraint)
  File "/Users/seb/repos/ml/automlbenchmark/amlb/benchmark.py", line 78, in __init__
    self.framework_module = import_module(self.framework_def.module)
  File "/Users/seb/.pyenv/versions/3.7.6/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 965, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'frameworks.autogluon'
```

### DIFF
--- a/amlb/framework_definitions.py
+++ b/amlb/framework_definitions.py
@@ -60,7 +60,7 @@ def _add_and_normalize_names(frameworks: Namespace):
     framework_names = dir(frameworks)
     for name in framework_names:
         framework = frameworks[name]
-        framework.name = name.lower()
+        framework.name = name
         if name.lower() != name:
             del frameworks[name]
             frameworks[name.lower()] = framework
--- a/amlb/framework_definitions.py
+++ b/amlb/framework_definitions.py
@@ -56,10 +56,8 @@ def _sanitize_and_add_defaults(frameworks, resource):
 
 
 def _add_framework_name(frameworks: Namespace):
-    """ Converts each framework definition to lowercase and adds a 'name' field. """
-    framework_names = dir(frameworks)
-    for name in framework_names:
-        framework = frameworks[name]
+    """ Adds a 'name' attribute to each framework. """
+    for name, framework in frameworks:
         framework.name = name
 
 
--- a/amlb/framework_definitions.py
+++ b/amlb/framework_definitions.py
@@ -36,7 +36,7 @@ def _load_and_merge_framework_definitions(frameworks_file: Union[str, List[str]]
 
 def _sanitize_definitions(frameworks: Namespace):
     """ Normalize names, add name field, remove invalid extensions. """
-    _add_and_normalize_names(frameworks)
+    _add_framework_name(frameworks)
     _remove_frameworks_with_unknown_parent(frameworks)
     _remove_self_reference_extensions(frameworks)
 
@@ -55,17 +55,12 @@ def _sanitize_and_add_defaults(frameworks, resource):
     _add_defaults_to_frameworks(frameworks, resource)
 
 
-def _add_and_normalize_names(frameworks: Namespace):
+def _add_framework_name(frameworks: Namespace):
     """ Converts each framework definition to lowercase and adds a 'name' field. """
     framework_names = dir(frameworks)
     for name in framework_names:
         framework = frameworks[name]
         framework.name = name
-        if name.lower() != name:
-            del frameworks[name]
-            frameworks[name.lower()] = framework
-        if "extends" in framework:
-            framework.extends = framework.extends.lower()
 
 
 def _add_default_module(framework, config):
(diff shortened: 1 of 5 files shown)


amlb/framework_definitions.py
```diff
@@ -60,7 +60,7 @@ def _add_and_normalize_names(frameworks: Namespace):
     framework_names = dir(frameworks)
     for name in framework_names:
         framework = frameworks[name]
-        framework.name = name.lower()
+        framework.name = name
         if name.lower() != name:
             del frameworks[name]
             frameworks[name.lower()] = framework
```
sebhrusen, line 66: we only need to use lowercase here: for the lookup dictionary

tests/unit/amlb/resources/test_framework_definition.py
```diff
@@ -0,0 +1,33 @@
...
+
+def test_framework_definition_lookup_is_case_insensitive():
+    res = NS(
+        _frameworks=NS(
+            lower=NS(name="lower"),
+            UPPER=NS(name="UPPER"),
+            Camel=NS(name="CaMel")
+        )
+    )
+    # binding `framework_definition` method to our resource mock: use pytest-mock instead?
```
sebhrusen, line 14: we will most probably need `mock` or `pytest-mock` very soon, but keeping this for a future PR

sebhrusen: Namespace appears very useful to create mock objects here :)

amlb/framework_definitions.py
```diff
@@ -55,17 +55,12 @@ def _sanitize_and_add_defaults(frameworks, resource):
...
     framework_names = dir(frameworks)
     for name in framework_names:
         framework = frameworks[name]
-        framework.name = name.lower()
-        if name.lower() != name:
-            del frameworks[name]
-            frameworks[name.lower()] = framework
-        if "extends" in framework:
-            framework.extends = framework.extends.lower()
+        framework.name = name
```
PGijsbers, line 63: (I can't make a github suggestion because it involves deleted and added lines...)
The docstring needs to be updated too, and the code can now be simplified (since we use the original name).

```python
def _add_framework_name(frameworks: Namespace):
    """ Adds a 'name' field to each framework. """
    for name, framework in frameworks:
        framework.name = name
```

At this point we might also consider removing the separate function call instead. 
But I suppose it is nice we maintain more isolated tests by keeping the function separate as is.

amlb/resources.py
```diff
@@ -81,7 +81,8 @@ def framework_definition(self, name):
         :param name:
         :return: name of the framework as defined in the frameworks definition file
         """
-        framework = self._frameworks[name.lower()]
+        lname = name.lower()
+        framework = next((f for n, f in self._frameworks if n.lower() == lname), None)
```
PGijsbers, line 85: Just thinking out loud here, would
```suggestion
        framework = {n.lower(): f for n, f in self._frameworks}.get(lname)
```
be simpler? I am not convinced.
(I realize in terms of optimization it is worse, but given the context it should be optimized for devs not compute)

tests/unit/amlb/resources/test_framework_definition.py
PGijsbers: Any particular reason for not using @pytest.mark.parametrize so each condition is checked (and can fail) separately? Similar to other tests.

PGijsbers (changes requested):

amlb/resources.py
sebhrusen: next is better here, it stops on first found. Building the entire lookup dict is unnecessary.

tests/unit/amlb/resources/test_framework_definition.py
sebhrusen: I'll change, but I find parametrized tests much harder to read and less flexible (big ojects like resources need to be rebuilt each time) with few benefits in most cases.

amlb/resources.py
PGijsbers: > next is better here, it stops on first found. Building the entire lookup dict is unnecessary.

is what I alluded to with

> (I realize in terms of optimization it is worse, but given the context it should be optimized for devs not compute)

PGijsbers: > do you find it hard to read?

Not really, I've seen it before but perhaps just not enough, because of the generator syntax the first thought is that we are filtering rather than looking up one element. But that's all there is to it, so either style works for me.

PGijsbers (approved):
Looks good to me

PR #178, opened 2020-10-23
Fix potential module collisions in tested framework

sebhrusen:
removed `amlb/` from the subprocess `PYTHONPATH` as otherwise any `amlb` submodule can create a collision with a module needed by the framework being tested.

tested with AutoGluon (it uses the `utils` module internally).


PGijsbers:
I would rather go [this way](https://github.com/openml/automlbenchmark/commit/3f66d324a1c93641bd454b5bc1752022f89fa4e5), making use of the changes in #176.

Notes:
 - the commit only demonstrates the changes required for AutoGluon
 - the AutoGluon folder was renamed because the #176 branch does not yet include the fixes you made in #177.

I think it is better because:
 - it avoids adding the "amlb" to a path altogether (that includes through a custom importer)
 - it avoids using any custom import logic
 - it removes the need for two different imports in `callee.py`

`python runbenchmark.py autogluon` runs without problems in that commit

sebhrusen:
https://github.com/openml/automlbenchmark/pull/178#issuecomment-715495761

Did you try it? It shouldn't work, I don't see how autogluon `venv` will be able to load `amlb.utils` this way. It should fail, because utils is loaded as a submodule, so Py will try to load `aml/__init__.py` first, and then load all the dependencies, which is exactly what we're trying to avoid in the `venv`.

sebhrusen:
What's the status with this? I still think that this solution is the most generic (should work for any future framework).

PGijsbers:
Sorry, it got off the radar. I can live with this, it looks like the best solution that works without adding another top-level package.

PR #182, opened 2020-11-06
group runners in same python package for clarity

sebhrusen:


PGijsbers (approved):

amlb/__init__.py
```diff
@@ -19,7 +17,6 @@
     "DockerBenchmark",
     "SingularityBenchmark",
     "AWSBenchmark",
-    "AWSRemoteBenchmark",
```
PGijsbers, line 22: What was the reason for no longer exposing `AWSRemoteBenchmark`?

sebhrusen: it's an old unfinished idea, maybe I'll get back to it one day, but currently it just doesn't work, so no good reason to expose it.
Here is the goal of this component:
```
idea is to handle results progressively on the remote side and push results to s3 as soon as they're generated
this would allow to safely run multiple tasks on single AWS instance without the risk of losing all results if instance crashes after the Nth task.
```

PR #183, opened 2020-11-11
Update Auto-sklearn code to latest version

mfeurer:
Auto-sklearn 0.11.1 uses a slightly different memory management than previous versions. Also, installs packaging as this was previously installed yet.

### DIFF
--- a/frameworks/autosklearn/exec.py
+++ b/frameworks/autosklearn/exec.py
@@ -1,6 +1,7 @@
 import logging
 import math
 import os
+import shutil
 import tempfile as tmp
 import warnings
 
@@ -80,11 +81,6 @@ def run(dataset, config):
         ensemble_memory_limit = config.framework_params.get('_ensemble_memory_limit', 'auto')
         # when memory is large enough, we should have:
         # (cores - 1) * ml_memory_limit_mb + ensemble_memory_limit_mb = config.max_mem_size_mb
-        total_memory_mb = utils.system_memory_mb().total
-        if ml_memory_limit == 'auto':
-            ml_memory_limit = max(min(config.max_mem_size_mb,
-                                      math.ceil(total_memory_mb / n_jobs)),
-                                  3072)  # 3072 is autosklearn defaults
         if ensemble_memory_limit == 'auto':
             ensemble_memory_limit = max(math.ceil(ml_memory_limit - (total_memory_mb - config.max_mem_size_mb)),
                                         math.ceil(ml_memory_limit / 3),  # default proportions
@@ -107,7 +103,8 @@ def run(dataset, config):
     constr_params["seed"] = config.seed
 
     log.info("Auto-sklearn constructor arguments: %s", constr_params)
-    log.info("Auto-sklearn fit() arguments: %s", training_params)
+    log.info("Auto-sklearn additional constructor arguments: %s", training_params)
+    log.info("Auto-sklearn fit() arguments: %s", fit_extra_params)
 
     auto_sklearn = estimator(**constr_params, **training_params)
     with utils.Timer() as training:
@@ -142,13 +139,26 @@ def save_artifacts(estimator, config):
             models_file = os.path.join(output_subdir('models', config), 'models.txt')
             with open(models_file, 'w') as f:
                 f.write(models_repr)
-        if 'debug' in artifacts:
+        if 'debug_as_files' in artifacts or 'debug_as_zip' in artifacts:
             print('Saving debug artifacts!')
             debug_dir = output_subdir('debug', config)
             ignore_extensions = ['.npy', '.pcs', '.model', '.ensemble', '.pkl']
-utils.zip_path(tmp_directory,
-                       os.path.join(debug_dir, "artifacts.zip"),
-                       filtr=lambda p: os.path.splitext(p)[1] not in ignore_extensions)
+            tmp_directory = estimator.automl_._backend.temporary_directory
+            if 'debug_as_files' in artifacts:
+                files_to_copy = []
+                for r, d, f in os.walk(tmp_directory):
+                    for file_name in f:
+                        base, ext = os.path.splitext(file_name)
+                        if ext not in ignore_extensions:
+                            files_to_copy.append(os.path.join(r, file_name))
+                for filename in files_to_copy:
+                    dst = filename.replace(tmp_directory, debug_dir+'/')
+                    os.makedirs(os.path.dirname(dst), exist_ok=True)
+                    shutil.copyfile(filename, dst)
+            else:
+                utils.zip_path(tmp_directory,
+                           os.path.join(debug_dir, "artifacts.zip"),
+                           filtr=lambda p: os.path.splitext(p)[1] not in ignore_extensions)
     except Exception as e:
         log.debug("Error when saving artifacts= {e}.".format(e), exc_info=True)
 
--- a/frameworks/autosklearn/exec.py
+++ b/frameworks/autosklearn/exec.py
@@ -11,6 +11,7 @@ os.environ['OPENBLAS_NUM_THREADS'] = '1'
 os.environ['MKL_NUM_THREADS'] = '1'
 import autosklearn
 from autosklearn.estimators import AutoSklearnClassifier, AutoSklearnRegressor
+from autosklearn.experimental.askl2 import AutoSklearn2Classifier
 import autosklearn.metrics as metrics
 from packaging import version
 
@@ -21,12 +22,16 @@ log = logging.getLogger(__name__)
 askl_version = version.parse(autosklearn.__version__)
 
 def run(dataset, config):
-    log.info(f"\n**** AutoSklearn [v{askl_version}]****\n")
-    save_metadata(config, version=askl_version)
+    askl_method_version = 2 if config.framework_params.get('_askl2', False) else 1
+    askl_string = "Auto-sklearn2.0" if askl_method_version == 2 else "Auto-sklearn"
+
+    log.info(f"\n**** {askl_string} [v{autosklearn.__version__}]****\n")
+    save_metadata(config, version=autosklearn.__version__)
     warnings.simplefilter(action='ignore', category=FutureWarning)
     warnings.simplefilter(action='ignore', category=DeprecationWarning)
 
     is_classification = config.type == 'classification'
+    dataset_name = config.name
 
     # Mapping of benchmark metrics to autosklearn metrics
     metrics_mapping = dict(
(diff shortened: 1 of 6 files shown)


frameworks/autosklearn/exec.py
```diff
@@ -50,45 +51,64 @@ def run(dataset, config):
...
-                                    math.ceil(ml_memory_limit / 3),  # default proportions
-                                    1024)  # 1024 is autosklearn defaults
-    log.info("Using %sMB memory per ML job and %sMB for ensemble job on a total of %s jobs.", ml_memory_limit, ensemble_memory_limit, n_jobs)
+        ml_memory_limit = max(
+            min(
+                config.max_mem_size_mb / n_jobs,
+                math.ceil(total_memory_mb / n_jobs)
+            ),
+            3072  # 3072 is autosklearn default and we use it as a lower bound
+        )
```
sebhrusen, line 71: this block seems to be used only for version > 0.11, maybe it should be included in the if clause below.
Also, it looks very similar to lines 82-86 below with one difference:
```
L67: config.max_mem_size_mb / n_jobs,
L84: config.max_mem_size_mb
```
L84 looks like a mistake (it was probably a mistake before your changes) if we want to fulfill the equation L81.

frameworks/autosklearn/exec.py
```diff
@@ -50,45 +51,64 @@ def run(dataset, config):
...
-                             **constr_extra_params,
-                             **training_params)
+        fit_extra_params['metric'] = perf_metric
+
+    constr_params["time_left_for_this_task"] = config.max_runtime_seconds
+    constr_params["n_jobs"] = n_jobs
+    constr_params["seed"] = config.seed
+
+    log.info("Auto-sklearn constructor arguments: %s", constr_params)
+    log.info("Auto-sklearn fit() arguments: %s", training_params)
```
sebhrusen, line 109: `fit() arguments` != `training_params`
the latter are additional optional params that can be passed in the framework definition.

you probably meant `fit_extra_params`

frameworks/autosklearn/exec.py
```diff
@@ -119,8 +139,23 @@ def save_artifacts(estimator, config):
...
+            files_to_copy = []
+            for r, d, f in os.walk(tmp_directory):
+                for file_name in f:
+                    base, ext = os.path.splitext(file_name)
+                    if ext not in ignore_extensions:
+                        files_to_copy.append(os.path.join(r, file_name))
+            for filename in files_to_copy:
+                dst = filename.replace(tmp_directory, debug_dir+'/')
+                os.makedirs(os.path.dirname(dst), exist_ok=True)
+                copyfile(filename, dst)
```
sebhrusen, line 156:
```suggestion
utils.zip_path(tmp_directory,
                       os.path.join(debug_dir, "artifacts.zip"),
                       filtr=lambda p: os.path.splitext(p)[1] not in ignore_extensions
``````
strongly recommending the archive as this can be uploaded from ec2 to s3 and then to downloaded to the machine running the benchmark.

File: frameworks/autosklearn/setup.sh [hunk 4]
```diff
@@ -15,6 +15,8 @@ if [[ -x "$(command -v apt-get)" ]]; then
     SUDO apt-get install -y build-essential swig
 fi
 
+PIP install packaging
```
sebhrusen, line 18: by convention we tend to put all those additional dependencies in a `requirements.txt` file, it's easier to see what is needed.
```
PIP install -r $HERE/requirements.txt
```

frameworks/autosklearn/exec.py
mfeurer: Thanks for pointing this out. Yes, line 84 was a mistake that was there before which I forgot to update. But I think the lines in 82-86 can safely be dropped as they are the same as the new lines 65-72.

frameworks/autosklearn/exec.py
mfeurer: Yes, indeed. I will update this locally, test locally, and then update this PR.

mfeurer:
Hey @sebhrusen I made the requested changes. However, I slightly deviated from your suggestion about creating debug zip files. For anyone not using S3 buckets it is more convenient to work with files, so I added an option to keep either the files or create a zip.

resources/frameworks.yaml
```diff
@@ -59,11 +59,10 @@ AutoGluon_best:
...
 
 autosklearn:
-  version: '0.8.0'
+  version: '0.11.1'
   project: https://automl.github.io/auto-sklearn/
-#  params:
-#    _save_artifacts: ['models']
-#    _n_jobs: 1
+  params:
+    _save_artifacts: ['debug_as_files']
```
sebhrusen, line 65: I'm fine with having 2 params for the debug format, but can we disable them by default?

sebhrusen (commented):
Quick question: how easy is it support `autosklearn 2.0`?
I understand that it supports only classification for now, but is it enough to do sth like the following:
```
askl2 = config.framework_params.get('_askl2', False)
estimator = (AutoSklearn2Classifier if askl2 and is_classification
                     else AutoSklearnClassifier if is_classification
                     else AutoSklearnRegressor)
```
and then the rest of the code (especially the params) will work for ASKL2?

frameworks/autosklearn/exec.py
```diff
@@ -119,8 +136,28 @@ def save_artifacts(estimator, config):
...
+                files_to_copy = []
+                for r, d, f in os.walk(tmp_directory):
+                    for file_name in f:
+                        base, ext = os.path.splitext(file_name)
+                        if ext not in ignore_extensions:
+                            files_to_copy.append(os.path.join(r, file_name))
+                for filename in files_to_copy:
+                    dst = filename.replace(tmp_directory, debug_dir+'/')
+                    os.makedirs(os.path.dirname(dst), exist_ok=True)
+                    shutil.copyfile(filename, dst)
```
sebhrusen, line 154: FYI, there's also a `utils.walk_apply` function for this kind of logic similar to the `utils.zip_path`:
```
utils.walk_apply(directory, 
                           function with signature fn(path, is_dir), 
                           filtr=function with signature fn(path)
                           )
```
not sure this will simplify code much more here though, at least for the filtering logic, for the copying logic, the first callback would still use:
```
dst = filename.replace(tmp_directory, debug_dir+'/')
os.makedirs(os.path.dirname(dst), exist_ok=True)
shutil.copyfile(filename, dst)
```

mfeurer: You mean like:
```python
                def _copy(filename):
                    dst = filename.replace(tmp_directory, debug_dir + '/')
                    os.makedirs(os.path.dirname(dst), exist_ok=True)
                    shutil.copyfile(filename, dst)
                utils.walk_apply(
                    tmp_directory,
                    lambda path, is_dir: _copy(path),
                    filtr=lambda path: (
                        os.path.splitext(path) not in ignore_extensions 
                        and not os.path.isdir(path)
                    ),  
                )
```?

It's a few lines shorter, but I'm not sure if it's more readable. What do you think?

File: resources/frameworks.yaml (hunk 5 above)
Comment (mfeurer) [reply resources/frameworks.yaml]: Yes, sure. Just pushed this change.

File: frameworks/autosklearn/exec.py (hunk 6 above)
Comment (sebhrusen) [reply frameworks/autosklearn/exec.py:154]: yeah, difference in length is minimal.
it's up to you, I tend to prefer high-order functions personally: I see better what's going on.

Comment (sebhrusen) [reply frameworks/autosklearn/exec.py:154]: btw, if you define 
```
def _copy(filename, **args):
```
then you don't need the lambda param, and it's more readable imo.

Comment (mfeurer) [reply frameworks/autosklearn/exec.py:154]: Sorry, I can't follow. Could you please give a full example?

Comment (sebhrusen) [reply frameworks/autosklearn/exec.py:154]:
```
                def _copy(filename, **_):
                    dst = filename.replace(tmp_directory, debug_dir + '/')
                    os.makedirs(os.path.dirname(dst), exist_ok=True)
                    shutil.copyfile(filename, dst)
                utils.walk_apply(
                    tmp_directory,
                    _copy,
                    filtr=lambda path: (
                        os.path.splitext(path) not in ignore_extensions 
                        and not os.path.isdir(path)
                    ),  
                )
```

**Comment by sebhrusen** [comment]
@mfeurer How about autosklearn 2 support: https://github.com/openml/automlbenchmark/pull/183#pullrequestreview-531626205
Is it something you prefer to keep for a separate PR?

File: frameworks/autosklearn/exec.py (hunk 6 above)
Comment (mfeurer) [reply frameworks/autosklearn/exec.py]: Done, thanks for providing me the exact code to do so.

**Comment by mfeurer** [comment]
Hey, I just updated the PR to also support Auto-sklearn 2.0 and remove meta-data for Auto-sklearn 1.0. This integration should work for now.

File: frameworks/autosklearn/exec.py [hunk 7]
```diff
@@ -42,53 +49,82 @@ def run(dataset, config):
...
-                                  math.ceil(total_memory_mb / n_jobs)),
-                              3072)  # 3072 is autosklearn defaults
-    if ensemble_memory_limit == 'auto':
-        ensemble_memory_limit = max(math.ceil(ml_memory_limit - (total_memory_mb - config.max_mem_size_mb)),
-                                    math.ceil(ml_memory_limit / 3),  # default proportions
-                                    1024)  # 1024 is autosklearn defaults
-    log.info("Using %sMB memory per ML job and %sMB for ensemble job on a total of %s jobs.", ml_memory_limit, ensemble_memory_limit, n_jobs)
+        ml_memory_limit = max(
+            min(
+                config.max_mem_size_mb / n_jobs,
```
PGijsbers, line 80: @sebhrusen I think we should probably have some configuration error if `config.max_mem_size_mb < total_memory_mb`?

@mfeurer considering you are able to define a dynamic default here (`max(memory_limit/n_jobs, 3072)`), I am curious if there is a reason for not making this the default in the package itself?

sebhrusen: > I think we should probably have some configuration error if config.max_mem_size_mb < total_memory_mb?

app currently emits a warning, not an error. Main reason is to allow users to run benchmarks locally using any constraint, without worrying too much about the memory constraint.
Could add a config flag to switch this warning to error by default.

mfeurer: > considering you are able to define a dynamic default here (max(memory_limit/n_jobs, 3072)), I am curious if there is a reason for not making this the default in the package itself?

I only updated the code which was there. I don't remember who wrote the original code. 

Changing this would only work for the single machine setup of the challenge (and most users), so I'm not sure yet if this would be a net improvement. I'll keep thinking about this, but don't know if we can change this before the benchmark cutoff.

PGijsbers: Yes that's fine. I just thought it would be something to consider.

PGijsbers: > Main reason is to allow users to run benchmarks locally using any constraint, without worrying too much about the memory constraint.

Ah, I think that's valid. Some switch could be a nice safeguard, made a note (#214) for later.

sebhrusen:
@mfeurer I'm pushing important updates on master. I'll resolve conflicts on your PR, will test locally, and approve/merge today.

sebhrusen (approved):
looks good: tested on docker with both v1 and v2.
Also removed the symlink to use supported `extend` in framework definitions.

sebhrusen:
thx for the upgrade @mfeurer , merging

PR #185, opened 2020-11-18
Run 2 folds by default when using (default) test constraint

sebhrusen:
run 2 folds by default in test mode to ensure that all rows from the dataset get through the training process


PGijsbers (approved):

PGijsbers:
Fine with me.

PR #189, opened 2020-11-23
add cpu, memory and swap limits to docker run

dev-rinchin:
The case: running multiple bencmark runs on singe machine with lots of resources

### DIFF
--- a/amlb/runners/docker.py
+++ b/amlb/runners/docker.py
@@ -52,11 +52,11 @@ class DockerBenchmark(ContainerBenchmark):
         script_extra_params = ""
         inst_name = self.sid
 
-        if "max_mem_size_mb" in self.constraint_def:
-            memory = "--memory={}m".format(self.constraint_def['max_mem_size_mb'])
-            memory += " " + "--memory-swap={}g".format(2 * float(self.constraint_def['max_mem_size_mb']) / 1024)
-        else:
-            memory = ""
+        memory = ""
+        if rconfig().docker.constraints_enabled:
+            if "max_mem_size_mb" in self.constraint_def:
+                memory = "--memory={}m".format(self.constraint_def['max_mem_size_mb'])
+                memory += " " + "--memory-swap={}".format(self.constraint_def['swap'])
 
         cmd = (
             "docker run --name {name} {options} "
--- a/amlb/runners/docker.py
+++ b/amlb/runners/docker.py
@@ -51,9 +51,17 @@ class DockerBenchmark(ContainerBenchmark):
             touch(d, as_dir=True)
         script_extra_params = ""
         inst_name = self.sid
+
+        if "max_mem_size_mb" in self.constraint_def:
+            memory = "--memory={}m".format(self.constraint_def['max_mem_size_mb'])
+            memory += " " + "--memory-swap={}g".format(2 * float(self.constraint_def['max_mem_size_mb']) / 1024)
+        else:
+            memory = ""
+
         cmd = (
             "docker run --name {name} {options} "
             "-v {input}:/input -v {output}:/output -v {custom}:/custom "
+            "{cpus} {memory} "
             "--rm {image} {params} -i /input -o /output -u /custom -s skip -Xrun_mode=docker {extra_params}"
         ).format(
             name=inst_name,
@@ -64,6 +72,8 @@ class DockerBenchmark(ContainerBenchmark):
             image=self._image_name,
             params=script_params,
             extra_params=script_extra_params,
+            cpus="--cpus={}".format(self.constraint_def['cores']) if "cores" in self.constraint_def else "",
+            memory=memory,
         )
         log.info("Starting docker: %s.", cmd)
         log.info("Datasets are loaded by default from folder %s.", in_dir)
--- a/amlb/runners/docker.py
+++ b/amlb/runners/docker.py
@@ -52,6 +52,11 @@ class DockerBenchmark(ContainerBenchmark):
         script_extra_params = ""
         inst_name = self.sid
 
+        if ' -f ' in script_params:
+             inst_name = inst_name + '_f' + script_params.split(' -f ')[1].split()[0]
+        elif ' --fold ' in script_params:
+            inst_name = inst_name + '_f' + script_params.split(' --fold ')[1].split()[0]
+
         memory = ""
         if rconfig().docker.constraints_enabled:
             if "max_mem_size_mb" in self.constraint_def:
--- a/resources/config.yaml
+++ b/resources/config.yaml
@@ -76,6 +76,8 @@ container: &container
 docker:
   <<: *container
   run_extra_options: '--shm-size=1024M'
+  constraints_enabled: false  # enable resources constraints for docker container
+  swap: '0' # --memory-swap option to docker run
 
 singularity:
   <<: *container
--- a/resources/constraints.yaml
+++ b/resources/constraints.yaml
@@ -4,27 +4,32 @@ test:
   folds: 1
   max_runtime_seconds: 600
   cores: 4
+  max_mem_size_mb: 32768
 
-1h4c:
+1h4c32g:
   folds: 10
   max_runtime_seconds: 3600
   cores: 4
+  max_mem_size_mb: 32768
 
-1h8c:
+1h8c32g:
   folds: 10
   max_runtime_seconds: 3600
   cores: 8
   min_vol_size_mb: 16384
+  max_mem_size_mb: 32768
 
-4h8c:
+4h8c32g:
   folds: 10
   max_runtime_seconds: 14400
   cores: 8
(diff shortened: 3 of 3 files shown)


sebhrusen:
@dev-rinchin, thanks for this proposal! I like the idea but I'd like to make this optional.

 In `resources/config.yaml`, could you please add the following entry to the `docker` namespace?
```yaml
docker:
  ...
  constraints_enabled: false     # if you have a better, feel free to suggest one
```

you can then use the following to apply the additional limitations:
```python
if rconfig().docker.constraints_enabled:
    # do your stuff
```

The fact is that currently the OpenML benchmarks are run mainly using docker on ec2 instances, and there I prefer to rely on docker's default behaviour regarding resource limitations (i.e. no limitations) to ensure that all available resources are used.

amlb/runners/docker.py
```diff
@@ -51,9 +51,17 @@ def _start_container(self, script_params=""):
             touch(d, as_dir=True)
         script_extra_params = ""
         inst_name = self.sid
+
+        if "max_mem_size_mb" in self.constraint_def:
+            memory = "--memory={}m".format(self.constraint_def['max_mem_size_mb'])
+            memory += " " + "--memory-swap={}g".format(2 * float(self.constraint_def['max_mem_size_mb']) / 1024)
```
sebhrusen, line 57: for the swap, I would default it to -1 and suggest to offer the possibility to override this thanks to a property in the `docker` namespace in `config.yaml§ as I don't see why your `2 * max_mem` estimation would generalize.

sebhrusen (changes requested):

amlb/runners/docker.py
dev-rinchin: Some memory should be available for other system processes, so memory in container must be set with higher value than limit for benchmark. In the case of full memory, this can cause the process to be killed by the OOM killer, I faced this problem several times due running the benchmarks. So I think we should set the default swap value greater than 0 or actual container limits should be set with a higher value than the limit for benchmark.

sebhrusen: If I undestand docker's doc above correctly, `0` doesn't mean `no-swap` but `unset`, and:
```
if --memory="300m" and --memory-swap is not set, the container can use 600m in total of memory and swap.
```

sebhrusen: > or actual container limits should be set with a higher value than the limit for benchmark

That's why I suggest that it should be possible to override those docker memory limits through the `config.yaml` property.

dev-rinchin: @sebhrusen ok, got it 👍 

Also I want to add `max_mem_size_mb` setting to resources/constraints.yaml like this:
```
1h8c32g:
  folds: 10
  max_runtime_seconds: 3600
  cores: 8
  min_vol_size_mb: 16384
  max_mem_size_mb: 32768
```
There aren't that many different configurations, so for now, it looks ok, what do you think?

sebhrusen: I support that: was also recently surprised that we forgot to add it there.

When no provided, it should default to all available memory.
Thanks for your help on this.

amlb/runners/docker.py
```diff
@@ -51,9 +51,22 @@ def _start_container(self, script_params=""):
             touch(d, as_dir=True)
         script_extra_params = ""
         inst_name = self.sid
+
+        if ' -f ' in script_params:
+             inst_name = inst_name + '_f' + script_params.split(' -f ')[1].split()[0]
+        elif ' --fold ' in script_params:
+            inst_name = inst_name + '_f' + script_params.split(' --fold ')[1].split()[0]
```
sebhrusen, line 58: how do you run your multiple docker instances?
1. separately?
2. using `--parallel=2` for example?

if (1), then you can specify the sid (and therefore the instance name) in the command line using `--session=...`

if (2), then we have to generate the `inst_name` properly: 
- `script_params` is generated internally in `container.py` and there can't be any `--fold` param, it's always `-f`.
- in parallel mode, we can also have same fold, but different tasks.
- I don't like the idea of parsing a command that is generated by the caller, this is pretty dirty.

If possible, I'd prefer for you to use first approach for now (using `--session`) and I'll think about a consistent way to support parallel jobs with containers.

resources/constraints.yaml
```diff
@@ -4,27 +4,32 @@ test:
   folds: 1
   max_runtime_seconds: 600
   cores: 4
+  max_mem_size_mb: 32768
 
-1h4c:
+1h4c32g:
```
sebhrusen, line 9: No, you simply can't rename all default constraints just because you need it.

As I said, the default should be `max_mem_size_mb=-1` meaning "use all memory available". That's what the existing constraint do and this must remain this way.

If you need custom constraints, please follow https://github.com/openml/automlbenchmark/blob/master/docs/HOWTO.md#constraints-definition

amlb/runners/docker.py
```diff
@@ -51,9 +51,22 @@ def _start_container(self, script_params=""):
...
+
+        memory = ""
+        if rconfig().docker.constraints_enabled:
+            if "max_mem_size_mb" in self.constraint_def:
+                memory = "--memory={}m".format(self.constraint_def['max_mem_size_mb'])
+                memory += " " + "--memory-swap={}".format(self.constraint_def['swap'])
```
sebhrusen, line 64: if you plan to add `max_mem_size_mb` property to constraints, then it should be supported by all runners, not only docker.

> constraint_def['swap']

swap logic doesn't make sense in constraint definition imo, it can be set globally.

resources/config.yaml
```diff
@@ -76,6 +76,8 @@ container: &container
 docker:
   <<: *container
   run_extra_options: '--shm-size=1024M'
+  constraints_enabled: false  # enable resources constraints for docker container
+  swap: '0' # --memory-swap option to docker run
```
sebhrusen, line 80: Please follow existing conventions: property names specify unit, I suggest `memory_swap_mb`

sebhrusen (changes requested):

resources/constraints.yaml
dev-rinchin: I did this because your article only had configuration for benchmarks - m5.2xlarge instances, so actually 32 gb was the default value for memory constraints. Why does the number of cores have a limit in the original constraints, but the memory doesn't?

What configurations do you plan to run in the second version of paper?

sebhrusen: @dev-rinchin the application was originally created for research purpose and the main target to run benchmarks was AWS, and we use docker on AWS for various reasons among which: faster + more robust setup on ec2 instances, easier to maintain docker images than AWS ones, docker images can be used anywhere...

As AWS was the main target, we could rely only on instance types to specify the benchmark constraints: same number of CPUs, same amount of memory on all those instances. Later we realized that some frameworks were more greedy with disk usage and we added a constraint for this a well that allows to add additional disk space to the ec2 instances.
However, we didn't need the memory constraint because we can't add memory to the instances, and there was no good reason to artificially limit memory for example to 16GB if the instance provides 32GB.

Now these were the original requirements, and I understand that some users like you want to bring new requirements, and I'm open to support this with the following conditions:
- new requirements that are not used for the paper should be optional, and the default behaviour must behave exactly how it behaves currently. In this case, this means no artificial limitation when it's no needed.
- new requirements should be implemented consistently: if we want to have a memory constraint available for constraint definitions, it should be applied consistently for all modes (local, docker, singularity, aws...). If the goal is to support it only for docker, then it should be a docker option (declared globally in the `docker` namespace in `config.yaml`).

Also note that in default local mode and aws mode (that runs the benchmark directly in local mode on the ec2 instance), it's almost impossible to enforce this memory constraint because most frameworks simply ignore it (we would have to monitor all the processes created by the framework...).

Right now, if your goal is to run multiple docker instances on one machine, and want to limit the memory for each docker instance, I would suggest that you just close this PR and use the existing option in `config.yaml`:
```
docker:
  <<: *container
  run_extra_options: '--shm-size=1024M'
```

in your local `~/.config/automlbenchmark/config.yaml`
just define 
```
docker:
  run_extra_options: "--shm-size=1024M --memory=8g --memory-swap=16g"
```
this will be passed directly to each docker instance.
You can also pass it directly at the command line:
```
python runbenchmark.py .... -m docker -Xdocker.run_extra_options="--memory=8g --memory-swap=16g"
```

Sorry for not having thought about this option earlier.

sebhrusen:
Docker memory and swap limitations should be applied 
- in user's local `~/.config/automlbenchmark/config.yaml`:
```
docker:
  run_extra_options: "--shm-size=1024M --memory=8g --memory-swap=16g"
```
those options are passed to every docker instance.

- for more flexibility, this can also be passed directly at the command line:
```
python runbenchmark.py .... -m docker -Xdocker.run_extra_options="--memory=8g --memory-swap=16g"
```

PR #192, opened 2020-11-25
Fix H2O handling of some exotic ARFF datasets

sebhrusen:
- upgrade h2o version
- support ARFF datasets using single quote as delimiter (otherwise fails if those values delimited with ' also contain a comma).
- ensure H2O Runs in its own venv.

### DIFF
--- a/amlb/benchmark.py
+++ b/amlb/benchmark.py
@@ -310,6 +310,7 @@ class TaskConfig:
         self.input_dir = input_dir
         self.output_dir = output_dir
         self.output_predictions_file = os.path.join(output_dir, "predictions.csv")
+        self.ext = ns()  # used if frameworks require extra config points
 
     def __json__(self):
         return self.__dict__
--- a/frameworks/H2OAutoML/__init__.py
+++ b/frameworks/H2OAutoML/__init__.py
@@ -1,13 +1,25 @@
-from amlb.utils import call_script_in_same_dir
+from amlb.benchmark import TaskConfig
+from amlb.data import Dataset
+from amlb.resources import config as rconfig
+from amlb.utils import call_script_in_same_dir, Namespace
 
 
 def setup(*args, **kwargs):
     call_script_in_same_dir(__file__, "setup.sh", *args, **kwargs)
 
 
-def run(*args, **kwargs):
-    from .exec import run
-    return run(*args, **kwargs)
+def run(dataset: Dataset, config: TaskConfig):
+    from frameworks.shared.caller import run_in_venv
+
+    data = dict(
+        train=dict(path=dataset.train.path),
+        test=dict(path=dataset.test.path),
+        target=dict(index=dataset.target.index)
+    )
+
+    config.ext.monitoring = rconfig().monitoring
+    return run_in_venv(__file__, "exec.py",
+                       input_data=data, dataset=dataset, config=config)
 
 
 def docker_commands(*args, setup_cmd=None):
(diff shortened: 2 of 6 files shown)


PGijsbers (commented):
I am a little concerned with how much more complicated H2O is getting than other tools. The quote-character should probably be supplied by the `dataset`/`config` variable, alternatively we should require frameworks to figure it out for themselves (in the future?), but I am not a fan of this middle-ground. Hopefully that would clean up the data loading a bit. 

As for saving artifacts, it seems quite complicated. Why is so much custom code required?

I hadn't noticed this before, but the ` max_runtime_secs_per_model=round(config.max_runtime_seconds/2)`, in my opinion, violates our own guidelines which strife to avoid any complicated hyperparameters. The `max_runtime_secs_per_model` requires the user to make some choice about how many models should be evaluated, which would depend on the dataset, search method, etc. I think it should use the framework's default (which internally can of course depend on `max_runtime_secs`).

sebhrusen:
@PGijsbers most frameworks are Python-based and follow sklearn restrictions, that's why they look simpler, but it's mainly because the benchmark app does all the preparation for them (data loading, conversion to numpy, encoding, taking missing data into account..  not trivial things).
On the other side, a few frameworks like H2OAutoML, MLPlan or GAMA are able to handle arff files directly, it's normal that we don't do the data preparation for them.

> The quote-character should probably be supplied by the dataset/config variable, alternatively we should require frameworks to figure it out for themselves (in the future?)

I disagree with that, it's a very unfair requirements for those frameworks. We don't require TPOT, autosklearn, AutoGluon or others to load arff files directly.
Besides, even pandas doesn't guarantee that the auto-detection will work, hence the `quotechar` param in https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html
H2O autodetects the quote char in most cases, we met only one dataset for which this was failing, pandas has the same issue, only the `arff` Python parser handles this correctly.

> As for saving artifacts, it seems quite complicated. Why is so much custom code required?

maintainers that want to export additional  artifacts for analysis are able to do so: this is implemented in AutoGluon, H2OAutoML, autosklearn, TPOT... the complexity mainly depends on the number of different artifacts available.

> the max_runtime_secs_per_model=round(config.max_runtime_seconds/2), in my opinion, violates our own guidelines which strife to avoid any complicated hyperparameters

yeah, it's debatable. Doesn't have any impact on training perf, was just added to avoid timeout. I guess we can get rid of it though.

PGijsbers:
> I disagree with that, it's a very unfair requirements for those frameworks. 

I agree it would be unfair to demand this of some but not others. But would providing the quote character not alleviate this? It seems like this would be helpful also for the other frameworks that take ARFF input.

> the complexity mainly depends on the number of different artifacts available.

That's fair enough. In GAMA I tried to add that bit of logic to the software itself (e.g. specify to keep "logs", "models" or "all") since I found it to be a common thing to do. I don't really want to impose this on others, in fact GAMA's storage is not that flexible so might need custom artifact saving anyway for the benchmark.

Neither of the above issues are very important to me right now, and I don't really have a good answer. But I did want to make a note of it, because I think it is good to make clear what we expect of the frameworks and what we don't, and also to see where the benchmark framework itself could help.

> Doesn't have any impact on training perf

Except I imagine it technically does in the more extreme cases? Unless H2O has safeguards internally, I imagine it can be set so that no (reasonable) model can train, or (unreasonable) model(s) can hold up one of the processes. A good bound would probably both be data and constraint dependent. However, as you said, most reasonable values (e.g. the one here) will behave with little to no difference from each other.

Issue #160, opened 2020-08-06
Provide frameworks definition file with only true latest versions.


sebhrusen:
For users who always want to benchmark some frameworks against the latest version/nightly build, we can provide a `frameworks_latest.yaml` definitions file.
On top of this, we can also start keeping for example a quarterly history of framework versions, allowing users to rerun older versions more easily and analyse progress.

---
From a conversation in https://github.com/openml/automlbenchmark/pull/159

> We could alternatively also have the frameworks at "latest", and just generate a snapshot benchmark definition every few months with fixed versions.

I like the idea, it would also allow users to (re)run older benchmarks easily as we would add a version of `frameworks.yaml` per quarter let's say. Although, for a minimum of stability and as the frameworks are always read from `frameworks.yaml` by default, I would keep the default `frameworks.yaml` with stable versions. Would you be ok with that?
```
frameworks.yaml  # current stable versions
frameworks_latest.yaml  # all frameworks set to latest (may be pulled directly from `master` branch)
frameworks_2020_Q1.yaml
frameworks_2020_Q2.yaml
...
```

So to use latest for example, user would have to run for example
```
python runbenchmark.py resources/frameworks_latest.yaml#mljarsupervised
```
but the default
```
python runbenchmark.py mljarsupervised
```
would still use the default and stable version from `frameworks.yaml`.
I can create a ticket for that, as you can see, if we want to make the switch easy, it requires some changes in how the `framework` parameter is handled to select the alternative `yaml` definition: either by overloading the syntax on the current param (as we do now for benchmarks), or by adding a new param. Personally, I tend to prefer overloading in this situation...

Well, just realized, it may already work using:
```
python runbenchmark.py mljarsupervised -Xframeworks.definition_file=resources/frameworks_latest.yaml
```
(the configuration is very flexible when using the `-X` magic)
I would need to try it.

sebhrusen:
@PGijsbers If you have any suggestion, let's continue the discussion on this ticket.

PGijsbers:
The only addition I have is that in the structure:

```
frameworks.yaml  # current stable versions
frameworks_latest.yaml  # all frameworks set to latest (may be pulled directly from `master` branch)
frameworks_2020_Q1.yaml
frameworks_2020_Q2.yaml
...
```

[...]

Issue #169, opened 2020-10-06
Benchmark multiple AutoML systems at once


PGijsbers:
When you want to run multiple AutoML systems, you currently have to call the script twice, e.g.:

```
python runbenchmark.py TPOT test test 
python runbenchmark.py auto-sklearn test test
```

It would be convenient to allow all a single command to run
 * all frameworks in a definition: `python runbenchmark.py frameworks.yaml test test`, and/or
 * multiple frameworks comma-separated: `python runbenchmark.py TPOT,auto-sklearn test test`, and/or
 * multiple frameworks from a text file: `python runbenchmark.py frameworksubset.txt test test`
With the text file containing a single line comma separated as above, or one framework per line.

The first option is convenient if you have your own framework definition from which you want to run all frameworks. The latter two are more convenient (and less error prone) if you are just interested in running a subset of the default `frameworks.yaml` definition.

sebhrusen:
The more I think about this issue, the less i'm convinced about it.

First, to make this really useful, we need to implement in such a way that all jobs (for all frameworks) are managed by the same queue, otherwise a simple loop like the one in `runstable.sh` is good enough, and I don't see the need to add complexity inside the app to support this.
```
for c in ${constraints[*]}; do
    for b in ${benchmarks[*]}; do
        for f in ${frameworks[*]}; do
#            echo "python runbenchmark.py $f $b $c -m $mode -p $parallel $extra_params"
            python runbenchmark.py $f $b $c -m $mode -p $parallel $extra_params
        done
    done
done
```

The problem with the loop is that each python process must be completed before starting the next one, so we may be waiting for a few trailing jobs before being able to start many ($parallel) new ones...

Now, if we decide to support only multiple frameworks, then we get the same issue with multiple benchmarks and multiple constraints as shown in the loop above.
Although this would be very practical, this also means adding a lot of complexity in the app to support this...

Will try to figure out how this could be implemented without compromising the existing logic, but I'm afraid this can't be done for `v2`.

PGijsbers:
I am not familiar enough with the job queuing to weigh in right now, but I am fine providing (approximate) support through the shell script and moving this out of scope for `v2`.

sebhrusen:
[...]

PR #196, opened 2020-11-26
Framework definitions (stable, latest, 2020Q4...)

sebhrusen:
https://github.com/openml/automlbenchmark/issues/160

This PR contains the following changes:
- modify framework setups to accept new `stable` version (= last official release) on top  of existing `latest` version (= last build, usually not a release), or a specific version number.
- `frameworks.yaml` now defined all versions as `stable`, so that users will run the latest release by default.
- extended definitions in `frameworks.yaml` with more information (short description taken from framework's website or paper,  reference paper) to make it easier to know the specificity of each framework.
- added some alternative framework definitions:
    - `frameworks_latest` set all versions to `latest`
    - `frameworks_2020Q2` set all versions to what they where at end of Q2 2020. Plan is to create a frozen version file like this regularly. Will try every quarter, but may be every 2 quarters...
- user can easily switch to those alternative definitions using for example `framework_name:latest` or `framework_name:2020Q2`:
```python runbenchmark.py autosklearn:2020Q2``` will verify that the current setup matches the requested version, if not if will execute the setup and run the benchmark.
- each framework folder has a `.installed` file created on first setup that contains the current version installed and allowing to reinstall if the requested version doesn't match the current one.
- framework runs (`exec.py`) return the **true** version of the framework for display in  `results` file.
- https://github.com/openml/automlbenchmark/issues/181 to provide a consistent way to version docker images and to display app version in `results` file.

### Known issues:

- the `.installed` file stores the version tag (e.g. 'stable', 'latest', '3.5.0') used in the framework definition: this is not always the TRUE version. For example, if`stable` is installed in November 2020 on a machine, then running the same framework with stable in May 2021 will use that same version. To force a  reinstall, user needs to force the setup using for example:
```python runbenchmark.py autosklearn -s force```
This may be improved in the future.
- each framework has only one `venv` for all versions.
- there's no guarantee that the framework integration will work for all past versions or with `latest` version. However a framework should be fixed as soon as possible if a new release version breaks the integration: until then, users can still use the last quarter version.

### CODE REVIEW

**Comment by sebhrusen** [comment]
@PGijsbers now I think have all the versions numbers right (app version, docker image tagging, framework real version shown in `results.csv`).
Also included a version bumping mechanism to ensure this remains right.

So, this allows to use have latest `stable` framework versions (= latest official release) as the default behaviour when running the benchmarks.
The drawbacks are still the ones listed in the PR description, but as this is for future `v2.0`, I'm fine with it.

The only thing I'm not sure is if we can use those "stable" defaults during the official benchmark: in my opinion we can't because over the few days that we run the benchmarks, a framework can be upgraded.
So we will have to create a `frameworks-2021Q1.yaml` file for the runs and use:
```python runbenchmark tpot:2021Q1 .... ```
not a big deal imo.

Are you ok with that?

**Comment by sebhrusen** [comment]
Also, i'd like to have this merged soon, ideally before latest framework changes to have everyone aware of the new framework definitions:
- https://github.com/openml/automlbenchmark/pull/183
- https://github.com/openml/automlbenchmark/pull/204
- https://github.com/openml/automlbenchmark/pull/211
- https://github.com/openml/automlbenchmark/pull/203

**Comment by sebhrusen** [comment]
confirming that the version bump works: https://github.com/openml/automlbenchmark/actions/runs/408820241

This means that all branches, commits not tagged with the GitHub release flow, will appear as:
```app_version: dev [mybranch,  commit_hash]``` in `results.csv`

Official release, will appear just with the version tag:
```app_version: v2.0.1``` for example

File: amlb/frameworks/definitions.py [hunk 1]
```diff
@@ -28,19 +32,24 @@ def _load_and_merge_framework_definitions(frameworks_file: Union[str, List[str]]
     if not isinstance(frameworks_file, list):
```
PGijsbers, line 32: Maybe we should remove the ` log.info("Loading frameworks definitions from %s.", frameworks_file)` line above this? It's rather confusing because it actually loads all definitions across tags. Also is there a reason we load all definitions, is there ever a fall back from e.g. `2020Q2` to the default tag?

amlb/resources.py
```diff
@@ -62,6 +63,48 @@ def project_info(self):
...
+        def git(cmd, defval=None):
+            try:
+                return run_cmd(f"git {cmd}", _log_level_=logging.DEBUG)[0].strip()
+            except Exception:
+                return defval
+
+        na = "NA"
+        version = git("--version")
+        is_git_repo = version and git("rev-parse --git-dir 2> /dev/null")
+        if version and is_git_repo:
```
PGijsbers, line 77: Duplication since `if version and is_git_repo` expands to `if version and version and git("rev-parse --git-dir 2> /dev/null")`.
```suggestion
        if is_git_repo:
```

amlb/resources.py
```diff
@@ -62,6 +63,48 @@ def project_info(self):
...
+            # git("fetch")
```
PGijsbers, line 78:
```suggestion
```

amlb/resources.py
```diff
@@ -77,13 +120,18 @@ def _seed(self):
...
         """
         :param name:
         :return: name of the framework as defined in the frameworks definition file
         """
         lname = name.lower()
-        framework = next((f for n, f in self._frameworks if n.lower() == lname), None)
+        if tag is None:
+            tag = default_tag
+        if tag not in self._frameworks:
+            raise ValueError("Incorrect tag `{}`: only those among {} are allowed.".format(tag, self.config.frameworks.tags))
```
PGijsbers, line 132: Since we're Py3.6+, is there a reason not to use `f-string` here?

docs/HOWTO.md
```diff
@@ -244,10 +244,10 @@ Then the datasets can be declared in the benchmark definition file as follow:
 - the `folds` attribute is also optional but recommended for those datasets as the default value is `folds=10` (default amount of folds in openml datasets), so if you don't have that many folds for your custom datasets, it is better to declare it explicitly here.
 - Remote files are downloaded to the `input_dir` folder and archives are decompressed there as well, so you may want to change the value of this folder in your [custom config.yaml file](#custom-configuration) or specify it at the command line with the `-i` or `--indir` argument (by default, it points to the `~/.openml/cache` folder).
 
-#### OpenML suites
+#### OpenML studies
```
PGijsbers, line 247: Actually OpenML refers to collections of tasks as a _suite_ since a year or two, see [docs](https://docs.openml.org/benchmark/). A study is a collection of runs (i.e. the results of running algorithms (flows) on a suite (tasks)). It's definitely possible I have used the incorrect naming before.
```suggestion
#### OpenML suites
```

```suggestion
#### OpenML suites
```

docs/HOWTO.md
```diff
@@ -244,10 +244,10 @@ Then the datasets can be declared in the benchmark definition file as follow:
...
 
-[OpenML] suites are a collection of OpenML tasks, for example <https://www.openml.org/s/218>.
-You can run the benchmark on an openml suite directly, without defining the benchmark in a local file:
+[OpenML] studies are a collection of OpenML tasks, for example <https://www.openml.org/s/218>.
+You can run the benchmark on an openml study directly, without defining the benchmark in a local file:
```
PGijsbers, line 250:
```suggestion
[OpenML] suites are a collection of OpenML tasks, for example <https://www.openml.org/s/218>.
You can run the benchmark on an openml suite directly, without defining the benchmark in a local file:
```

PGijsbers (approved):
> So we will have to create a frameworks-2021Q1.yaml file ... Are you ok with that?

More than okay, I would do this even if we had a guarantee versions don't change.
By capturing versions in a definition file it makes it easier for others to replicate our experiments without having to pull true versions from logs first.

We do have to open an issue to update our baselines to respect the scikit-learn version of the definition.

amlb/benchmark.py
```diff
@@ -261,10 +264,15 @@ def _append(self, board):
         Scoreboard.all(rconfig().output_dir).append(board).save()
 
     def _setup_done(self, mark=False):
-        marker_file = os.path.join(self._framework_dir, '.marker_setup_safe_to_delete')
-        setup_done = os.path.isfile(marker_file)
+        installed = os.path.join(self._framework_dir, '.installed')
```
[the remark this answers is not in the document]
sebhrusen: true, don't know why I implemented it this way originally.

amlb/frameworks/definitions.py
sebhrusen: Only reason is to minimize changes. 
But loading everything on startup you also brings some benefit as all definition files are loaded and validated on startup: you don't have to test them all, one after another...

We may want to change this approach when we'll have too many 202xQn definition files, but for now, it allows to keep API as it was, without to change the resources logic, and this should be very useful for https://github.com/openml/automlbenchmark/issues/169

amlb/resources.py
sebhrusen: matter of taste I guess but I try not to use `f-string` if the expression is too long, I think it makes the string noisy.
For example here:
- f"Incorrect tag `{tag}`: only those among {tags} are allowed."    -> I like, the string is easy to read.
- f"Incorrect tag `{tag}`: only those among {self.config.frameworks.tags} are allowed." -> I don't like, the placeholder represents too much of the sentence and I find it noisy in the context.

docs/HOWTO.md
sebhrusen: Ok sorry, I got confused because I had a chat with @ledell recently, and she was referring to studies when I was referring to suites.
Now OpenML UI is not clear as it still refers to `studies` in the menu: https://www.openml.org/search?type=study
so I thought that `study` was the right word...

@ledell FYI

sebhrusen: Looks like it's not only me that is confused, taken from the docs link you provided:
```
# create the benchmark suite
study = openml.study.create_benchmark_suite( 
    name="MidSize Suite", 
    alias=None,
    description="illustrating how to create a benchmark suite", 
    task_ids=task_ids,
)
```
this example names `study` the result of `create_benchmark_suite` :)

sebhrusen (commented):
@PGijsbers thanks for review, committing your suggestions and addressing others

amlb/benchmark.py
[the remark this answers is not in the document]
sebhrusen: although after splitting this in to 2 methods, we see that no approch is perfect.
- single method: 2 different semantic for one method -> not nice.
- 2 methods: share common knowledge about external world (marker file) -> hidden dependency..

docs/HOWTO.md
sebhrusen: another confusing usage:
```python
                    help="The benchmark type to run as defined by default in resources/benchmarks/{benchmark}.yaml, "
                         "a path to a benchmark description file, or an openml suite or task. OpenML references should "
                         "be formatted as 'openml/s/X' and 'openml/t/Y', for studies and tasks respectively. Defaults to `%(default)s`.")
```

amlb/benchmark.py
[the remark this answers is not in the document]
PGijsbers: To avoid your problem about shared knowledge, you can specify the marker file through a parameter (`marker_file`) which is passed on from the `setup` function which calls both. Then the knowledge of where marker file name is still contained in a single place.

Considering we would later want to expand it to allow e.g. multiple environments (for multiple framework versions) one method won't work either way.

amlb/resources.py
PGijsbers: Fair enough! For me the latter is still preferred so I don't have to go back and forth to see what goes where, but to each their own.

amlb/benchmark.py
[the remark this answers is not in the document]
sebhrusen: yes, I know that the file could be externalized... but I already regret the previous implementation tbh... what I liked with the original implementation is that everything was encapsulated in one place and avoided this logic scattering.
I'll concede though...

> Considering we would later want to expand it to allow e.g. multiple environments (for multiple framework versions) one method won't work either way.

I don't understand what you have in mind here. If you mean having multiple venv locally to run multiple versions of the framework at the same time:
1. I don't think it's very useful or a very desirable feature.
2. It means that venvs would stack up locally: not nice.
3. a single file can still hold multiple versions, i don't see the problem here.
4. as soon as we don't have any clear idea about a future change, an implementation should not be designed in details with this potential feature in mind: this is a common mistake.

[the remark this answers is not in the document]
PGijsbers: Ah, I interpreted the known issue "each framework has only one venv for all versions." as an implicit plan to support multiple venvs (for different versions). It could be a useful feature when evaluating version differences (either as a regression test or when issues arise with a new version).

> It means that venvs would stack up locally: not nice.

Agreed.

PR #211, opened 2020-12-06
LightAutoML integration

dev-rinchin:
Hi! I've added LightAutoML framework.
https://github.com/sberbank-ai-lab/LightAutoML


frameworks/lightautoml/exec.py
```diff
@@ -0,0 +1,101 @@
...
+
+from frameworks.shared.callee import call_run, result, output_subdir, utils
+from lightautoml.tasks import Task
+from lightautoml.automl.presets.tabular_presets import TabularAutoML, TabularUtilizedAutoML
+
+log = logging.getLogger(__name__)
+
+
+def run(dataset, config):
+    log.info("\n**** lightautoml (R) ****\n")
```
sebhrusen, line 17: after recent updates, new requirement here: we need the version in the log message and saved as follow:
```
log.info("\n**** lightautoml (R) [{version}] ****\n")
save_metadata(config, version=version)
```

`save_metadata` function is in `frameworks.shared.callee` module

frameworks/lightautoml/exec.py
```diff
@@ -0,0 +1,101 @@
...
+        flags = (np.isnan(oof_pred).sum(axis=1) == 0)
+
+        oof_score = log_loss(
+            y_train[flags],
+            oof_pred[flags, :]
+        )
+
+    log.debug(probabilities)
+    log.debug(config.output_predictions_file)
+    print('OOF score: {}'.format(oof_score))
```
sebhrusen, line 73: doesn't look necessary, those scores are computed independently by the app.

frameworks/lightautoml/exec.py
```diff
@@ -0,0 +1,101 @@
...
+
+    return result(
+        output_file=config.output_predictions_file,
+        predictions=np.argmax(probabilities, axis=1),
```
sebhrusen, line 77: I guess it doesn't work for regression, correct?
Do you actually support regression?
If not, please raise an exception as soon as possible with an explicit message.

frameworks/lightautoml/setup.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+HERE=$(dirname "$0")
+AMLB_DIR="$1"
+VERSION=${2:-"latest"}
```
sebhrusen, line 4: new requirements:
-  default is now 'stable' -> points to last release.
see other frameworks to show how it's handled, usually sth like:
```
if [[ "$VERSION" == "stable" ]]; then
    PIP install --no-cache-dir -U ${PKG}
elif [[ "$VERSION" =~ ^[0-9] ]]; then
    ...
```

resources/frameworks.yaml
```diff
@@ -113,7 +113,9 @@ TPOT:
 #    population_size: 25
 #    verbosity: 2
 
-
+lightautoml:
+  project: https://github.com/sberbank-ai-lab/LightAutoML
+  version: '0.2.5'
```
sebhrusen, line 118: new requirements:
- please insert your framework in alphabetical order in the frameworks definitions files.
- `frameworks.yaml` should have version set to `stable` in this file (pointing to last release).
- in `frameworks.yaml`, a short description and references are welcome (see others in that file).
- please also add an entry to new `frameworks_latest.yaml`.

sebhrusen (changes requested):
@dev-rinchin : overall the integration looks good, thanks for this new addition.

Could you please pull latest version  from `master` and rebase your changes on top of this?
I made some changes to framework definitions.

After that, I'll test it locally and it should be good for the merge, thanks!

frameworks/lightautoml/exec.py
dev-rinchin: Yes, we support regression. Fixed, check please.

dev-rinchin:
Thanks for the answer! I've rebased to latest version from master, and added some fixes, please check it.

frameworks/lightautoml/exec.py
```diff
@@ -0,0 +1,96 @@
...
+        truth=y_test,
+        target_is_encoded=is_classification,
+        training_duration=training.duration,
+        predict_duration=predict.duration,
+    )
+
+
+def save_artifacts(automl, config):
+    try:
+        artifacts = config.framework_params.get('_save_artifacts', False)
```
sebhrusen, line 83:
```suggestion
        artifacts = config.framework_params.get('_save_artifacts', [])
```

frameworks/lightautoml/exec.py
```diff
@@ -0,0 +1,96 @@
...
+
+import pandas as pd
+import numpy as np
+import pickle
+import warnings
+
+from frameworks.shared.callee import call_run, result, output_subdir, utils, save_metadata
+from lightautoml.tasks import Task
+from lightautoml.automl.presets.tabular_presets import TabularAutoML, TabularUtilizedAutoML
+from lightautoml import __version__
```
sebhrusen, line 12:
```suggestion
import logging
import os
import pickle
import warnings

import matplotlib
import numpy as np
import pandas as pd
matplotlib.use("agg")  # no need for tk

from lightautoml.tasks import Task
from lightautoml.automl.presets.tabular_presets import TabularAutoML, TabularUtilizedAutoML
from lightautoml import __version__

from frameworks.shared.callee import call_run, result, output_subdir, utils, save_metadata
```
I'm being picky here,  but just grouping import by usual conventions (standard libs to most specific libraries) and using lexicographic order.

Also wanted to be able to run it directly on macOS: as you're internally using matplotlib, using `agg` backend is usually enough to make code portable on macOS.
However, it looks like you're spawning subprocesses and those fail with the usual "No module named _tkinter" on macOS, so the backend would have to be set on your side.
Not making it a requirement, but I think your users would be happy to be able to run your tool directly on macOS.

sebhrusen (approved):
ready to merge, please see my last  2 comments.

I had difficulties to build docker images due to the very large `torch` library that makes pip timeout, but I'll fix this separately.

frameworks/lightautoml/exec.py
sebhrusen: raises an exception otherwise

frameworks/lightautoml/exec.py
dev-rinchin: > Also wanted to be able to run it directly on macOS
Ok, I will fix that in our library, thank you!

PR #217, opened 2020-12-10
Fix docker image discovery

sebhrusen:
Given that the project repository is defined as
```
project_repository: https://github.com/openml/automlbenchmark#stable
```
where project_version='stable' points to the last release by default.

This PR ensures that the  docker images discovery works as follow:
- if using a released version of benchmark app: search by default image `automlbenchmark/{framework}:{framework-version}-{project_version}` (or just `automlbenchmark/{framework}:{framework-version}` if `project_version`==`stable`).
this is expected to be the behaviour when running official, reproducible benchmarks.
- if using a dev version (= any non-release version) of the benchmark app: search images in the following order:
   1. `automlbenchmark/{framework}:{framework-version}-dev`
   2. `automlbenchmark/{framework}:{framework-version}-{project_version}` (or just `automlbenchmark/{framework}:{framework-version}` if `project_version`==`stable`)
- if user specifies a custom image using `-Xdocker.image=foo`, search only for image `foo`.

If the image is not found locally or remotely (docker repo), then the application will automatically ask to build it.

### DIFF
--- a/amlb/frameworks/definitions.py
+++ b/amlb/frameworks/definitions.py
@@ -17,7 +17,7 @@ def load_framework_definitions(frameworks_file: Union[str, List[str]], config: N
     Loads the definition(s) from the file(s),
     :param frameworks_file:
     :param config:
-    :return: Namespace containing each framework definition,
+    :return: Namespace containing all framework definitions per label (definition namespace).
     """
     frameworks = _load_and_merge_framework_definitions(frameworks_file, config)
     for tag, defs in frameworks:
--- a/amlb/frameworks/definitions.py
+++ b/amlb/frameworks/definitions.py
@@ -61,6 +61,7 @@ def _sanitize_and_add_defaults(frameworks, config):
     for _, framework in frameworks:
         if "extends" not in framework:
             _add_default_module(framework, config)
+            _add_default_image(framework, config, props=['image'])
     _update_frameworks_with_parent_definitions(frameworks)
 
     _add_defaults_to_frameworks(frameworks, config)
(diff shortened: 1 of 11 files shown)


amlb/frameworks/definitions.py
```diff
@@ -138,19 +139,19 @@ def _add_default_params(framework):
         framework.params = Namespace.dict(framework.params)
 
 
-def _add_default_image(framework: Namespace, config: Namespace):
+def _add_default_image(framework: Namespace, config: Namespace, props=None):
```
PGijsbers, line 142:
```suggestion
def _add_default_image(framework: Namespace, config: Namespace, props: Optional[List[str]] = None):
```
will have to also import `Optional` :/

PGijsbers (approved):
Looks good to me, but haven't had time to test it yet. Feel free to merge, but I could test it later if  you prefer.

amlb/frameworks/definitions.py
sebhrusen: added the typing annotation, but in general I think we can be relaxed with private methods.
@PGijsbers using this comment to suggest some guidelines.
Typing annotations can be useful to define the contract of public methods/functions, much less for private ones as those are susceptible to have their contract changed at any time.
I think it's much more important to keep all method signatures small (2-3 positional params max) to make them easy to reason about, and intuitive.
Type system should be here to help, not to make things look more complex, or worse, add unnecessary time-wasting constraints (having written a lot of code in both strongly typed languages and loosely types ones, I can tell that both have their benefits depending on the context, the abstration level...  and we should not be religious about it).
Same with testing: many unit tests currently test private methods. This should be avoided: private methods are allowed to change, and then we spend a lot of time fixing those tests. In Java world for example, you never write UTs for private methods. In the same way, our tests should focus on scenarios on the public functions/methods.

Issue #116, opened 2020-05-13
Expose data as pandas Dataframes instead of numpy arrays.


sebhrusen:
Currently, data can be accessed by frameworks either as a raw file (ARFF file when using OpenML, ARFF or CSV when using explicit file), or as a `numpy` array.

Frameworks that can load raw files directly can use their own logic to identify columns types.
For others, the application offers the `data_type` attribute for each feature, but due to the various data loaders, there are some inconsistencies in the type names for historical reasons, and even in the type mapping (openML loader keeps string columns as `string` whereas the file loader considers them as `categorical`).

## Proposal 1:
select a list of types and use only those from 'amlb.data' module, loaders are responsible to ensure that format specific types get converted to one in this list.
For example: categorical, string, integer, real, boolean, datetime.

## Proposal 2:
embrace `pandas` DataFrames instead of `ndarray` (in `amlb.data.DataSplit`): will probably require small changes for each framework integration.
Would be good for a v2.0.

# Conclusion
Proposal 2 it will be

Innixma:
I'm in favor of Proposal 2, unless there are frameworks that require or gain benefits from Proposal 1. `pandas DataFrame` are much less prone to having strange things happen with them regarding dtypes than `numpy ndarray` in my experience, and if necessary, the inner `numpy ndarray` can be accessed from the `pandas DataFrame` object.

sebhrusen:
To anyone interested, I'm about to spend some effort on this: frameworks will be able to directly ingest pandas dataframes instead of numpy arrays and therefore support all types supported by pandas.

PR #224, opened 2020-12-14
Support ignored attributes from OpenML datasets

sebhrusen:


amlb/datasets/openml.py
```diff
@@ -122,13 +123,18 @@ def get_values(f):
...
+        feat_indexer = iter(count())
+        return [Feature(new_idx,
                         f.name,
                         f.data_type,
                         values=get_values(f),
                         has_missing_values=has_missing_values(f),
                         is_target=is_target(f)
-                        ) for i, f in sorted(self._oml_dataset.features.items())]
+                        )
+                for new_idx, f in zip(feat_indexer,
```
PGijsbers, line 134: Any particular reason this is not just done with the `enumerate` built-in?

PGijsbers, line 134: Any particular reason this is not just done with the enumerate built-in?

amlb/datasets/openml.py
```diff
@@ -140,7 +146,7 @@ def attributes(self):
             log.debug("Loading attributes from dataset %s.", self._oml_dataset.data_file)
             with open(self._oml_dataset.data_file) as f:
                 ds = arff.load(f)
-                self._attributes = ds['attributes']
+                self._attributes = [a for a in ds['attributes'] if a[0] not in self._oml_dataset.ignore_attribute]
```
PGijsbers, line 149: What is the reason for loading the arff file instead of using the `OpenMLDataset.features`?
The `OpenMLDataset.ignore_attribute` can be `None`, I believe this will cause issues with the `in` keyword.

amlb/datasets/openml.py
```diff
@@ -218,16 +230,36 @@ def _get_split_path_for_dataset(ds_path, split='train', fold=0):
...
+    else:
+        rows = np.array(rows)
+    full_attributes = full_dataset['attributes']
+    if cols is None:
+        cols = slice(None)
+        attributes = full_attributes
+    elif isinstance(cols, list):
+        cols = np.array(cols)
+        attributes = [full_attributes[i] for i in cols]
+    else:
```
PGijsbers, line 248: Which scenario is this? `cols` is only ever a `list` (in the way it is called right now), I understand the `None` option since it is still the default (even if it is never left at default). Am I missing something?

PGijsbers (changes requested):
I guess it's a bit verbose because we are working on the `arff` file, which is exactly the layer that `openml` wants to abstract away from.

amlb/datasets/openml.py
sebhrusen: code is a living thing :)
looks completely unnecessary now, was introduced with a previous (and wrong) approach.

amlb/datasets/openml.py
sebhrusen: legacy, it's a fallback mechanism for older version of openml-python (see comment line 113).
We can clean this up in a future PR.

sebhrusen: > The OpenMLDataset.ignore_attribute can be None, I believe this will cause issues with the in keyword.

good catch: fails on the test datasets indeed... fixing

amlb/datasets/openml.py
sebhrusen: also not necessary anymore: tried to pass `slice` as params at some point.
will simplify

sebhrusen:
> I guess it's a bit verbose because we are working on the arff file, which is exactly the layer that openml wants to abstract away from.

@PGijsbers 95% of this code was written at a time when openml-python was not providing key features for the benchmark (like the categorical values) and if it can now be simplified thanks to improvements in the openml library, then let's do it!

PGijsbers:
Some things might be easier, but I also think there might be some quirks in processing the pandas data and then writing it back to arff files. I propose I add it to my list so I can look over it later (i.e. not before `v2`).

PGijsbers (approved):

sebhrusen:
>  Some things might be easier, but I also think there might be some quirks in processing the pandas data and then writing it back to arff files

1. numpy data, not pandas (yet).
2. rewriting arff files is actually an optimization: when running the  same fold multiple times on the same machine, you don't have to reprocess the splits each time (can be significant for larger datasets).
3. for frameworks using arff files directly, this is even necessary. For those, it also allows to release data that remains in memory when generating the splits, otherwise a framework using the arff files would be penalized because the openml data would remain in memory for nothing.
4. the main improvement I see would be for `OpenmlDatasplit.data`: on first use of the dataset, those split data could be kept  available in memory and then we wouldn't have to reload the split file. However this is only true for datasets consuming the numpy data, and we would need to be careful of not leaking memory for other frameworks in that case (see 3.)
5. Anyway, I'll probably need to revise this for https://github.com/openml/automlbenchmark/issues/116 and the switch to pandas... which is the next big change.

PR #230, opened 2020-12-24
Download less dataset metadata when loading openml suite

PGijsbers:
While we lose the official dataset description, this puts a lot less strain on the server as not each dataset metadata is downloaded individually.


Innixma (approved):
Worked well for me with `-p 100`, it was much faster and eliminated all of the database access errors. Thanks!

Issue #44, opened 2019-07-17
Automatically submit task results to OpenML


seb-h2o:
We should provide the option to automatically submit results obtained for a given (task, fold, framework) to openml (openml Py client provides `OpenMLRun`, that we need to `publish`)

PR #236, opened 2021-01-04
Post-run Upload to OpenML

### LINKED ISSUE #45: Submit existing task results to OpenML
Once we have implemented the logic to automatically upload new results to OpenML (https://github.com/openml/automlbenchmark/issues/44), we should be able to add a script to publish results a posteriori and use it to publish our existing results from https://github.com/openml/automlbenchmark/tree/master/reports.

PGijsbers:
Posting as draft since I would like some feedback on whether this seems a sensible approach (and/or to see if I am missing something).

These changes allow a run which was completed under the newest version of the benchmark framework to be uploaded to OpenML.
This does not yet upload results as they come in (i.e. closes #45, not #44).

It currently includes:
 - Creating a flow ([example](https://test.openml.org/f/35887)), which takes into account both:
    -  the benchmark version (the tag if available, first 8 characters of the commit hash otherwise), specifies the fork as well
    -  the 'true' automl version
 - Runs record the used benchmark hyperparameters (time, memory, cores, seed) and the original console invocation ([example](https://test.openml.org/r/14070), note: original console invocation is not uploaded due to [a bug in `openml-python`](https://github.com/openml/openml-python/pull/1015))
 - New `runbenchmark.py` arguments:
   - You can tag runs with `--tag TAG_NAME`, this tag will be added to the OpenML run if uploaded
   - You can use the openml test server by setting `--test-server`

Still working on:
 - [x] Command to upload all (complete) results in the results directory, see `upload_results.py`.
 - [ ] Add hyperparameter overwrites for automl systems to run description.
 - [ ] Static tests for each framework (at least those planned for the next evaluation).

Planning to exclude (at least for this PR):
 - Fine-grained flows, where the hyperparameters of the automl frameworks are included (potentially with the automl framework as subflow). I think this becomes rather complicated when we don't have access to the automl frameworks from the upload environment.
 - Automatically merging multiple result folders into one (e.g. when on the first try, a fold failed, and this is executed again separately).


PGijsbers:
A lot of `TaskConfig.__init__` parameters are now only for forwarding data so it gets written to the `metadata.json` file.
Should we instead change the `save_metadata` function so it obtains things from `resources.get` directly (e.g. `tag`, `command`, `git_info`, which should not change between tasks)?

PGijsbers:
This PR will be superseded by work in https://github.com/openml/automlbenchmark/tree/oml_on_upgrade_core after `upgrade_core` is merged with `master`.

Issue #187, opened 2020-11-19
Benchmark Update: Regression and more Classification!


PGijsbers:
We plan to do a new evaluation soon and want to share information about what we have been working on. The new evaluation will be included in a journal paper submission and available through OpenML.

The first big change is the addition of regression tasks to the benchmark. We curated a list of regression tasks that we think both represent real-world problems and is (in principle) compatible with all frameworks. The list is available on OpenML suite 269 (https://www.openml.org/s/269). We will likely use the root mean squared error to evaluate regression performance (but we encourage authors to verify other common regression metrics too).

The second change is the addition of new datasets to the classification benchmark. We are currently working on the final few datasets. We will keep you updated through a this Github issue. The list of only new datasets is available in suite 270 (https://www.openml.org/s/270), while the full list of all classification datasets for the new evaluation is available in suite 271 (https://www.openml.org/s/271). 

**In case you find problems with the datasets or tasks themselves, please also report them there.**

Thank you all for your contributions to the benchmark!

pplonski:
Great! Thank you for the information.

Maybe a little off-topic. What about running AutoML systems that need to set a parameter to run? My AutoML by default is in `Explain` mode, for benchmark it should be run with `Compete` mode. Should I update the default before the new benchmark evaluation?

PGijsbers:
The default framework definition for MLJar in the benchmark currently sets the parameter. [Definition](https://github.com/openml/automlbenchmark/blob/master/resources/frameworks.yaml#L135):

```yaml
mljarsupervised:
  version: '0.6.0'
  project: https://github.com/mljar/mljar-supervised
  params:
    mode: "Compete"   # set mode for Compete, default mode is Explain
```

This is the desired way to set non-default hyperparameters, so it should be good to go 👍

pplonski:
Thank you @PGijsbers, Is it allowed to set parameters for benchmarks? Or framework should run without any additional parameters?

sebhrusen:
@pplonski no, this is not allowed to set parameters in this file, as we'll be using it for the AutoML paper.
We potentially accept only one high-level parameter like this one, but we're also working on including measures of model complexity to evaluate the various tools.

Also, for frameworks like yours that use a non-default param, we'll create 2 entries:
- `mljarsupervised` will use only default parameters.
- `mljarsupervised_compete` can be also added using the definition above.
Only one those will be used for the official runs, so please tell us which one you want us to use, thanks.

PGijsbers:
[...]

PR #237, opened 2021-01-05
Update autogluon to v0.1 API

Innixma:
Updates AutoGluon to use the new v0.1 API (Currently in pre-release).

This PR makes old versions of AutoGluon not able to run (such as v0.0.15) because the API changes are breaking.

If it is important that old versions should still work, I can try to see about adding some try/except checks to determine which API to use.

I have tested this code on the new automlbenchmark study (104 datasets), and AutoGluon_bestquality succeeds on 1h8c constraint with 250GB SSD allocated on 103/104 datasets on all 10 folds (failing dataset is KDDCup09-Upselling due to OOM prior to getting to AutoGluon code).


Innixma:
AutoGluon v0.1 is planned for release on Feb 8th 2021. If the 2021 Q1 benchmark run hasn't ran by then, I could update this PR to use official v0.1 at that time for use in the 2021 Q1 benchmark.

Innixma:
@PGijsbers AutoGluon v0.1.0 has officially released, so this PR should be ready for testing / review on your end.

PGijsbers (approved):
Tested with `python runbenchmark.py autogluon validation test -m aws -f 0 -p 7` and all passed.
@sebhrusen what constraints do you normally use?

sebhrusen:
@PGijsbers to test validation datasets, I keep the default constraints (=test, i.e 10min/fold)

Innixma:
A few recommendations for general configs in automlbenchmark:

To avoid out-of-disk errors, I'd recommend that the default disk space allocations for AWS EC2 instances be increased. Currently, they have 16 GB of disk space in `1h8c` config, which isn't a lot to work with on some of the larger datasets that take up over 1 GB of space on their own + the size of the package and dependencies.

Also, the current default disk storage is very slow (hard disk). I suggest switching to `gp2` as the default, which uses SSDs. This is important for AutoGluon because AutoGluon loads and saves to disk many times during fit and inference (not all models are in memory at once to avoid OOM). I noticed AutoGluon inference and training slowed down significantly if using the slow default disk volume.

For my own testing, I used 500 GB of `gp2` space and never ran into disk space issues. I consider this config to be sufficient to accurately assess AutoGluon and other AutoML packages that rely on larger models / disk caching. This could probably be reduced to 250 GB, but I feel that the cost of disk for a short lived EC2 instance is negligible compared to cost of compute.

For example, for 1 hour m5.2xlarge run:

Cost of m5.2xlarge: $0.384/hr
Cost of 250 GB `gp2`: $0.035/hr
Therefore, giving 250 GB to the machines would be less than 10% of the compute cost.

PGijsbers:
Finally had the opportunity to test it also locally on docker, and the validation set passed 👍 so I'll go ahead and merge.

(We had a chat already, but for the public:)
We currently use AWS spot instances, which have lower pricing than indicated. As such using `gp2 ssd` does bring considerable additional costs. As per #187, we do _want_ to add 250GB of SSD storage, but whether or not we get to use them will ultimately depend on our funding and the costs at the time for executing the benchmark.

sebhrusen:
@PGijsbers I wished we hadn't merged this one already: @ledell was trying it on Saturday and had a serious issue: AG did apparently time out on some dataset, however the models created at runtime are always created in some subfolder of the output directory, and are therefore uploaded to s3 if not previously deleted by the `save_artifacts` function.
Because of the timeout, the `save_artifacts` function was not called, the (large amount of) models not deleted, and this ended up with TB of data uploaded to s3 AND downloaded automaticallly on the machine running the benchmark…

Several things will need to be fixed before the next run:
- @Innixma, AG should apply apply the following changes regarding artifacts : write models to a `tmp` folder and move them for export with `save_artifacts` iff required by user.
- I'll review the auto-download to local machine logic to prevent downloading artifacts if not asked explicitely (safety net).

Innixma:
Thanks for the info @sebhrusen . I had assumed this was intended for debugging purposes since it had always functioned in that fashion in automlbenchmark (even before this PR merge), but I agree it isn't a very good experience in general when it takes up significant S3 space.

Regarding the version used for official benchmarking, we can probably use the upcoming v0.1.1 or a pre-release since it drastically reduces the disk usage via an optimization we made to KNN models.

Re the following:

> AG should apply apply the following changes regarding artifacts : write models to a tmp folder and move them for export with save_artifacts iff required by user.

What folder is safe? By default it saves to `output_dir = output_subdir("models", config)`

Are you suggesting it save to `output_dir = 'tmp/'`?

And re 'move them for export', are you sure it's the right approach to move them on disk? Wouldn't it make more sense for the `save_artifacts` function to be able to download from folder A on machine 1 to folder B on machine 2 / S3?

sebhrusen:
@Innixma my concern is not much about disk usage for now but just to prevent models from being uploaded when the user didn't ask for it.
The current logic works, except if AG times out and the app interrupts—or in the worst case kill—the process.
In that case, the existing models remain in `output_dir` and the contract is that everything in output_dir got uploaded to s3.

> Are you suggesting it save to output_dir = 'tmp/'?

you mean "/tmp"? I would use `tempfile.mkdtemp`. If you have a better idea, feel free to suggest.
In the case where models are requested by user, you could also use [this](https://stackoverflow.com/questions/7119630/in-python-how-can-i-get-the-file-system-of-a-given-file-path) to detect the current volume, and pass it to `tempfile.mkdtemp` to be sure that the tmp folder is on the same volume (see. below).

> And re 'move them for export', are you sure it's the right approach to move them on disk? Wouldn't it make more sense for the save_artifacts function to be able to download from folder A on machine 1 to folder B on machine 2 / S3?

When running on AWS all files should be on the same volume, so moving with `shutil.move` will not require any copying and should be almost instant.
Right now I prefer to keep the framework integration code platform-agnostic: it should not know if it's running on aws, docker, and so on...

PGijsbers:
@sebhrusen Sorry 😢 I did not encounter this issue with the validation datasets under test constraints. I wasn't aware other people were still testing it. Because the PR had been open so long, I assumed I was the only one to still test it.

sebhrusen:
@PGijsbers no problem, the fix can be in a separated PR. Just want to make it clear that the fix is needed before the next run.

Innixma:
@sebhrusen I've created #274 to resolve the issue.

sebhrusen:
Thanks a lot @Innixma , will look at it asap.

PR #244, opened 2021-01-20
[WIP] Adding Github Workflow

PGijsbers:
Adding a Github Workflow. This PR is only to trigger the workflow so I can iterate on it.


sebhrusen:
@PGijsbers looks like your commit graph is a it messed up.
This branch is on top of `oml_upload` but contains recent commits already in `master`… did you do a merge instead of rebasing `oml_upload` first? can you please clean up this?

.github/workflows/run_all_frameworks.yml
```diff
@@ -0,0 +1,68 @@
...
+          python -c "import numpy; print(numpy.__file__)"
+
+  run_frameworks:
+    name: Run validation on all frameworks
+    runs-on: ubuntu-latest
+    needs: baseline
+    strategy:
+      matrix:
+        python-version: [3.8]
+        framework: [autoxgboost, autogluon, gama, h2oautoml, mlplan, tpot, autosklearn]
```
sebhrusen, line 48: can you please list them in alphabetical order so identify easily when one is missing?
I'd also add the reference ones: constantpredictor, randomforest, tunedrandomforest.

.github/workflows/run_all_frameworks.yml
```diff
@@ -0,0 +1,68 @@
...
+        task: [bioresponse, dresses-sales, eucalyptus, internet-advertisements, micro-mass, kc1, APSFailure]
```
sebhrusen, line 49: do we really need all those tasks? also some of those datasets are not small and I don't want to have to wait for the validation workflow for several hours after each commit: test constraint is 10min per task.
If we keep 7 datasets, that we run with 10 frameworks, it makes a  lot of computation. I guess it runs in parallel, but are we not limited in hardware usage?

.github/workflows/run_frameworks.yml
```diff
@@ -0,0 +1,65 @@
...
+        echo "::set-output name=frameworks::`git diff --name-only HEAD..$GITHUB_BASE_REF | grep -o -i -P 'frameworks/(?!shared).*/' | uniq | sed -e 's/frameworks//' -e 's/\///g' | perl -p -e 's/\n//'`"
+
+
+  run_frameworks:
+    name: ${{ matrix.framework }}/${{ matrix.task }}
+    runs-on: ubuntu-latest
+    needs: framework_changes
+    strategy:
+      matrix:
+        framework: [autoxgboost, autogluon, gama, h2oautoml, mlplan, tpot, autosklearn]
```
sebhrusen, line 42: alphabetical order please

.github/workflows/run_frameworks.yml
```diff
@@ -0,0 +1,65 @@
...
+        task: [bioresponse, dresses-sales, eucalyptus, internet-advertisements, micro-mass, kc1, APSFailure]
```
sebhrusen, line 43: same concern as previously

.github/workflows/run_frameworks.yml
```diff
@@ -0,0 +1,65 @@
...
+          key: pip-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            pip-
+      - uses: actions/setup-python@v2
+        if: contains(needs.framework_changes.outputs.output1, matrix.framework)
+        with:
+          python-version: '3.8'
+      - name: Run benchmark
+        if: contains(needs.framework_changes.outputs.output1, matrix.framework)
+        uses: ./.github/actions/runbenchmark
```
sebhrusen, line 62: is it still WIP?
how does it work? is output1 always a string or can there be more than one framework changed? not sure how `contains` work in this case.

.github/workflows/run_frameworks.yml
```diff
@@ -0,0 +1,65 @@
...
+        with:
+          framework: ${{ matrix.framework }}
```
sebhrusen, line 64: we don't need to run the full framework matrix, can't we just pass `needs.framework_changes.outputs.output1` here?

PGijsbers:
@sebhrusen Thanks for the review. Sorry I wasn't clear, this is just a work-in-progress branch to test the workflow. Once it looks fine I will create a separate PR (on a fresh branch) for that. The reason for this PR/branch is that I needed to trigger the workflows. As far as I was aware draft PR's don't trigger workflows.

.github/workflows/run_all_frameworks.yml
```diff
@@ -0,0 +1,68 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+      - name: Install Requirements
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+      - name: Run constantpredictor on openml iris
+        run: |
+          python runbenchmark.py constantpredictor openml/t/59 1h4c
```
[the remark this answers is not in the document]
PGijsbers: The reason I specified this is that;
 - it tests `openml` task specification which is not done in the validation tests (but it might be better as a separate test)
 - the `test` constraints don't test all folds, so I used `1h4c` , I should have probably done `-f 0 1 2 ... 9` instead to make that explicit. (I don't want the regular validation to be done on 10 folds, since that will incur an unreasonable workload).

.github/workflows/run_all_frameworks.yml
PGijsbers: As far as I could tell, we are not limited (other than max 256 parallel jobs). I haven't ran this on all frameworks yet (some still break). During testing I found that quite often most tasks were started in parallel, so in practice it might not take longer than ~20 minutes for the workflow.
I would stay on the side of thoroughness, we should expand our regular unit test suite for quick tests with good coverage. This workflow is then only to make sure we didn't miss anything before merging with `master`, not for regular debugging.

.github/workflows/run_frameworks.yml
PGijsbers: Values are cast to string, and `contains(search, item)` is `true` if `item` is contained in `search` ([doc](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#contains)).
`output1` is one string with all changed frameworks without separator (e.g. `autosklearnh2oautoml`).

.github/workflows/run_all_frameworks.yml
PGijsbers: Will do (just commenting for now, I'll make changes on Monday).

.github/workflows/run_frameworks.yml
PGijsbers: It runs the full matrix, but no steps are executed for frameworks which are not in `needs.framework_changes.outputs.output1`. A nice improvement would be to define the matrix dynamically with [fromJSON](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#fromjson). But it's lower on my list as it doesn't really speed up much (those skipped jobs take maybe a second or two each). It would make it much nicer to review jobs in the workflow though, so I would still like to add it.

sebhrusen:
@PGijsbers either you messed up a merge/rebase locally, or the target branch is wrong for this PR.

PGijsbers:
It was right at first, I just wanted to iterate on the workflow quickly in a separate branch (and make the PR so CI triggers) and then merge it back to `oml_upload`. But at this point I am just sorting out the workflow and then I will set up a new branch with only the workflow stuff (targeting master). I probably pinged you by adding the label, I forgot about that 😓  didn't mean to.

sebhrusen:
@PGijsbers apparently it's just that the target branch is way behind `master`: it needs to be rebased, and then this one on top of the target.

PGijsbers:
Work is moved to #294 based on a fresh branch off master.

PR #246, opened 2021-01-22
Allow exit on error of a job

PGijsbers:
Add the `-e` parameter to allow an error of a job to propagate back up and terminate the execution of `runbenchmark.py`. 
Used in Github Workflow (work in progress) to correctly identify failed evaluations. Also useful if you don't want to run subsequent tasks if one fails.

### DIFF
--- a/amlb/job.py
+++ b/amlb/job.py
@@ -6,16 +6,14 @@
   - SimpleJobRunner runs the jobs sequentially.
   - ParallelJobRunner queues the jobs and run them in a dedicated thread
 """
-from concurrent.futures import ThreadPoolExecutor, ProcessPoolExecutor, as_completed
+from concurrent.futures import ThreadPoolExecutor, ProcessPoolExecutor
 from enum import Enum, auto
 import logging
-import multiprocessing
 import queue
 import signal
 import threading
 import time
 
-from .resources import config as rconfig
 from .utils import Namespace, Timer, InterruptTimeout, raise_in_thread, signal_handler
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


amlb/job.py
```diff
@@ -15,6 +15,7 @@
 import threading
 import time
 
+from .resources import config as rconfig
```
sebhrusen, line 18: can't you find a different way to pass the param here? constructor?
I did as much as I could to keep this important low-level component fully standalone (depends only on `utils` that is itself standalone), otherwise I'm afraid we'll start to introduce various options here and make this component more fragile.

runbenchmark.py
```diff
@@ -47,6 +47,8 @@
                          "•auto: setup is executed only if strictly necessary. •skip: setup is skipped. •force: setup is always executed before the benchmark. •only: only setup is executed (no benchmark).")
 parser.add_argument('-k', '--keep-scores', type=str2bool, metavar='true|false', nargs='?', const=True, default=True,
                     help="Set to true [default] to save/add scores in output directory.")
+parser.add_argument('-e', action='store_true', dest="exit_on_error",
+                    help="If set, *any* task that does not complete with a model will cause the script to terminate.")
```
sebhrusen, line 51:
```suggestion
parser.add_argument('-e', '--exit-on-error', action='store_true', dest="exit_on_error",
                    help="If set, the first task that does not complete with a model will cause the entire script to terminate.")
```
suggesting this to make it prettier when running 
```
python runbenchmark.py --help
```
Ideally, the long syntax should be enough to guess/remember/find an option.

also wondering if the `-k` option above still makes sense.

sebhrusen (changes requested):
Please find a way of not using config in `job.py`

amlb/job.py
PGijsbers: I wasn't aware, I'll find a way around it 👍

runbenchmark.py
PGijsbers: I personally never used `-k` (I want the default behavior). Was there a specific reason for implementing it in the first place?

sebhrusen: Honestly, I can't remember, which is a good sign that it was not a good idea: I also always want the default behaviour.
We can get rid of it in a separate PR

amlb/job.py
sebhrusen: Thanks. Actually, that's one of those components for which I could/should write a bunch of unit tests now.

PGijsbers: Done.

sebhrusen (approved):
Looks good, thanks @PGijsbers

PGijsbers:
And thank you for the feedback!

PR #258, opened 2021-03-04
Escalate logic for hard interruptions

sebhrusen:
Various frameworks behave differently when trying to interrupt them after the hard timeout because they spawn subprocesses (Java, R, Python...).
What I call the hard timeout is the time limit after which we give up on the job, it usually corresponds to `max_runtime_seconds` + 1h, where `max_runtime_seconds` is actually the max training time (we may want to rename this in a future version).
The idea is to try weak interruptions first (raise an error in the thread, SIGINT) to be able to fail gracefully and obtain a `results.csv` file with a proper error message, and if the process is still running, then we try to kill it, in which case there may be no results.csv.


PGijsbers:
I think it's a good idea to do this. We should make a mental note to check the time spent training when analysing the results though. `SIGINT` might interrupt search and then produce predictions?
(also could you check slack 😅 ?)

sebhrusen:
@PGijsbers 
> SIGINT might interrupt search and then produce predictions?

Yes, in the `callee` interruption case, it will depend if/how the framework intercepts the signal. That's why I try with the `TimeoutError` first. Do you think I should completely remove the SIGINT signal and only try 
1. TimeoutError
2. SIGQUIT
3. SIGKILL
?

or maybe replace SIGINT with SIGTERM?
Those 2 are almost equivalent except that it's much easier and common for a framework to intercept SIGINT (just need to catch `KeyboardInterruptException`) than SIGTERM. Outside of this both sohuld fail gracefully, and I can easily ensure that SIGTERM is also intercepted at the job level so that the app would still produce a results file, but without any predictions this time.

PGijsbers:
There's also no harm to record the predictions if the framework manages to produce them after all- but we should mark it differently from regular results somehow, so they don't accidentally end up with results that adhere to the set constraints. Saving the results/logs could be helpful.

sebhrusen:
> There's also no harm to record the predictions if the framework manages to produce them after all.

We're talking about the hard timeout here, which is mainly a safety to avoid having the ec2 instances hanging.
Actually SIGINT is a weaker interruption than throwing an error in this case, so I removed it.

PR #261, opened 2021-03-11
Only add SIGQUIT/SIGKILL for non-Windows systems

PGijsbers:
I should have ran the benchmark locally on Windows before accepting #258, turns out it broke on Windows 😓  Signal.SIGQUIT/SIGKILL don't exist on Windows, which raises an exception otherwise. This patch avoids that.

### DIFF
--- a/amlb/job.py
+++ b/amlb/job.py
@@ -9,7 +9,7 @@
 from concurrent.futures import ThreadPoolExecutor, ProcessPoolExecutor
 from enum import Enum, auto
 import logging
-import math
+import platform
 import queue
 import signal
 import threading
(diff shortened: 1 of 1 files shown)


sebhrusen (approved):
Ok for me. I didn't even know we could run the app on Windows, given `setup.sh`.
I thought that, as a Windows users, you were using docker.

Btw the way, did you face a scenario where it went escalating to `SIGQUIT` on Windows?
And if it raised an exception because the OS doesn't support it, maybe we were obtaining  the desired effect :)

PGijsbers:
> I thought that, as a Windows users, you were using docker.

Only when using frameworks, but `constantpredictor` works without issue.

> Btw the way, did you face a scenario where it went escalating to SIGQUIT on Windows?

It triggers when the dictionary is defined (`signal` does not have the `SIGQUIT` attribute on Python for Windows), so unfortunately no cigar 😉

PR #262, opened 2021-03-11
Show constant columns separately from the results

PGijsbers:
With more columns added over time to the results table, the printed table would generally wrap long lines. This makes the table very hard to read. At the same time a lot of the information is redundant. This table separately prints some constant information so the variable data in the table can be displayed neatly on more screens.

This takes the output from:
image

to:

image

I would also consider the following changes:
 - don't show the result for each metric but only the one optimized for (or otherwise print them with lower precision),
 - rename some columns for brevity, e.g. `duration` columns to `time (s)` `train (s)` and `predict (s)`
 - drop `seed`
 
 None of these changes would affect the stored data.

### DIFF
--- a/amlb/benchmark.py
+++ b/amlb/benchmark.py
@@ -265,7 +265,15 @@ class Benchmark:
         if rconfig().results.save:
             self._save(board)
 
-        log.info("Summing up scores for current run:\n%s", board.as_printable_data_frame().dropna(how='all', axis='columns').to_string())
+        df = board.as_printable_data_frame().dropna(how='all', axis='columns')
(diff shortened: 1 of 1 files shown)


sebhrusen (changes requested):
could this logic be defined in `results.py` for  consistency?

This is the file where the columns are declared, filled, formatted, and we shouldn't have to look at other files when we want to change the list of columns.

sebhrusen:
@PGijsbers I need this when testing https://github.com/openml/automlbenchmark/pull/293, also getting annoyed by the wide printed frame, so will include it there.

PR #263, opened 2021-03-12
add mlnet.cli to automl benchmark

LittleLittleCloud:
[mlnet](https://github.com/dotnet/machinelearning) is a C# machinelearning library, [mlnet.cli](https://docs.microsoft.com/en-us/dotnet/machine-learning/automate-training-with-cli) is an automl cli tool implemented by mlnet.

#265


PGijsbers:
Thank you for contributing to the benchmark project! Great to see a new tool (and language) represented 🎉 
Please indicate when this PR is ready for review by leaving a request through Github (through the "request review" on the right). If there are any questions, don't hesitate to ask.

Don't forget to update the documentation as well! We want to keep a little description with links to relevant literature/docs of each framework in at [docs/automl_overview.md](https://github.com/openml/automlbenchmark/blob/master/docs/automl_overview.md).

LittleLittleCloud:
Hi @PGijsbers 
Thanks, I'm really excited to be able to contribute a new framework to automl benchmark.

Currently all the tests for local mode have been passed, could you show me how to test in docker and aws mode? Specifically for aws mode,, do I need to open a aws account for that

PGijsbers:
If you don't have access/credits for `aws` you feel comfortable using, then we can do a test run. 
As long as `python runbenchmark.py my_framework validation -m docker -s force` runs successfully, it should be good to go (if you're on Windows you might need to wait for #269 to merge).

frameworks/MLNet/exec.py
```diff
@@ -0,0 +1,116 @@
...
+    columns=['column_{}'.format(i) for i in range(column_num)]
+    train_df = pd.DataFrame(dataset.train.X, columns=columns)
+    train_df['label'] = dataset.train.y
+    test_df = pd.DataFrame(dataset.test.X, columns=columns)
+    test_df['label'] = dataset.test.y
+    train_dataset = os.path.join(temp_output_folder, 'train.csv')
+    test_dataset = os.path.join(temp_output_folder, 'test.csv')
+    log.info("saving train to {}".format(train_dataset))
+    train_df.to_csv(train_dataset, index=False, header=True)
+    test_df.to_csv(test_dataset, index=False, header=True)
```
sebhrusen, line 43: does MLNET support ARFF files?
If so you can access the files path directly using `dataset.train.path`, as well as training information like `dataset.target`, `dataset.predictors`... 

Those work not only with OpenML but also when using a custom file (which could be a csv...).
So it's much preferable to pass the files directly to MLNet and avoid all those data conversion that can be very expensive for larger datasets.
If MLNet doesn't support `arff` files, but only `csv`, I would suggest to do the conversion only for those.

frameworks/MLNet/exec.py
```diff
@@ -0,0 +1,116 @@
...
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    DOTNET_INSTALL_DIR = os.path.join(dir_path, '.dotnet')
+    os.environ['MODELBUILDER_AUTOML'] = 'NNI'
+    os.environ['DOTNET_ROOT'] = DOTNET_INSTALL_DIR
+    mlnet = os.path.join(DOTNET_INSTALL_DIR, 'mlnet')
+    save_metadata(config)
+    name = config.name
+    temp_output_folder = os.path.join(config.output_dir,name)
+    if not os.path.exists(temp_output_folder):
+        os.mkdir(temp_output_folder)
```
sebhrusen, line 29:
```suggestion
    # from frameworks.shared.callee import output_subdir
    temp_output_folder = output_subdir(name, config)
```
this standardizes folder structure and ensures that you get a log file for each task/fold.

frameworks/MLNet/exec.py
```diff
@@ -0,0 +1,116 @@
...
+    train_df['label'] = dataset.train.y
+    test_df = pd.DataFrame(dataset.test.X, columns=columns)
+    test_df['label'] = dataset.test.y
+    train_dataset = os.path.join(temp_output_folder, 'train.csv')
+    test_dataset = os.path.join(temp_output_folder, 'test.csv')
+    log.info("saving train to {}".format(train_dataset))
+    train_df.to_csv(train_dataset, index=False, header=True)
+    test_df.to_csv(test_dataset, index=False, header=True)
+    with Timer() as training:
+        cmd = '{} {}'.format(mlnet, sub_command)
```
sebhrusen, line 45: feel free to use f-strings everywhere if you think it improves readability, also consecutive strings are automatically concatenated in Python:
```python
cmd = f"{mlnet} {sub_command}"\
      f" -- dataset {train_dataset} --test-dataset {test_dataset}"
```
or you can use a single multiline:
```python
cmd = f"""{mlnet} {sub_command}
--dataset {train_dataset} --test-dataset {test_dataset}
--train-time {train_time_in_seconds}
"""
```

frameworks/MLNet/exec.py
```diff
@@ -0,0 +1,116 @@
...
+        if config.type == 'regression':
+            predict_cmd = '{} {}'.format(mlnet, 'predict')
+            predict_cmd += ' --model {} --dataset {} --task-type regression'.format(model_path, test_dataset)
+            predict_cmd += ' > {}'.format(output_prediction_txt)
+            with Timer() as prediction:
+                run_cmd(predict_cmd)
+            prediction_df = pd.read_csv(output_prediction_txt)
+            rename_df = prediction_df.rename(columns={'Score':'predictions'})
+            rename_df['truth'] = dataset.test.y
+            rename_df.to_csv(config.output_predictions_file)
```
sebhrusen, line 106: can't you use `save_predictions` like you did for classification?
This guarantees better compatibility for various kind of datasets.

sebhrusen (changes requested):

resources/frameworks.yaml
```diff
@@ -155,8 +155,11 @@ TPOT:
 #    population_size: 25
 #    verbosity: 2
 
-
-
+MLNet:
+  version: '16.5.8-dev'
+  setup_args: '16.5.8-dev /mnt/c/Users/xiaoyuz/source/repos/machinelearning-tools/artifacts/packages/Release/Shipping/'
```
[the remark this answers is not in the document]
LittleLittleCloud: ### Resolved

.gitignore
```diff
@@ -41,6 +41,10 @@ _site/
 Singularityfile
 *.sif
 
+# Dotnet
+**/.dotnet/**
+**/dotnet-install.sh**
+
```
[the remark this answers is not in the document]
LittleLittleCloud: ### Resolved

frameworks/MLNet/setup.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+HERE=$(dirname "$0")
+MLNET='mlnet'
+VERSION="$1"
+DOTNET_INSTALL_DIR="$HERE/.dotnet"
```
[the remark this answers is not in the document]
LittleLittleCloud: ### Resolved

frameworks/MLNet/setup.sh
```diff
@@ -0,0 +1,23 @@
...
+MLNET="$DOTNET_INSTALL_DIR/mlnet"
+DOTNET="$DOTNET_INSTALL_DIR/dotnet"
+SOURCE="$2"
+# install mlnet if necessary
+if [[ ! -x "$MLNET" ]]; then
+    if [[ ! -x "$DOTNET" ]]; then
+        wget https://dot.net/v1/dotnet-install.sh   
+        ./dotnet-install.sh -c Current --install-dir "$DOTNET_INSTALL_DIR" -Channel 3.1
```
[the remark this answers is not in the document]
LittleLittleCloud: ### Resolved

frameworks/MLNet/exec.py
```diff
@@ -0,0 +1,116 @@
...
+
+            return dict(
+                training_duration=training.duration,
+                predict_duration=prediction.duration,
+            )
+        
+
+
+if __name__ == '__main__':
+    call_run(run)
```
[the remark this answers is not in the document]
LittleLittleCloud: ### Resolved

frameworks/MLNet/exec.py
```diff
@@ -0,0 +1,116 @@
...
+from amlb.data import Dataset
+from amlb.datautils import reorder_dataset
+from amlb.results import NoResultError, save_predictions
+from amlb.utils import dir_of, path_from_split, run_cmd, split_path, Timer
+import json
+from frameworks.shared.callee import call_run, result, save_metadata
+
+log = logging.getLogger(__name__)
+
+def run(dataset: Dataset, config: TaskConfig):
```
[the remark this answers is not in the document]
LittleLittleCloud: ### Resolved

frameworks/MLNet/exec.py
LittleLittleCloud: Unfortunately, MLNet.cli only support csv currently, so I guess the only thing I can do here is to avoid conversion when dataset's format is csv? Could you point out the code/property I can use to determine the file format?

frameworks/MLNet/exec.py
LittleLittleCloud: Resolved

frameworks/MLNet/exec.py
LittleLittleCloud: ### Resolved

frameworks/MLNet/exec.py
LittleLittleCloud: ### Resolved

LittleLittleCloud:
hi @sebhrusen, @PGijsbers Any follow up to this PR :-)

sebhrusen:
@LittleLittleCloud sorry, very very busy lately, I'll have a look at it from Wed.

frameworks/MLNet/exec.py
sebhrusen: you could check the extension of `dataset.train.path` for example

frameworks/MLNet/exec.py
```diff
@@ -0,0 +1,96 @@
...
+    if config.type not in avaible_task_list:
+        raise ValueError(f'{config.type} is not supported.')
+
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    DOTNET_INSTALL_DIR = os.path.join(dir_path, 'lib')
+    os.environ['MODELBUILDER_AUTOML'] = 'NNI'
+    os.environ['DOTNET_ROOT'] = DOTNET_INSTALL_DIR
+    os.environ['MLNetCLIEnablePredict'] = 'True'
+    mlnet = os.path.join(DOTNET_INSTALL_DIR, 'mlnet')
+    temp_output_folder = output_subdir('', config=config)
```
sebhrusen, line 29: there are lot of things going into this temp folder, and as it is inside the output dir, all the files created there get uploaded to s3 at the end of the benchmark when it runs on aws, and at the end, downloaded to the machine that started the benchmark, so we want to avoid downloading unnecessary artifacts.

I would recommend to create this tmp folder using:
```
with tempfile.TemporaryDirectory() as tmpdir:
     ...
 ```
 to ensure that the content gets deleted at the end.
 
 if you want to keep the logs, or various artifacts, we accept a param named `_save_artifacts` (see `autosklearn` for an example), for example a user creating its custom framework:
 ```
 myMLNet:
    extends: MLNet
    params:
       _save_artifact = ['logs', 'models']
```
would be able to retrieve the framework logs and the serialized models.
In your case it would mean sth like:
```python
log_dir = output_subdir('logs', config=config) if 'logs' in artifacts else os.path.join(tmpdir, 'logs')
mlnet_output_dir = output_subdir('models', config=config) if 'models' in artifacts else os.path.join(tmpdir, 'models')
```
this way the `config.output_dir` is not polluted if the extra artifacts were not requested by the user.

frameworks/MLNet/exec.py
```diff
@@ -0,0 +1,96 @@
...
+        columns=[f'column_{i}' for i in range(column_num)]
+        train_df = pd.DataFrame(dataset.train.X, columns=columns)
+        train_df['label'] = dataset.train.y
+        test_df = pd.DataFrame(dataset.test.X, columns=columns)
+        test_df['label'] = dataset.test.y
+
+        log.info(f'saving train to {train_dataset}')
+        train_df.to_csv(train_dataset, index=False, header=True)
+        log.info(f'saving test to {test_dataset}')
+        test_df.to_csv(test_dataset, index=False, header=True)
```
sebhrusen, line 46: a simpler alternative probably to this:
- if `dataset.train.path` extension  is `.csv` then it is already a `csv` file and the `label` column is `dataset.target.name`.
- if `dataset.train.path` extension  is `.arff` then you could use the fact that ARFF format is a superset of CSV, which would allow you to convert it to csv format without having to load it in memory: I can help providing a utility function for this if you want. We know the column names and we know that after line `@DATA` it's basically a csv file.

sebhrusen (commented):
@LittleLittleCloud, thanks for the previous changes.
Now adding new comments focusing on standardizing the integration logic.

Otherwise, I could test the benchmark locally, the setup on macOS was smooth and the test run worked great!

Did you try to run the `validation` to see if it handles some weird datasets correctly?
For a complete validation, I recommend:
```python runbenchmark.py mlnet validation test -m docker -s force```

File: resources/frameworks.yaml [hunk 13]
```diff
@@ -155,8 +155,10 @@ TPOT:
 #    population_size: 25
 #    verbosity: 2
 
-
-
+MLNet:
+  version: '16.5.10'
```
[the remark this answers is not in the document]
LittleLittleCloud: `MLNet` is still in pre-release state so we don't have a `stable` version yet. Can I treat `stable` the same with `latest` for now.

frameworks/MLNet/exec.py
LittleLittleCloud: I can make the change to avoid creating a new csv file from original dataset if dataset's extension is `.csv`. However, if the original dataset's extension is `.arff`, since MLNet can't load .arff dataset, we still need to save the converted dataset into a separate copy so it will still be loaded into memory. 

However, if you just want to avoid creating `Dataframe` because it's a null operation for MLNet and it's very costy for large dataset, and you want to replace it with a cheaper operation. I can follow what you suggested and it would be great if you can provide function on how to do the conversion

frameworks/MLNet/exec.py
LittleLittleCloud: Sure

frameworks/MLNet/exec.py
```diff
@@ -0,0 +1,96 @@
...
+import math
+import os
+import pandas as pd
+from amlb.benchmark import TaskConfig
+from amlb.data import Dataset
+from amlb.datautils import reorder_dataset
+from amlb.results import NoResultError, save_predictions
+from amlb.utils import dir_of, path_from_split, run_cmd, split_path, Timer
+import json
+from frameworks.shared.callee import call_run, result, save_metadata,  output_subdir
```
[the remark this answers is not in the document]
LittleLittleCloud: I'm not quite following, do you want me to reorder the import statement?

frameworks/MLNet/exec.py
LittleLittleCloud: ### Resolved

resources/frameworks.yaml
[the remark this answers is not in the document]
LittleLittleCloud: ### Resolved

frameworks/MLNet/setup.sh
```diff
@@ -0,0 +1,26 @@
...
+    if [[ ! -x "$DOTNET" ]]; then
+        wget -P "$DOTNET_INSTALL_DIR" https://dot.net/v1/dotnet-install.sh
+        chmod +x "$DOTNET_INSTALL_DIR/dotnet-install.sh"
+        "$DOTNET_INSTALL_DIR/dotnet-install.sh" -c Current --install-dir "$DOTNET_INSTALL_DIR" -Channel 3.1
+    fi
+
+    $DOTNET --version
+    $DOTNET tool install mlnet --add-source "$SOURCE" --version "$VERSION" --tool-path "$DOTNET_INSTALL_DIR"
+else
+$DOTNET tool update mlnet --add-source "$SOURCE" --version "$VERSION" --tool-path "$DOTNET_INSTALL_DIR"
```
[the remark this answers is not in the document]
LittleLittleCloud: ### Resolved

frameworks/MLNet/exec.py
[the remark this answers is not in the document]
sebhrusen: yes, grouping + reordering please, thank you.

frameworks/MLNet/exec.py
sebhrusen: > However, if you just want to avoid creating Dataframe because it's a null operation for MLNet and it's very costy for large dataset, and you want to replace it with a cheaper operation. I can follow what you suggested and it would be great if you can provide function on how to do the conversion

Yes, that's what I mean, I'll provide the function soon.

frameworks/MLNet/exec.py
```diff
@@ -0,0 +1,117 @@
...
+        tmpdir = output_subdir('artifacts', config=config)
+
+    train_dataset_path: str
+    test_dataset_path: str
+    label: str = 'label'
+    temp_output_folder = os.path.join(tmpdir, str(config.fold))
+    log_path = os.path.join(temp_output_folder, 'log.txt')
+    label = dataset.target.index
+
+    if dataset.train.format == 'csv':
```
LittleLittleCloud, line 53: when it's 'csv', simply use the dataset path to load data.
when it's 'arff', use `dataset.train.data` to avoid loading data twice.

frameworks/MLNet/exec.py
```diff
@@ -0,0 +1,117 @@
...
+        test_dataset_path = dataset.test.path
+    else:
+        # .arff
+        train_dataset_path = os.path.join(tmpdir, f'train.csv')
+        test_dataset_path = os.path.join(tmpdir, f'test.csv')
+        columns = [f'col_{i}' for i in range(dataset.train.data.shape[-1])]
+        columns[label] = 'label'
+
+        # might use a bit extra memory, but it's faster
+        pd.DataFrame(dataset.train.data, columns=columns).to_csv(train_dataset_path, index=None)
```
LittleLittleCloud, line 64: After comparing between 1) directly saving ndarray into csv and 2) create df and call `df.to_csv` instead, it appears that the 2) is still a better option because it's much faster and by avoiding creating a variable to save df, the memory will be released immediately after `df.to_csv` get called. Because `dataset.train.data` is a lazy property, dataset will only be loaded into memory when get called and will only be loaded once so perf issue here should be small.

sebhrusen (approved):
Looks good to me: tested the docker image with validation benchmark, everything worked perfectly!
Thanks for this contribution.
Please address the small comment in `frameworks.yaml` and then I'll merge.

resources/frameworks.yaml
```diff
@@ -155,8 +155,10 @@ TPOT:
 #    population_size: 25
 #    verbosity: 2
 
-
-
+MLNet:
+  version: 'latest'
+  description: |
+    MLNET.CLI is a automated machine learning tool implemented by ml.net.
```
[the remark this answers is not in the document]
LittleLittleCloud: ## Resolved

PR #269, opened 2021-03-16
Docker for Windows

PGijsbers:
There were a few issues with using docker mode `-m docker` in Windows.
The main pitfalls were:
 - default git behavior checking out `.sh` files locally with `crlf` ending, which can't be executed by a bash shell.
 - live output monitoring using `select` which is not supported for file descriptors in Windows.
 
 I could not find a proper supported way to monitor process output non-blocking in Windows (without resorting to a new package), so I made a custom function based on various solutions proposed on StackOverflow.

Also fixes a bug where `git` was not detected due to `2>/dev/null` in a command, which does not work on Windows.
If that error ignoring was desirable, I can make it platform specific (I think my fix preserves the intended functionality, but I might have missed the point).

Tested with `python runbenchmark.py h2oautoml openml/t/59 -f 0 -m docker -s force ` in Windows (it would be appreciated if you could run it on a non-Windows platform to make sure I did not break anything, @sebhrusen).

### DIFF
--- a/amlb/resources.py
+++ b/amlb/resources.py
@@ -8,6 +8,7 @@ import os
 import random
 import re
 import sys
+import warnings
 
 from amlb.benchmarks.parser import benchmark_load
 from amlb.frameworks import default_tag, load_framework_definitions
@@ -72,9 +73,21 @@ class Resources:
                 return defval
 
         na = "NA"
-        version = git("--version")
-        is_git_repo = version and git("rev-parse --git-dir 2> /dev/null")
-        if is_git_repo:
+        git_version = git("--version")
+        if git_version is None:
+            warnings.warn(
+                "Cannot identify git version. Is git in path?",
+                RuntimeWarning
+            )
+
+        is_repo = git("rev-parse") is not None
+        if not is_repo:
+            warnings.warn(
+                "Directory '{}' is not a git repository.".format(os.getcwd()),
+                RuntimeWarning
+            )
+
+        if git_version and is_repo:
             repo = git("remote get-url origin", na)
             branch = git("rev-parse --abbrev-ref HEAD", na)
             commit = git("rev-parse HEAD", na)
--- a/amlb/resources.py
+++ b/amlb/resources.py
@@ -74,18 +74,7 @@ class Resources:
 
         na = "NA"
         git_version = git("--version")
-        if git_version is None:
-            warnings.warn(
-                "Cannot identify git version. Is git in path?",
-                RuntimeWarning
-            )
-
         is_repo = git("rev-parse") is not None
-        if not is_repo:
-            warnings.warn(
-                "Directory '{}' is not a git repository.".format(os.getcwd()),
-                RuntimeWarning
-            )
 
         if git_version and is_repo:
             repo = git("remote get-url origin", na)
--- a/amlb/resources.py
+++ b/amlb/resources.py
@@ -8,7 +8,6 @@ import os
 import random
 import re
 import sys
-import warnings
 
 from amlb.benchmarks.parser import benchmark_load
 from amlb.frameworks import default_tag, load_framework_definitions
--- a/amlb/utils/process.py
+++ b/amlb/utils/process.py
@@ -86,20 +86,14 @@ def as_cmd_args(*args, **kwargs):
 
 def live_output_windows(process: subprocess.Popen, timeout, **ignored) -> Tuple[str, str]:
     """ Custom output forwarder, because select.select is not Windows compatible. """
-    def wrap_if_needed(s):
-        """ Wrap the stream in TextIO if stream produces bytes. """
-        if hasattr(process, "text_mode") and process.text_mode:
-            return s
-        return io.TextIOWrapper(s, encoding="utf-8")
-
     outputs = dict(
         out=dict(
-            stream=wrap_if_needed(process.stdout),
+            stream=process.stdout,
             queue=queue.Queue(),
             lines=[],
         ),
         err=dict(
-            stream=wrap_if_needed(process.stderr),
+            stream=process.stderr,
             queue=queue.Queue(),
             lines=[],
         ),
--- a/amlb/utils/process.py
+++ b/amlb/utils/process.py
@@ -1,5 +1,5 @@
 from concurrent.futures import ThreadPoolExecutor
-from functools import reduce, wraps
+from functools import reduce, wraps, partial
 import inspect
 import io
 import logging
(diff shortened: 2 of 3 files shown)


amlb/utils/process.py
```diff
@@ -83,6 +84,47 @@ def as_cmd_args(*args, **kwargs):
...
+
+    for output in outputs.values():
+        t = threading.Thread(target=forward_output, args=(output["stream"], output["queue"]))
+        t.daemon = True
+        t.start()
+
+    while process.poll() is None:
+        for output in outputs.values():
+            try:
+                line = output["queue"].get(timeout=0.5)
```
sebhrusen, line 120: magic number?
I'd rather use `get_no_wait` here (this will almost always block for nothing on the `err` stream) with some kind of `ready` flag.

something like:
```python
        out=dict(
            stream=wrap_if_needed(process.stdout),
            queue=queue.Queue(),
            ready=threading.Event(),
            lines=[],
        ),

    def forward_output(stream, queue_, ready):
        for line in stream:
            queue_.put(line)
            ready.set()


while process.poll() is None and any([o['ready'].wait() for o in outputs.values()]):
   for output in outputs.values():
       with outpout['ready']._cond:
           if output['ready'].is_set():
               try: 
                    line = output['queue'].get_no_wait()
                    ...
                except queue.Empty:
                       output['ready'].clear()
```

amlb/resources.py
```diff
@@ -72,9 +73,21 @@ def git(cmd, defval=None):
...
+                "Cannot identify git version. Is git in path?",
+                RuntimeWarning
+            )
+
+        is_repo = git("rev-parse") is not None
+        if not is_repo:
+            warnings.warn(
+                "Directory '{}' is not a git repository.".format(os.getcwd()),
+                RuntimeWarning
+            )
```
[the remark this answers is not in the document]
PGijsbers: I had results without git versions, which made me think it would be useful to get a message about why that was (it was actually the `2>/dev/null` failing). But I forgot about the docker mode when writing this. I don't mind a single extra log line in docker personally (but will remove it if you feel strongly about it).

amlb/utils/process.py
PGijsbers: Wouldn't the same problem be caused by the `o['ready'].wait()` call here? By default it's blocking, but even with a `timeout`, we're waiting for the `Event` to be set on a stream which may not produce any output. I can't think of a good way around this.

While I can't think of a way to always respond to the quickest event (without rather busy waiting), you could dynamically adjust wait times: longer for streams that you expect to produce output first, shorter for those that typically don't. But I feel this is some unnecessary over-engineering. I did adjust the code to always empty a queue before starting to wait on the next one.

Regarding the magic number, two questions:
 - why is there both `timeout` and `activity_timeout`? 
 - I don't think either describe the timeout that the polling time should be (because you can rely on `select`), do you propose I make a new parameter for that?

amlb/resources.py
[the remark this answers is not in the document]
sebhrusen: `git` is not a requirement to run the app locally, and its absence doesn't prevent any single feature from working at this level, so I would not emit any warning here.
I consider both error and warning levels as "call to action" levels: call to fix, or call to inspect/confirm.
Here, everything is normal.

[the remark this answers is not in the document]
PGijsbers: Fair enough.

amlb/utils/process.py
sebhrusen: `timeout` is the process timeout, so you can't use this.
`activity_timeout` is more tricky and aims at detecting process inactivity: it is currently used when building docker images or running setup commands. Building an image for example can take a very long time (e.g. autoxgboost) and we want to detect if something's wrong as soon as possibly. So the `activity_timeout` (10min by default) will terminate the process if there's no output during a certain period. So you can't use this either.

> do you propose I make a new parameter for that?

No, much better to have a magic number in this case, it's just almost always a code smell indicating that something could have been done differently.

> Wouldn't the same problem be caused by the o['ready'].wait() call here? By default it's blocking, but even with a timeout, we're waiting for the Event to be set on a stream which may not produce any output. I can't think of a good way around this.

You're right. `Event` is not suitable for this. `Condition` would probably work though as `forward_output` could call `notify_all` instead of `ready.set()`.
But you can go with the timeout approach if you want something running quickly. I just like multithreading issues too much for not trying to solve them without using timeouts.

amlb/utils/process.py
```diff
@@ -83,6 +84,75 @@ def as_cmd_args(*args, **kwargs):
...
+                try:
+                    line = output["queue"].get(timeout=0.5)
+                    output["lines"].append(line)
+                    print(line.rstrip())
+                except queue.Empty:
+                    break
+    return ''.join(outputs["out"]["lines"]), ''.join(outputs["err"]["lines"])
+
+
+def live_output_unix(process, timeout, input=None, mode='line', **ignored):
```
sebhrusen, line 123: please keep `activity_timeout` for this parameter, and default to `None`.
`timeout` would mean that a call to this function would time out after `timeout`: this is not the case.
However, each time we're trying to read a pipe, if one read times out, then the entire function ends.

amlb/utils/process.py
```diff
@@ -112,39 +182,10 @@ def run_cmd(cmd, *args, **kwargs):
...
-
-        output, error = zip(*iter(lambda: read_pipe([process.stdout if process.stdout else 1,
-                                                     process.stderr if process.stderr else 2], params.activity_timeout),
-                                  ['', '']))
-        print()  # ensure that the log buffer is flushed at the end
-        return ''.join(output), ''.join(error)
+    if platform.system() == "Windows":
+        live_output = partial(live_output_windows, timeout=params.activity_timeout)
+    else:
+        live_output = partial(live_output_unix, timeout=params.activity_timeout)
```
sebhrusen, line 188:
```suggestion
        live_output = partial(live_output_unix, mode=params.live_output, activity_timeout=params.activity_timeout)
```

amlb/resources.py
```diff
@@ -8,6 +8,7 @@
 import random
 import re
 import sys
+import warnings
```
sebhrusen, line 11: unused

amlb/utils/process.py
sebhrusen: also you dropped the `mode` logic:
```python
        if mode is True:
            mode = 'line'
```

amlb/utils/process.py
PGijsbers: Thanks for the help!

PGijsbers:
@sebhrusen if you can verify I did not break unix support, I'll go ahead and merge.

sebhrusen (approved):
looks good

PR #270, opened 2021-03-17
add FLAML to automl benchmark

qingyun-wu:
FLAML is a lightweight Python library that finds accurate machine learning models automatically, efficiently, and economically.


PGijsbers:
Thank you for the contribution! I hope to be able to review over the next few days.
In the meanwhile, would you be able to update [`docs/automl_overview.md`](https://github.com/openml/automlbenchmark/blob/master/docs/automl_overview.md) with a brief description of FLAML following the format that is used for the other frameworks?

resources/frameworks.yaml
```diff
@@ -42,14 +51,15 @@ autosklearn:
     It leverages recent advantages in Bayesian optimization, meta-learning and ensemble construction.
   project: https://automl.github.io/auto-sklearn/
   refs: [http://papers.nips.cc/paper/5872-efficient-and-robust-automated-machine-learning.pdf]
-#  params:
-#    _save_artifacts: ['models', 'debug_as_files']
-#    _n_jobs: 1
+# params:
+#   _save_artifacts: ['models', 'debug_as_files']
+#   _n_jobs: 1
```
sebhrusen, line 56: this and all changes below are irrelevant for this PR.
Please only modify what is related for your framework.

resources/constraints.yaml
```diff
@@ -5,6 +5,22 @@ test:
...
+10m1c:
+  folds: 10
+  max_runtime_seconds: 600
+  cores: 1
+  
+1h1c:
+  folds: 10
+  max_runtime_seconds: 3600
+  cores: 1
+
```
sebhrusen, line 23: irrelevant for this PR.
Please only modify what is related for your framework.

.gitignore
```diff
@@ -44,3 +44,6 @@ Singularityfile
 #whitelist
 !reports/**/*.csv
 !reports/**/*.png
+
+catboost_info/*
+plot/*
```
sebhrusen, line 49:
```suggestion
```

To avoid setting a new precedent, the recommended way for all those 3rd party artifacts is to put them under flaml/setup or flaml/lib depending if they're needed only during setup or also at runtime.
Those 2 folders are already excluded in git.
Besides, it standardizes the folder structure, making each framework easier to understand and maintain.

If those folders are generated at runtime: you need to find a way to redirect the output either:
- in a tmp folder.
- somewhere in the `config.output_dir`.

frameworks/flaml/exec.py
```diff
@@ -0,0 +1,109 @@
...
+import pandas as pd
+import time, threading, _thread
+from amlb.benchmark import TaskConfig
+from amlb.data import Dataset
+
+log = logging.getLogger(__name__)
+
+
+def run(dataset: Dataset, config: TaskConfig):
+    log.info("\n**** FLAML ****\n")
```
sebhrusen, line 33: can you please call `save_metadata(config)` immediately after this line: I want to have a clear view of the `config` object state when it is saved, and therefore it should be done as early as possible.

Also, if you have a way to obtain the runtime version of the `flaml` package and include this in the log message + metadata (see other frameworks for examples), this would be great.

frameworks/flaml/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+                y_enc=dataset.train.y_enc
+            ),
+
+            test=dict(
+                X_enc=X_test_enc,
+                y_enc=dataset.test.y_enc
+            ),
+
+            problem_type=dataset.type.name  
+        )
```
sebhrusen, line 26: Do you still need this or were you just testing?
The encoded data should only be used by frameworks that only support numeric data.

frameworks/flaml/exec.py
```diff
@@ -0,0 +1,109 @@
...
+
+    is_classification = config.type == 'classification'
+    enc = config.framework_params.get('_enc', False)
+    if enc:
+        import numpy as np
+        X_train = dataset.train.X_enc
+        y_train = dataset.train.y_enc
+        X_test = dataset.test.X_enc
+        y_test = dataset.test.y_enc
+        train = label = None
```
sebhrusen, line 49: same question as for `__init__`

frameworks/flaml/requirements.txt
```diff
@@ -0,0 +1,3 @@
+liac-arff==2.4
+openml==0.10.2
+boto3>=1.9,<2.0
```
sebhrusen, line 3: are those really needed at runtime??

frameworks/flaml/exec.py
```diff
@@ -0,0 +1,109 @@
...
+os.environ['NUMEXPR_NUM_THREADS'] = '1'
+import tempfile as tmp
+import warnings
+from flaml import AutoML
+from frameworks.shared.callee import call_run, result, output_subdir, utils
+from amlb.utils import Timer
+import pandas as pd
+import time, threading, _thread
+from amlb.benchmark import TaskConfig
+from amlb.data import Dataset
```
sebhrusen, line 27: please reorganize and group imports for readability, from most generic to most specific
```python
import core_python

import third_parties

import your_framework

import amlb
import frameworks.shared.callee
```

sebhrusen (changes requested):
Qingyun, thanks for this new addition.
Added a few remarks for now to keep consistency with other frameworks: that's also what allow us to make changes easier when we need to add a new feature for example.

sebhrusen:
@qingyun-wu Regarding the new constraints, I understand that FLAML is optimized for small resources (time + cpu).
I'll be happy to add the new constraints in a separate PR if you provide the rationale behind it.
Until then, you can also provide custom constraints when you use the app: see https://github.com/openml/automlbenchmark/blob/master/docs/HOWTO.md#constraints-definition

resources/constraints.yaml
qingyun-wu: Sure. Thanks for the reminder! I am adding this because FLAML is optimizing for small resources. But I agree with what you suggested: can do this in a separate PR.

frameworks/flaml/requirements.txt
qingyun-wu: removed

frameworks/flaml/__init__.py
qingyun-wu: removed. Thanks!

frameworks/flaml/exec.py
qingyun-wu: done. Thanks!

frameworks/flaml/exec.py
qingyun-wu: revised. Thank you!

frameworks/flaml/exec.py
qingyun-wu: removed. Thank you!

frameworks/flaml/exec.py
```diff
@@ -0,0 +1,80 @@
...
+
+from flaml import AutoML, __version__
+
+from frameworks.shared.callee import call_run, result, save_metadata
+
+log = logging.getLogger(__name__)
+
+
+def run(dataset, config):
+    log.info("\n**** FLAML ****\n")
```
sebhrusen, line 15:
```suggestion
    log.info(f"\n**** FLAML [v{__version__}] ****\n")
```

frameworks/flaml/exec.py
```diff
@@ -0,0 +1,80 @@
...
+    column_names, _ = zip(*dataset.columns)
+    column_types = dict(dataset.columns)
+    train = pd.DataFrame(dataset.train.data, columns=column_names).astype(
+        column_types, copy=False)
+    label = dataset.target.name
+    X_train = y_train = None
+    test = pd.DataFrame(dataset.test.data, columns=column_names).astype(
+        column_types, copy=False)
+    X_test = test.drop(columns=label)
+    y_test = test[label]
```
sebhrusen, line 32: sorry that you have to do that.
I'll work very soon on switching the `amlb` app to `pandas` instead of `numpy`: this should simplify many things, bring support for `string` columns for frameworks that support them, and also ensure that all frameworks get the same data and metadata.

frameworks/flaml/exec.py
```diff
@@ -0,0 +1,80 @@
...
+    if perf_metric is None:
+        log.warning("Performance metric %s not supported.", config.metric)
+
+    training_params = {k: v for k, v in config.framework_params.items()
+     if not k.startswith('_')}
+
+    with Timer() as training:
+        aml.fit(X_train, y_train, train, label, perf_metric, config.type, 
+            n_jobs=n_jobs, 
+            log_file_name=config.output_predictions_file.replace('.csv','.log'),
```
sebhrusen, line 63: we're trying to put logs in a dedicated folder:
```python
from frameworks.shared.callee import  …,  output_subdir
…
log_dir = output_subdir("logs", config)  # this creates a subfolder unique for the given task/fold
…
log_file_name=os.path.join(log_dir, "flaml.log")
```

resources/frameworks.yaml
```diff
@@ -18,6 +18,15 @@ __dummy_framework_with_defaults:
 ### AutoML frameworks ###
 #########################
 
+flaml:
+  version: '0.2.3'
```
sebhrusen, line 22: same remark as https://github.com/openml/automlbenchmark/pull/263#discussion_r601859022

frameworks/flaml/setup.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+HERE=$(dirname "$0")
+#AMLB_DIR="$1"
+# by passing the module directory to `setup.sh`, it tells it to automatically create a virtual env under the current module.
+# this virtual env is then used to run the exec.py only, and can be configured here using `PIP` and `PY` commands.
+. $HERE/../shared/setup.sh $HERE
+#. $AMLB_DIR/frameworks/shared/setup.sh $HERE
+PIP install  flaml==0.2.3
```
sebhrusen, line 8: see comment in `frameworks.yaml` and see `RandomForest/setup.sh` for a template supporting multiple versions.
Maybe we will find a way to consolidate this later.

sebhrusen (commented):
@qingyun-wu Thanks for this, I like the simplicity of the integration.
I just have small requirements, especially one regarding the support for multiple versions: we want to have `frameworks.yaml` dinamically pointing to the last stable version by default, and be able to specify a fixed numbered version in some cases.

frameworks/flaml/exec.py
qingyun-wu: Updated. Thanks for the suggestion!

resources/frameworks.yaml
qingyun-wu: Resolved! Thanks!

frameworks/flaml/setup.sh
qingyun-wu: Revised accordingly. Thanks for the pointer.

frameworks/flaml/exec.py
qingyun-wu: Thanks for the info. I have revised accordingly.

frameworks/flaml/setup.sh
```diff
@@ -1,8 +1,26 @@
 #!/usr/bin/env bash
 HERE=$(dirname "$0")
-#AMLB_DIR="$1"
+AMLB_DIR="$1"
+VERSION=${2:-"stable"}
+REPO=${3:-"https://github.com/microsoft/FLAML.git"}
+PKG=${4:-"flaml"}
+if [[ "$VERSION" == "latest" ]]; then
+    VERSION="master"
```
sonichi, line 8: main

qingyun-wu: thanks for catching it. revised!

sebhrusen (commented):
tested docker image + ran validation test: it looks good.
Please apply the changes for predict duration before I can merge.

The request for `catboost` tmp folder can wait a bit, but would be nice to have for local runs, thanks.

qingyun-wu:
Thanks sebhrusen for your mering and the suggestion! Will resolve the catboost_info issue soon.

PR #271, opened 2021-03-23
Add 'add documentation' step to PR instructions

PGijsbers:
Also fix an error where the nightly_h2o build did not actually contain a setup_cmd that was referred to.

I didn't include the old description since it's not in the `examples/custom/frameworks.yaml`.

### DIFF
--- a/docs/HOWTO.md
+++ b/docs/HOWTO.md
@@ -405,9 +405,9 @@ autosklearn_oldgen:
   description: "this will use the latest autosklearn version from the old generation"
 
 H2OAutoML_nightly:
-  extends: H2OAutoML
-  version: 'latest'
-  description: "this will use the nightly H2O build if there's no H2O version already installed (use --setup=force to refresh the version)"
+  module: frameworks.H2OAutoML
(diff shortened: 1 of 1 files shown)


docs/HOWTO.md
```diff
@@ -405,9 +405,9 @@ autosklearn_oldgen:
   description: "this will use the latest autosklearn version from the old generation"
 
 H2OAutoML_nightly:
-  extends: H2OAutoML
-  version: 'latest'
-  description: "this will use the nightly H2O build if there's no H2O version already installed (use --setup=force to refresh the version)"
+  module: frameworks.H2OAutoML
+  setup_cmd: 'LATEST_H2O=`curl http://h2o-release.s3.amazonaws.com/h2o/master/latest` && pip install --no-cache-dir -U "http://h2o-release.s3.amazonaws.com/h2o/master/${{LATEST_H2O}}/Python/h2o-3.29.0.${{LATEST_H2O}}-py2.py3-none-any.whl"'
+  version: 'nightly'
```
sebhrusen, line 410: sorry for that, I simplified this by supporting the `latest` version in h2o's `setup.sh`, but you're right, it's probably better to keep it to document the `setup_cmd` param.

PGijsbers: No problem :) Agreed that for the actual use case there is a better way to do it (now), but for documentation purposes we need _something_.

PR #274, opened 2021-03-28
Fixed AutoGluon model files saving to S3 on exception

Innixma:
Followup to #237 

Based on testing I've done with the `automlbenchmark/runbenchmark.py AutoGluon_bestquality test -f 0` this should resolve any issues related to uploading large files to S3 without specifying `'models'` in the save artifacts.

One thing I've noticed (has always been there but was wondering if it could be resolved):

When training AG, I see some log output that isn't coming from autogluon that breaks up logging:

```
Fitting model: XGBoost_BAG_L1 ... Training model for up to 43.66s of the 43.66s of remaining time.
	0.8115	 = Validation roc_auc score
	0.61s	 = Training runtime
	0.01s	 = Validation runtime
Fitting model: NeuralNetMXNet_BAG_L1 ... Training model for up to 42.98s of the 42.98s of remaining time.
{ 'columns': [ ['loc', 'float'],
	0.8011	 = Validation roc_auc score
	6.91s	 = Training runtime
	0.41s	 = Validation runtime
Fitting model: NeuralNetFastAI_BAG_L1 ... Training model for up to 35.64s of the 35.64s of remaining time.
               ['v(g)', 'float'],
               ['ev(g)', 'float'],
               ['iv(g)', 'float'],
               ['n', 'float'],
               ['v', 'float'],
               ['l', 'float'],
               ['d', 'float'],
               ['i', 'float'],
               ['e', 'float'],
               ['b', 'float'],
               ['t', 'float'],
               ['lOCode', 'float'],
               ['lOComment', 'float'],
               ['lOBlank', 'float'],
               ['lOCodeAndComment', 'float'],
               ['uniq_Op', 'float'],
               ['uniq_Opnd', 'float'],
               ['total_Op', 'float'],
               ['total_Opnd', 'float'],
               ['branchCount', 'float'],
               ['problems', 'object']],
  'problem_type': 'binary',
  'target': {'classes': ['no', 'yes'], 'name': 'problems'},
  'test': {'data': '/tmp/tmpjmn2x06a/test.data.npy'},
  'train': {'data': '/tmp/tmpjmn2x06a/train.data.npy'}}
```

This appears to be coming from some kind of `print(dataset)` call, but I don't see where it occurs and I explicitly call `del dataset` prior to training. Is there a reason this information is being logged in the middle of model training? I'd prefer to either have it be prior to model training or removed entirely.

### DIFF
--- a/frameworks/AutoGluon/exec.py
+++ b/frameworks/AutoGluon/exec.py
@@ -3,6 +3,7 @@ import os
 import shutil
 import warnings
 import gc
+import tempfile
 from io import StringIO
 warnings.simplefilter("ignore")
 
@@ -53,18 +54,19 @@ def run(dataset, config):
         column_names, _ = zip(*dataset.columns)
         column_types = dict(dataset.columns)
         train = pd.DataFrame(dataset.train.data, columns=column_names).astype(column_types, copy=False)
-        print(f"Columns dtypes:\n{train.dtypes}")
+        log.info(f"Columns dtypes:\n{train.dtypes}")
         test = pd.DataFrame(dataset.test.data, columns=column_names).astype(column_types, copy=False)
 
     del dataset
     gc.collect()
 
-    output_dir = output_subdir("models", config)
+    models_dir = tempfile.mkdtemp() + os.sep  # passed to AG
+
     with utils.Timer() as training:
         predictor = TabularPredictor(
             label=label,
             eval_metric=perf_metric.name,
-            path=output_dir,
+            path=models_dir,
             problem_type=problem_type,
         ).fit(
             train_data=train,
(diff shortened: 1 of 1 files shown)


sebhrusen (approved):
Looks good, thanks @Innixma

sebhrusen:
@Innixma regarding the logging issue.
> This appears to be coming from some kind of print(dataset) call, but I don't see where it occurs and I explicitly call del dataset prior to training. Is there a reason this information is being logged in the middle of model training? I'd prefer to either have it be prior to model training or removed entirely.

that's really weird, there is a `print(params.dataset)` indeed in `shared/callee.py`, but called before calling the `run` function, so I don't understand why it get mixed up with AG logs, probably some sort of race condition with the console handler: I'll replace the `print` with a call to `log.info`, or maybe simply remove it if it's not that useful (maybe it was added for debugging at some point, although it can be useful for framework maintainers to see quickly the params they have access to).

sebhrusen:
@Innixma yes, calling `log.info` fixed it (https://github.com/openml/automlbenchmark/pull/275). Thanks for noticing this.

Also noticed many isntances of the following warning from `PyTorch` apparently?
```
[W ParallelNative.cpp:206] Warning: Cannot set number of intraop threads after parallel work has started or after set_num_threads call when using native parallel backend (function set_num_threads)
[W ParallelNative.cpp:206] Warning: Cannot set number of intraop threads after parallel work has started or after set_num_threads call when using native parallel backend (function set_num_threads)
```
 Is it because I ran it on macOS
This is polluting your logs, maybe you want to consider disable.

Innixma:
Hi Seb,

Thanks for fixing the log issue re the dataset!

Regarding the torch log spam, it is because you are running on MacOS. This started happening with torch 1.7, torch 1.6 didn't have any warnings like that. Not sure why, and we weren't able to figure out how to disable the logs. It might be related to our use of the outdated fastai v1. We are planning to update to fastai v2 soon, which we hope will magically solve the log spam issue :) .

If you have any tips on fixing / silencing the warning, would be happy to hear!

PR #276, opened 2021-03-31
Add the possibility to declare a framework as abstract

sebhrusen:
This is mainly useful to define a "template" framework from which children could extend to inherit all its properties without wanting to be able to run the template itself.

A typical use-case is MLPLan that comes with 2 training engines.


Innixma:
Would it also make sense long-term to convert the `def run(dataset, config):` to some kind of `class FrameworkRunner`?

A lot of the data loading code is shared between all of the frameworks, and they can be viewed as individual ML models that take data in and generate output aka `output = model.fit(data, **kwargs)`. This would simplify the work for frameworks to have slightly modified variants of their `exec.py`.

sebhrusen:
https://github.com/openml/automlbenchmark/issues/279

Issue #266, opened 2021-03-15
Add task type to results.csv


sebhrusen:
for easy filtering, add column `task type` among `binary`, `multiclass`, `regression`.

PR #277, opened 2021-03-31
expose problem type column in results.csv

sebhrusen:
https://github.com/openml/automlbenchmark/issues/266


PGijsbers (approved):
Looks reasonable to me. Only tested with one task of each type and `constantpredictor`.

Issue #214, opened 2020-12-08
Add option to error out if resource constraints can not be met


PGijsbers:
The current behavior if the constraint specifies more memory (and cores?) than available on the machine is to log it as a warning. I think it would be useful to have an option which can be turned on which instead stops execution if such violations are identified. Using this option would avoid a scenario where the warning is missed and experiments are run with different resources than expected.

sebhrusen:
@PGijsbers we need to review together the resource constraints logic (esp. memory reserved for the OS, disk space and so on), I'm not really happy with this today: they're not  clear.

PGijsbers:
Added in #288

PR #288, opened 2021-04-14
Improved system requirements check

sebhrusen:
https://github.com/openml/automlbenchmark/issues/214

### DIFF
--- a/amlb/benchmark.py
+++ b/amlb/benchmark.py
@@ -14,10 +14,11 @@ from importlib import import_module, invalidate_caches
 import logging
 import math
 import os
+import re
 import signal
 import sys
 
-from .job import Job, SimpleJobRunner, MultiThreadingJobRunner
+from .job import Job, JobError, SimpleJobRunner, MultiThreadingJobRunner
 from .datasets import DataLoader, DataSourceType
 from .data import DatasetType
 from .resources import get as rget, config as rconfig, output_dirs as routput_dirs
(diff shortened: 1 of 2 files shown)


amlb/benchmark.py
```diff
@@ -360,32 +361,45 @@ def __json__(self):
...
+            handle_unfulfilled(f"Assigned memory ({assigned_mem} MB) is within {sys_mem.available} MB of system total memory {sys_mem.total} MB): "
+                               f"We recommend a {os_recommended_mem} MB buffer, otherwise OS memory usage might interfere with the benchmark task.")
 
         if self.min_vol_size_mb > 0:
             sys_vol = system_volume_mb()
             os_recommended_vol = rconfig().benchmarks.os_vol_size_mb
             if self.min_vol_size_mb > sys_vol.free:
-                log.warning("WARNING: Available volume memory (%(available).f MB / total=%(total).f MB) doesn't meet requirements (%(required).f MB)!",
-                            dict(required=self.min_vol_size_mb+os_recommended_vol, available=sys_vol.free, total=sys_vol.total))
+                handle_unfulfilled(f"Available volume memory ({sys_vol.free} MB / total={sys_vol.total} MB) doesn't meet requirements ({self.min_vol_size_mb+os_recommended_vol} MB)!")
```
PGijsbers, line 402: Should we use something akin to 'disk space' or 'storage' to further avoid confusion?

resources/config.yaml
```diff
@@ -40,6 +40,10 @@ benchmarks:                     # configuration namespace for the benchmarks def
     - '{root}/resources/benchmarks'
   constraints_file:             # list of yaml files describing the benchmarks runtime constraints.
     - '{root}/resources/constraints.yaml'
+  on_unfulfilled_constraint: 'auto'  # one of ('auto', 'warn', 'fail'), used when checking the os resources if one memory/core/volume constraint can not be fulfilled:
+                                     #   if 'auto' benchmark, will fail on important unfulfilled constraints (cores, memory) on non local modes, and warn otherwise.
+                                     #   if 'warn' a warning message will always be emitted;
```
PGijsbers, line 45:
```suggestion
                                     #   if 'warn' only a warning message will be emitted;
```

amlb/benchmark.py
sebhrusen: you mean renaming `min_vol_size_mb` to `min_disk_space_mb`?
I can do that if you think it's better: I recently introduced some logic allowing to rename config entries in a backwards compatible way, so I don't mind renaming.
I also have the idea to remove all those unit suffixes (for time, bytes…) and write:
```
min_disk_space = 64gb
```
instead of
```
min_disk_space_mb = 65536
```
Will probably do this later as it's not urgent, and I want to do it for all units, so maybe it's worth waiting for that task and switch to the new name at the same time.

sebhrusen: > Should we use something akin to 'disk space' or 'storage' to further avoid confusion?

Oh, just realized that you were probably talking about the message: "Available disk space", correct? Yes, it would be less confusing indeed.

PGijsbers: Yes, sorry for being too brief there. The field name was clear, I foresaw misinterpretations of the warning message.

PGijsbers: I think declaring `disk_space = 64gb` in a configuration is nicer (because it allows the user to specify unit size). But I actually kind of like units in variable names if it's otherwise not immediately clear what unit is used. But as long as it's easy enough to deduce what unit is used without running the code I'm ok with anything.

PGijsbers (approved):

Issue #282, opened 2021-04-08
Refactor OpenML integration to leverage OpenML client also to load data.


sebhrusen:
OpenML client is currently used to extract metadata and download data, but for some legacy reason (issues with an older version probably), data are loaded using the `arff` library.
If the OpenML client now provides everything needed, the `amlb` module should also use it to load the data correctly.

PGijsbers:
The latest version (`0.12`) is much more mature than the one we started out with (I believe `0.7`). It includes e.g. pandas support with correct dtypes. OpenML will soon switch over from `arff` to `parquet` by default, which should make data loading quicker/less memory intensive. However we'll still allow `arff` downloads - but perhaps we should see if other frameworks that currently rely on the data file can work with parquet as well (this is a separate issue).

Innixma:
Is there a path forward to attempt an S3-only usage of data loading, to avoid interacting with openml server? I've noticed that running parallel benchmarks or starting trials too quickly will get OpenML errors that wouldn't occur if we were accessing the S3 files directly. This makes using AutoMLBenchmark fairly difficult as about ~1/4th of the benchmark runs I've done have many failures due to the server errors, and if someone else is using AutoMLBenchmark at the same time, we may cause errors in each-other's runs due to overloaded servers.

Innixma:
Example of the exception:

```
[INFO] [amlb:04:52:12.383] Running `AutoGluon_bestquality:latest` on `ag` benchmarks in `local` mode.
[INFO] [amlb.frameworks.definitions:04:52:12.418] Loading frameworks definitions from /repo/resources/frameworks.yaml.
[INFO] [amlb.resources:04:52:12.493] Loading benchmark constraint definitions from /repo/resources/constraints.yaml.
[INFO] [amlb.benchmarks.file:04:52:12.499] Loading benchmark definitions from /repo/resources/benchmarks/ag.yaml.
[INFO] [amlb.job:04:52:12.550] 
-----------------------------------------------------------
Starting job local.ag.4h8c.abalone.0.AutoGluon_bestquality.
[INFO] [amlb.utils.process:04:52:12.550] [local.ag.4h8c.abalone.0.AutoGluon_bestquality] CPU Utilization: 30.7%
[INFO] [amlb.utils.process:04:52:12.552] [local.ag.4h8c.abalone.0.AutoGluon_bestquality] Memory Usage: 2.5%
[...]

PR #293, opened 2021-04-16
Issue 282: leverage OpenML client to load data

sebhrusen:
https://github.com/openml/automlbenchmark/issues/282
also includes https://github.com/openml/automlbenchmark/issues/116 allowing clients to consistently consume pandas Dataframes.

Load data using openML in most cases, and generate arff or csv files on the fly for frameworks requiring files.
This is the exact opposite from the previous logic that was always generating the arff files, and loading them for frameworks asking for ndarray or pandas.

This should avoid type issues on certain columns of the ndarray, and I should be able to suppress the encoding logic in `amlb`: frameworks requiring numerical data with use ndarray and others will use pandas.

the `Dataset` / `DataSplit` abstractions now follow the following contract:
- `data`, `X`, `y` are `pd.DataFrame` objects.
- `data_enc`, `X_enc`, `y_enc` are `np.ndarray` objects (or potentially sparse `scipy` matrices).
- `path` points to the split files in the original file format.
- `data_path(format)` allow clients to ask for files in a specific format (only `csv`, `arff` and `parquet` are currently supported).


amlb/datasets/openml.py
```diff
@@ -139,133 +127,154 @@ def get_values(f):
...
+            test.to_csv(test_path, header=True, index=False)
+        return train_path, test_path
+
+
+__data_splitters__ = [ArraySplitter, DataFrameSplitter, ArffSplitter, CsvSplitter]
+__supported_file_formats__ = ['arff', 'csv']
+
+
+def _get_data_splitter_cls(split_format='array'):
+    return next(ds for ds in __data_splitters__ if ds.format == split_format)
```
PGijsbers, line 277: It's not wrong, but a `StopIteration` error might be a little hard understand (especially for someone integrating with the framework requesting an unavailable split format). Maybe add a friendlier message?

```python
supported = {ds.format: ds for ds in __data_splitters__}
if split_format not in supported :
     raise ValueError(f"No splitter for format '{split_format}' available, only {list(supported)} are supported.")
return supported[split_format]
```

resources/constraints.yaml
```diff
@@ -28,3 +28,8 @@ test:
   cores: 8
   min_vol_size_mb: 131072
 
+# use 256 volume size GB for all 8c constraints
+# 1h8c256v
+
+# double check docker defaults on linux (aws)
+
```
PGijsbers, line 35:
```suggestion
```

:)

docs/HOWTO.md
```diff
@@ -64,6 +64,7 @@ aws:
   resource_files:  # this allows to automatically upload custom config + frameworks to the running instance (benchmark files are always uploaded).
     - '{user}/config.yaml'
     - '{user}/frameworks.yaml'
+    - '{user}/constraints.yaml'
```
PGijsbers, line 67: Just double-checking that this is deliberate since the remainder of the PR is unrelated.

PGijsbers (changes requested):
Good first step 👍 Mostly good for merge I think, but I left a few remarks.

resources/constraints.yaml
sebhrusen: best reminders ever :)

frameworks/AutoGluon/exec.py
```diff
@@ -46,22 +44,14 @@ def run(dataset, config):
...
-    if load_raw:
-        train, test = load_data_raw(dataset=dataset)
-    else:
-        column_names, _ = zip(*dataset.columns)
-        column_types = dict(dataset.columns)
-        train = pd.DataFrame(dataset.train.data, columns=column_names).astype(column_types, copy=False)
-        log.info(f"Columns dtypes:\n{train.dtypes}")
-        test = pd.DataFrame(dataset.test.data, columns=column_names).astype(column_types, copy=False)
-
-    del dataset
```
Innixma, line 59: Are there ways to ensure the data is only present once in memory? This was my attempt previously to do this, but I suspect it actually exists multiple times in memory. A good test would be to see if AutoGluon still crashes on `KDDCup09-Upselling` from OOM before getting to AG code on an m5.2xlarge. If the duplicate instances can be eliminated, this would free up a lot of memory for AutoML systems to work with when the datasets are large.

sebhrusen: running the benchmark with `--profiling` option will provide with some additional logs, especially for main memory changes when loading data.
right before calling the subprocess running the client code, and just after  having rerialized the dataframes, I call 
```
dataset.release()
```
to ensure than the data doen't remain in memory.

Then, on the "callee" side, the data are just pickled and if your code doesn't manipulate the frames (it should not be needed anymore, now that we automatically provide pd.DataFrame with the correct dtypes), then I don't see any reason for duplications.

Innixma: ok, sounds great!

tests/unit/amlb/datasets/file/test_file_dataloader.py
```diff
@@ -0,0 +1,345 @@
...
+    assert not any([p.has_missing_values for p in ds.predictors])
+
+    assert ds.train.path == ds_def.train
+    assert ds.test.path == ds_def.test
+    assert ds.train.data_path('csv') == ds.train.path
+    try:
+        ds.train.data_path('arff')       # file loader doesn't support file auto-conversion yet
+        pytest.fail("should have raised")
+    except ValueError as e:
+        assert "Dataset kc2_train is only available in csv format" in str(e)
```
PGijsbers, line 64: The [`pytest.raises`](https://docs.pytest.org/en/stable/reference.html?highlight=raises#pytest-raises) syntax was made for this :)

tests/unit/amlb/datasets/file/test_file_dataloader.py
```diff
@@ -0,0 +1,345 @@
...
+    assert not any([p.has_missing_values for p in ds.predictors])
+
+    assert ds.train.path == ds_def.train
+    assert ds.test.path == ds_def.test
+    assert ds.train.data_path('arff') == ds.train.path
+    try:
+        ds.train.data_path('csv')       # file loader doesn't support file auto-conversion yet
+        pytest.fail("should have raised")
+    except ValueError as e:
+        assert "Dataset kc2_train is only available in arff format" in str(e)
```
PGijsbers, line 120: [`pytest.raises`](https://docs.pytest.org/en/stable/reference.html?highlight=raises#pytest-raises)

tests/unit/amlb/datasets/file/test_file_dataloader.py
```diff
@@ -0,0 +1,345 @@
...
+    assert not any([p.has_missing_values for p in ds.predictors])
+
+    assert ds.train.path == ds_def.train
+    assert ds.test.path == ds_def.test
+    assert ds.train.data_path('csv') == ds.train.path
+    try:
+        ds.train.data_path('arff')       # file loader doesn't support file auto-conversion yet
+        pytest.fail("should have raised")
+    except ValueError as e:
+        assert "Dataset iris_train is only available in csv format" in str(e)
```
PGijsbers, line 172: [`pytest.raises`](https://docs.pytest.org/en/stable/reference.html?highlight=raises#pytest-raises)

tests/unit/amlb/datasets/file/test_file_dataloader.py
```diff
@@ -0,0 +1,345 @@
...
+    assert not any([p.has_missing_values for p in ds.predictors])
+
+    assert ds.train.path == ds_def.train
+    assert ds.test.path == ds_def.test
+    assert ds.train.data_path('arff') == ds.train.path
+    try:
+        ds.train.data_path('csv')       # file loader doesn't support file auto-conversion yet
+        pytest.fail("should have raised")
+    except ValueError as e:
+        assert "Dataset iris_train is only available in arff format" in str(e)
```
PGijsbers, line 224: [`pytest.raises`](https://docs.pytest.org/en/stable/reference.html?highlight=raises#pytest-raises)

tests/unit/amlb/datasets/file/test_file_dataloader.py
```diff
@@ -0,0 +1,345 @@
...
+    assert len([p for p in ds.predictors if p.has_missing_values]) == 2
+
+    assert ds.train.path == ds_def.train
+    assert ds.test.path == ds_def.test
+    assert ds.train.data_path('csv') == ds.train.path
+    try:
+        ds.train.data_path('arff')       # file loader doesn't support file auto-conversion yet
+        pytest.fail("should have raised")
+    except ValueError as e:
+        assert "Dataset cholesterol_train is only available in csv format" in str(e)
```
PGijsbers, line 275: [`pytest.raises`](https://docs.pytest.org/en/stable/reference.html?highlight=raises#pytest-raises)

tests/unit/amlb/datasets/file/test_file_dataloader.py
```diff
@@ -0,0 +1,345 @@
...
+    assert len([p for p in ds.predictors if p.has_missing_values]) == 2
+
+    assert ds.train.path == ds_def.train
+    assert ds.test.path == ds_def.test
+    assert ds.train.data_path('arff') == ds.train.path
+    try:
+        ds.train.data_path('csv')       # file loader doesn't support file auto-conversion yet
+        pytest.fail("should have raised")
+    except ValueError as e:
+        assert "Dataset cholesterol_train is only available in arff format" in str(e)
```
PGijsbers, line 326: [`pytest.raises`](https://docs.pytest.org/en/stable/reference.html?highlight=raises#pytest-raises)

tests/unit/amlb/datasets/file/test_file_dataloader.py
```diff
@@ -0,0 +1,345 @@
...
+
+@pytest.mark.use_disk
+def test_load_binary_task_csv(file_loader):
+    ds_def = ns(
+        train=os.path.join(res, "kc2_train.csv"),
+        test=os.path.join(res, "kc2_test.csv"),
+        target="problems"
+    )
+    ds = file_loader.load(ds_def)
+    assert ds.type is DatasetType.binary
```
PGijsbers, line 41: These tests have a lot of duplicate code. I think it makes sense to create functions to re-use, e.g. `kc2_matches_expectations(ds)` that test the dataset on `kc2` characteristics (and similar for other datasets)?

PGijsbers (changes requested):
It's a lot to digest. You already indicated there was more going on in this PR but this is still more than I expected 😅 Anyway I figured I would post the remarks I have so far before my lunch break. I'll continue reviewing later.

tests/unit/amlb/datasets/file/test_file_dataloader.py
sebhrusen: good point, I never remember this one.

tests/unit/amlb/datasets/file/test_file_dataloader.py
sebhrusen: I know, there is some duplicated code but there are also some small differences between each test: I will try to improve this.

amlb/datautils.py
```diff
@@ -284,23 +258,97 @@ def inverse_transform(self, vec, **params):
...
+    print(X_fit.dtypes)
+    if isinstance(strategy, dict):
+        for dt, s in strategy.items():
+            X_fit.select_dtypes(include=dt)
+    else:
+        imputed = _impute_pd(X_fit, *X_s, missing_values=missing_values, strategy=strategy)
+    print("*** imputed ***")
+    print(imputed[0])
+    print(_rows_with_nas(imputed[0]))
+    print(imputed[0].dtypes)
```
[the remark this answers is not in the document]
sebhrusen: oh yes, I have to review those… it's even possible that this function is not used at all: I had issues with encoding when I was trying to pass pandas DF to some frameworks… and I ended up keeping the numpy array for those, so that pandas encoding is not supported at all, so I should just remove this code.

amlb/datautils.py
```diff
@@ -284,23 +258,97 @@ def inverse_transform(self, vec, **params):
...
+        print(X_ori.dtypes)
+        print("*** imputed ***")
+        print(X_np)
+        df = pd.DataFrame(X_np, columns=X_ori.columns, index=X_ori.index).convert_dtypes()
+        print("*** as df ***")
+        print(df)
+        df.astype(X_ori.dtypes.to_dict(), copy=False, errors='raise')
+        print("*** with dtypes ***")
+        print(df)
+        print(df.dtypes)
```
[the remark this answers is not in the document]
sebhrusen: same as above

frameworks/autosklearn/exec.py
```diff
@@ -15,12 +15,14 @@
 import autosklearn.metrics as metrics
 from packaging import version
 
-from frameworks.shared.callee import call_run, result, output_subdir, utils
+from frameworks.shared.callee import call_run, result, output_subdir
+from frameworks.shared.utils import Timer, system_memory_mb, walk_apply, zip_path
```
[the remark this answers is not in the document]
sebhrusen: I know, but I prefer this practice of listing explicitly what is being used

tests/unit/amlb/datasets/file/test_file_dataloader.py
sebhrusen: by the way the file loading support still has several quirks: and the main benefit of those tests is to list many of them.

frameworks/RandomForest/__init__.py
```diff
@@ -8,18 +8,20 @@ def setup(*args, **kwargs):
 
 
 def run(dataset: Dataset, config: TaskConfig):
-    from amlb.datautils import impute
+    from amlb.datautils import impute_array
     from frameworks.shared.caller import run_in_venv
 
-    X_train_enc, X_test_enc = impute(dataset.train.X_enc, dataset.test.X_enc)
+    encode = config.framework_params.get('_encode', True)
+    X_train, X_test = impute_array(dataset.train.X_enc, dataset.test.X_enc) if encode else (dataset.train.X, dataset.test.X)
```
PGijsbers, line 15: I notice that we sometimes impute in `exec.py` (decision tree) and sometimes in `__init__.py` (here). Is there a reason for this distinction? If not we should we consistent.

PGijsbers (changes requested):
I haven't had the time to run validation checks yet (though perhaps if we merge the workflow PR first, that'll be useful?). I think I have a reasonable idea of what's going on, but mostly I had to review file-by-file and might have missed some issues that arise from usage across file boundaries.

amlb/utils/process.py
```diff
@@ -447,9 +492,9 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
 class Monitoring:
 
-    def __init__(self, name=None, frequency_seconds=300, check_on_exit=False, thread_prefix="monitoring_"):
+    def __init__(self, name=None, frequency_seconds=60, check_on_exit=False, thread_prefix="monitoring_"):
```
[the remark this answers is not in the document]
sebhrusen: correct—I could also change the unit to hertz to anny everyone :) — will rename in a separate PR.

frameworks/shared/caller.py
```diff
@@ -40,52 +44,87 @@ def run_cmd_in_venv(caller_file, cmd, *args, **kwargs):
...
-        dataset.release()
+        ds = _make_input_dataset(input_data, dataset, tmpdir)
 
         config.result_dir = tmpdir
         config.result_file = mktemp(dir=tmpdir)
 
         params = json_dumps(dict(dataset=ds, config=config), style='compact')
+        log.debug("Params passed to subprocess:\n%s", params)
+        cmon = rconfig().monitoring
+        monitor = (dict(frequency_seconds=cmon.frequency_seconds,
```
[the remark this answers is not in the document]
sebhrusen: correct, I should probably disable this by default though.
This was useful for me to verify that data was not residing only in one process memory at a time.

amlb/utils/cache.py
```diff
@@ -49,12 +52,11 @@ def decorator(self):
 def memoize(fn):
```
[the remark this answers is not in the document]
sebhrusen: originally `memoize` was only supported for functions with one single arg, hence the distinction.
Now, `memoize` should also work with functions without args: but semantically, it's more common to use `memoize` when there are arguments anyway.
Also `cached` is simpler: could be removed probably, but not a priority.

frameworks/RandomForest/__init__.py
sebhrusen: hmm, do we? Is it not done in `exec.py`only when this one is not executed in a subprocess (e.g. constantpredictor).
I agree that we should be consistent, as soon as it does not make things more complex.

requirements.txt
```diff
@@ -1,8 +1,10 @@
...
-pandas>=0.23,<2.0
+numpy>=1.20,<2.0
+pandas>=1.2.4,<2.0
 psutil>=5.4,<6.0
 ruamel.yaml>=0.15,<1.0
 openml==0.11.0
-scikit-learn>=0.22.2
+scikit-learn>=0.24
+pyarrow>=4.0
+tables>=3.6
```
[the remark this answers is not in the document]
sebhrusen: I think you're right, will do this in a separate PR before versioning.

frameworks/DecisionTree/exec.py
```diff
@@ -17,8 +17,8 @@ def run(dataset: Dataset, config: TaskConfig):
 
     is_classification = config.type == 'classification'
 
-    X_train, X_test = impute(dataset.train.X, dataset.test.X)
-    y_train, y_test = dataset.train.y, dataset.test.y
+    X_train, X_test = impute_array(dataset.train.X_enc, dataset.test.X_enc)
```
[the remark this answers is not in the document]
sebhrusen: yes, this has always been the case, nothing new in this PR.
In this PR, I started to write some code with the idea to apply a different imputation strategy depending on the feature type, but it was completely out of scope (if I let myself fixing all those issues here, it will never end).
This is worth a separate  PR though.

amlb/datautils.py
```diff
@@ -284,23 +258,97 @@ def inverse_transform(self, vec, **params):
...
     :param missing_values:
-    :param strategy: 'mean', 'median', 'most_frequent', 'constant' (from sklearn 0.20 only, requires fill_value arg)
+    :param strategy: 'mean', 'median', 'mode', ('constant', value), None
     :return:
     """
-    # TODO: impute only if np.isnan(X_fit).any() ?
-    imputer = Imputer(missing_values=missing_values, strategy=strategy)
-    imputed = imputer.fit_transform(X_fit)
+    if strategy is None:
+        return [X_fit, *X_s]
```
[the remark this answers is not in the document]
sebhrusen: I prefer to keep it as it will be useful soon: it comes from my short attempt at applying a different strategy for differenty column types.

docs/HOWTO.md
sebhrusen: Yes, it's only useful for users applying custom constraints in AWs mode, but this was forgotten.

frameworks/RandomForest/__init__.py
PGijsbers: It looks like you're right. Why is it that we had `DecisionTree` run in the main process, but `RandomForest` in a subprocess?

sebhrusen: I think I used RandomForest as a test+reference implementation framework when implementing the venv+subprocess with caller/callee logic.

sebhrusen: RF runs fast, that's one of the reason.
I dont know if it's something you could take into consideration for the workflow, when there are changes in `amlb`: run `constantpredictor` + `RandomForest` to cover both use-cases (with/without subprocess).

frameworks/GAMA/__init__.py
```diff
@@ -9,11 +9,21 @@ def setup(*args, **kwargs):
 
 def run(dataset: Dataset, config: TaskConfig):
     from frameworks.shared.caller import run_in_venv
+    from frameworks.shared.serialization import ser_config
+    ser_config.pandas_serializer = 'parquet'  # can't use pickle due to pandas version mismatch
```
sebhrusen, line 13: @Innixma just FYI, with this PR, we allow serialization of the pandas dataframes to parquet format to pass data to the subprocess.
I can easily add an option to prevent the deserialization so that `exec.py` can directly consume the parquet files if you're interested.

sebhrusen: This would be a small improvement before we can directly use the openml parquet file.

sebhrusen: Or, better: the same way I provide autoconversion to `csv` files which can then be accessed as `dataset.train.data('csv')`, I can do the same with parquet.

sebhrusen: @Innixma I just tried it, and it seems to work fine for AG.
I will include the change (AG consuming parquet files instead of pandas frame) if you're interested and mention you to look at the changes.

Innixma: This sounds reasonable, the idea would be to pass the parquet string path as `train_data` param to `predictor.fit`. This way the data is loaded internally and there is no duplication of the data in memory. I'm assuming the parquet data loaded via pandas will be identical to the pandas DataFrame currently passed into exec via `dataset.train.data`. If so, this should be ideal.

For test data, it can be loaded from parquet inside of exec.py after `predictor` has been fit, that way test data won't be in memory during `fit`, which should be optimal. While we could pass the parquet string path of test to `predictor.predict`, probably better to just load it prior so we can do things such as drop the label column prior to passing to predict/predict_proba as currently done.

frameworks/AutoGluon/__init__.py
```diff
@@ -11,15 +11,12 @@ def run(dataset: Dataset, config: TaskConfig):
     from frameworks.shared.caller import run_in_venv
 
     data = dict(
-        train=dict(data=dataset.train.data),
-        test=dict(data=dataset.test.data),
+        train=dict(path=dataset.train.data_path('parquet')),
+        test=dict(path=dataset.test.data_path('parquet')),
```
sebhrusen, line 15: @Innixma here is the switch to parquet files.
The conversion is imply using `pandas.to_parquet` on the original dataframe, and the later doesn't stay in memory.

frameworks/AutoGluon/exec.py
```diff
@@ -46,22 +46,13 @@ def run(dataset, config):
...
-    gc.collect()
+    train, test = dataset.train.path, dataset.test.path
+    label = dataset.target.name
+    problem_type = dataset.problem_type
```
sebhrusen, line 51: @Innixma switch to parquet part2

frameworks/AutoGluon/__init__.py
Innixma: looks good to me!

frameworks/AutoGluon/exec.py
Innixma: Yes it is ignored, so it should be fine

requirements.txt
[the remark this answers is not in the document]
sebhrusen: https://github.com/openml/automlbenchmark/issues/306
Will look into this soon.

amlb/data.py
```diff
@@ -86,54 +83,55 @@ class Datasplit(ABC):
...
         """
-        return self.data[:, self.dataset.target.index]
+        return self.data.iloc[:, [self.dataset.target.index]]
 
     @lazy_property
     @profile(logger=log)
     def data_enc(self) -> np.ndarray:
-        encoded_cols = [f.label_encoder.transform(self.data[:, f.index]) for f in self.dataset.features]
+        data = np.where(self.data.notna(), self.data.values, None)
+        print("*****")
```
sebhrusen, line 133:
```suggestion
```

amlb/datasets/file.py
```diff
@@ -247,58 +269,80 @@ def load_metadata(self):
...
 
         for f in features:
-            col = df.iloc[:, f.index]
+            col = self._ds.iloc[:, f.index]
             f.has_missing_values = col.hasnans
             if f.is_categorical():
-                unique_values = col.dropna().unique() if f.has_missing_values else col.unique()
-                f.values = [str(v) for v in sorted(unique_values)]
+                # unique_values = col.dropna().unique() if f.has_missing_values else col.unique()
+                # f.values = [str(v) for v in sorted(unique_values)]
```
PGijsbers, line 341:
```suggestion
```

amlb/datasets/openml.py
```diff
@@ -139,133 +131,194 @@ def get_values(f):
...
+                            else self._get_categorical_values(c) if pat.is_categorical_dtype(dt)
+                            else 'STRING'
+                           ))
+                          for c, dt in zip(df.columns, df.dtypes)]
+            arff.dump(dict(
+                description=description,
+                relation=name,
+                attributes=attributes,
+                data=df.values
+                # data=(df * 1).values  # converts bool to 0/1 and NAs to NaN
```
PGijsbers, line 271:
```suggestion
```

PGijsbers (approved):

PR #240, opened 2021-01-15
Update gama

PGijsbers:
Update GAMA so it works with `>=20.2.2` as well as the old supported version(s).
Tested in 'local' mode from a docker container on the `validation` set.

PR #238, opened 2021-01-05
ML-Plan: Added prediction runtime for test predictions.

mwever:
* Added feature for returning the time taken for making the predictions with the finally chosen candidate.

Note: ML-Plan takes the time for fitting the finally chosen solution candidate into account and tries to meet the timeout including fitting the solution. The prediction time for the test data, however, is not anticipated.

Furthermore, we just released a new version (0.2.4) with several bug fixes and a few enhancements. I did not want to mess up the frameworks.yaml files so I did not touch them. As far as I can see, except for the Q2.2020 file, the version of ML-Plan always refers to the "latest" version. If this is also the version which is benchmarked everything should be fine.

PR #294, opened 2021-04-22
Github workflow for automatic framework evaluation

PGijsbers:
Supersedes #244 which was based on a feature branch.

This introduces a workflow which is to be executed on each PR with the following behavior:
 - detect whether the changes are to shared framework code, only to specific frameworks, or neither (e.g. doc changes)
 - if shared code is changed, execute a quick-to-fail baseline check with `constantpredictor` for fast feedback
 - evaluate only those frameworks with changed code, or all frameworks in the case of a shared code change (if the baseline check passed).
 
 or, if this makes it clearer:
image

I left some remarks with `Files changed` tab about possible improvements and issues.
The main thing that I assume needs work before approval is that I had to hack in a way to have `runbenchmark.py` fail with non-zero error code, which allows the jobs in the workflow to fail.

ML-Plan, GAMA and Tuned Random Forest currently fail because #238, #240 and #295 still need to be merged.

**note:** The commit history is incredibly dirty due to having to debug through CI. A squash-merge should be fine, right?


runbenchmark.py
```diff
@@ -140,6 +140,11 @@
     bench.setup(amlb.SetupMode[args.setup])
     if args.setup != 'only':
         res = bench.run(args.task, args.fold)
+        if res is None and args.exit_on_error:
+            #  The desired behavior was to actually have a non-zero exit code.
+            #  I thought this worked before but there are currently so many layers
+            #  of exception handling, I'm not sure how to propagate it back up.
+            raise Exception("No results, likely due to error during evaluation.")
```
PGijsbers, line 147: So the original reason I added the `-e` flag was to allow the benchmark script to exit with a non-zero code, which would allow the failure to be picked up by continuous integration. I am not sure if I hadn't done it correctly at the time, or if the changes since abandoned this behavior.
I realize this is probably not the best way to bring it back, but there are several layers of exception handling so I wasn't sure where or how to go about it. I stuck with this for now until feedback.

.github/workflows/run_all_frameworks.yml
```diff
@@ -0,0 +1,128 @@
...
+                is_common=0
+                break
+            fi
+          done
+
+          # Indicates which jobs should be executed (0) or not (1)
+          skip_evaluation=0
+          if [[ $is_common -eq 0 ]];
+          then
+            FRAMEWORKS='["autogluon", "autosklearn", "gama", "h2oautoml", "mlplanweka", "tpot", "constantpredictor", "randomforest", "tunedrandomforest"]'
```
PGijsbers, line 50: I am not aware if we have any explicit list of frameworks we intend to support throughout changes made to the shared codebase. It would probably be good to have this and load it here instead of using hard coded values.

.github/workflows/run_all_frameworks.yml
```diff
@@ -0,0 +1,128 @@
...
+      - id: find-required-tests
+        name: Detect Common Changes  # detect if any changes occurred that should trigger all frameworks.
+        run: |
+          changed_files=$(git diff --name-only HEAD..$GITHUB_BASE_REF)
+          common_files="amlb/** "\
+          "frameworks/shared/** "\
+          ".github/workflows/run_all_frameworks.yml "\
+          ".github/runbenchmark/action.yml "\
+          "runbenchmark.py "\
+          "requirements.txt"
```
PGijsbers, line 31: Not exactly sure what the best way to do this is. The idea is that we need to test all frameworks if shared code has changed, so we need a way to define shared code. I was also considering an alternative where everything is shared code unless it's specifically not, but that's not very clean either due to the `frameworks` folder which contains both shared and specific code. Note that we don't want to trigger this CI on e.g. doc changes.

.github/workflows/run_all_frameworks.yml
```diff
@@ -0,0 +1,128 @@
...
+    runs-on: ubuntu-latest
+    needs:
+     - baseline
+     - detect_changes
+    if: ${{ success() }} || ${{ cancelled() }}
+    strategy:
+      matrix:
+        python-version: [3.8]
+        framework: ${{ fromJson(needs.detect_changes.outputs.frameworks) }}
+        task: [APSFailure, bioresponse, dresses-sales, eucalyptus, internet-advertisements, kc1, micro-mass]
```
PGijsbers, line 110: This should probably be changed to fetch from `validation.yaml`, but developing so far has been a bit of a PITA so I hope it's okay to keep it hard-coded for now.

runbenchmark.py
sebhrusen: I'm fixing this separately now: I call return in `finally` blocks in a few places…
I'm confused sometimes with this `finally` behaviour in Python compared with some other languages.

sebhrusen: see https://github.com/openml/automlbenchmark/pull/297/files
I probably changed the indent when improving/rewriting the job runner a few weeks ago.

.github/workflows/run_all_frameworks.yml
sebhrusen: yes, we also need to talk about framework's lifecycle soon, e.g. what should we do with "dead" frameworks like `oboe`, `hypertoptsklearn`? If `autoxgboost` still alive?
Also, now that the frameworks point to the most recent release by default, it can break our workflow on each new release on their side, which can be considered both a good or a bad thing.

.github/workflows/run_all_frameworks.yml
sebhrusen: I'm fine with hardcoded, it's not as if we're going to change `validation.yaml` often.

Actually wondering if we always need all of them: I would run them all for framework-specific changes, and a subset (2 or 3 max) for common changes given that we have a much larger test matrix with all frameworks being included: I would even stick to `test` datasets in this case (shorter time constraints + one dataset of each kind).
wdyt?

sebhrusen (commented):
I'm ready to approve this for merge, just adding some comments for improvements that could be done soon.

runbenchmark.py
PGijsbers: Will do, thanks! It's hard to tell what is/is not intention at times :)

.github/workflows/run_all_frameworks.yml
PGijsbers: Good point. I propose that we actually run CI with the latest fixed-version definition. We can (and should) run CI on `stable` upgrades as well, but I don't think it should interfere with our general PR workflow.

.github/workflows/run_all_frameworks.yml
PGijsbers: I think that's a good idea (test for shared, validation for specific). Shared changes which affect a validation dataset but none of the test sets are probably unlikely enough to justify the additional compute time. Will update.

.github/workflows/run_all_frameworks.yml
```diff
@@ -0,0 +1,125 @@
...
+          git checkout -b $GITHUB_BASE_REF origin/$GITHUB_BASE_REF
+          git checkout $GITHUB_HEAD_REF
+      - id: find-required-tests
+        name: Detect Common Changes  # detect if any changes occurred that should trigger all frameworks.
+        run: |
+          changed_files=$(git diff --name-only HEAD..$GITHUB_BASE_REF)
+          common_files="amlb/** "\
+          "frameworks/shared/** "\
+          ".github/workflows/run_all_frameworks.yml "\
+          ".github/runbenchmark/action.yml "\
```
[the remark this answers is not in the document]
PGijsbers: I specifically only included those things that affect this Workflow. I don't think it makes sense to run this Workflow to test changes to a different workflow?

PGijsbers:
Changes since review:
 - Baseline check expanded to use all test tasks (and removed `constantpredictor` from "all frameworks" since it's covered in baseline)
 - Perform checks on `test` for shared code changes, `validation` for framework specific changes.
 - Fixed cache usage (it's not exactly clean, but at least it works now, have to make it flexible with Python versions later).
 - Fixed bug that caused an error when more than one framework was changed
 
 It looks like `TunedRandomForest` fails due to time constraints. I suppose I can investigate that after the weekend.

PGijsbers:
The `TimeoutError` that restricted `TunedRandomForest`'s parameter search was swallowed which led to timeouts. I put the fix in this PR (since it's very minor anyway).

PGijsbers:
Based on #293 I added `RandomForest` as a baseline because it is the simplest example with a framework running in a subprocess with its own virtual environment.

.github/workflows/run_all_frameworks.yml
```diff
@@ -0,0 +1,156 @@
...
+        run: |
+          git fetch --all
+          git checkout -b $GITHUB_BASE_REF origin/$GITHUB_BASE_REF
+          git checkout $GITHUB_HEAD_REF
+      - id: find-required-tests
+        name: Detect Common Changes  # detect if any changes occurred that should trigger all frameworks.
+        run: |
+          changed_files=$(git diff --name-only HEAD..$GITHUB_BASE_REF)
+          common_files="amlb/** "\
+          "frameworks/shared/** "\
```
sebhrusen, line 29: what about including `config.yaml`? A change/renaming in this file can have an impact on the framework behaviour.
Or even include the whole `resources` folder to remain on the safe side.

sebhrusen (approved):
see my small suggestion below, otherwise, feel free to merge.

PR #295, opened 2021-04-22
Install requirements.txt for Tuned Random Forest

PGijsbers:
`Stopit` module is used to ensure fitting normally stops in reasonable time for
Tuned Random Forest to return a model. For whatever reason it seemed to
no longer be installed.


sebhrusen (approved):
I probably dropped it by mistake at some point, thanks for noticing this.

PR #300, opened 2021-05-05
Update Auto-sklearn integration to use float32

mfeurer:
Auto-sklearn is trained on data from OpenML-Python which by default is float32. Changing the integration to also pass float32 data to Auto-sklearn.

### DIFF
--- a/frameworks/autosklearn/exec.py
+++ b/frameworks/autosklearn/exec.py
@@ -13,6 +13,7 @@ import autosklearn
 from autosklearn.estimators import AutoSklearnClassifier, AutoSklearnRegressor
 from autosklearn.experimental.askl2 import AutoSklearn2Classifier
 import autosklearn.metrics as metrics
+import numpy as np
 from packaging import version
 
 from frameworks.shared.callee import call_run, result, output_subdir, utils
@@ -59,8 +60,10 @@ def run(dataset, config):
         perf_metric,
     )
     log.info("Environment: %s", os.environ)
-
-    X_train = dataset.train.X_enc
+    
+    # Cast x to float32 to save memory, keep y at the default (most likely float64) to have
+    # higher precision for regression tasks
+    X_train = dataset.train.X_enc.astype(np.float32)
     y_train = dataset.train.y_enc
     predictors_type = dataset.predictors_type
     log.debug("predictors_type=%s", predictors_type)


PGijsbers:
@sebhrusen @ledell Should we accept this? I feel this is a slippery slope. I understand the need for a smaller memory footprint (some datasets are large-ish compared to typical available memory), but we aim to avoid custom preprocessing as much as possible. Perhaps we should allow lower precision datasets directly from the framework, or expect the AutoML frameworks to deal with this internally (and maybe allow this for now)?

PGijsbers:
After thinking a bit more about this, we are already partially on that slope with custom data loading in e.g. H2O's [import file](https://github.com/openml/automlbenchmark/blob/9b3ef45f52ca92c318909a9a2399bb8ae8865cb3/frameworks/H2OAutoML/exec.py#L88). It's probably best to have some explicit guidelines for what is or is not desirable. Should the line be drawn similar to initialization, i.e. a file loader is accepted so long as it does not need any additional hyperparameter configuration (other than trivial ones like naming the dataframe)?

But I wouldn't want to 'force' data loaders. I think a more elegant solution is allow the data to be loaded/processed within the `fit` function, so you don't _need_ to do an extra call. But that's tricky because then the frameworks have to self-report how much time they spend in search vs loading the data. I suppose we can do some measurements to see how much time it takes to e.g. convert the datasets to see if it has a significant impact on the search time. I assume right now it doesn't matter much yet, but it might be an issue in the future.

sebhrusen:
Btw, it looks like with the switch to openml data loading (#293), the numpy array dtype (when using X_enc) is always `float32`, so, as autosklearn is consuming the numpy array instead of the pandas dataframe, it should make everyone happy.

PGijsbers:
Auto-sklearn should [support pandas](https://automl.github.io/auto-sklearn/development/examples/40_advanced/example_pandas_train_test.html) nowadays. Though there are some stability issues ([auto-sklearn#1041](https://github.com/automl/auto-sklearn/issues/1141)), so I suppose the easiest way here is for them to keep using `numpy` for now (and enjoy those 4 byte floats).

> I can remove the logic for H2OAutoML data loading if you're concerned

Given the explanation and that it's not permanent I don't think it's a point of concern.
When OpenML has moved over to `parquet` we should probably consider offering that in our data feed as well. This will make the difference smaller between loading data within the `train` context as opposed to before it (given that loading data from parquet will be much faster).

Innixma:
There is one subtlety here in terms of memory efficiency: If the framework is given the file path instead of the loaded data, it can load the data in a way such that the original data doesn't have to persist in memory beyond its initial loading. This isn't possible if loaded from an outside context, as the original DataFrame will still exist in memory even if it is deleted inside the framework's code. For example, AutoGluon tries not to edit the outer context data, so it performs a copy of the data internally which doubles memory usage. If it is loaded from file however, AutoGluon makes no copy because there is no outer context to alter.

I suppose it gets tricky when the claim is that pandas doesn't load files correctly, but hopefully that would not be the case for parquet files.

sebhrusen:
@Innixma why doesn't your integration consume the arff file directly then?

Innixma:
@sebhrusen AutoGluon doesn't support arff files. I'm unsure how many users work with arff files compared to CSV/parquet. If it's very easy to add, I might consider implementing it, but it sounds like the plan for automlbenchmark is to move to parquet files.

mfeurer:
> At the same time as we're trying to standardize the way we feed frameworks with data, we'll see requests for optimizations…

I guess this is due to a lack of knowledge how data is/can be loaded within the benchmarking framework. Is there any documentation on the supported data formats and how the framework is expected to load the data?

As there is no canonical data format for ML data, and arff is a horrible format, Auto-sklearn does not provide a data loader. To improve things we're working on an improved pandas integration and it should be available in the next iteration.

> I'm unsure how many users work with arff files compared to CSV/parquet.

Hopefully none. AFAIK OpenML uses arff files as the developers were WEKA users back then.

> or expect the AutoML frameworks to deal with this internally

I'll implement this, it is much more helpful to have this within Auto-sklearn. We can then keep the data 64bit if we have sufficient memory and sample it to lower precision if we don't.

> Btw, it looks like with the switch to openml data loading (#293), the numpy array dtype (when using X_enc) is always float32, so, as autosklearn is consuming the numpy array instead of the pandas dataframe, it should make everyone happy.

I'm now actually wondering whether we want to update OpenML-Python to return float64 to not loose too much precision here (or allow the user to specify the precision).

ledell:
> Hopefully none. AFAIK OpenML uses arff files as the developers were WEKA users back then.

I believe they use ARFF over CSV simply because it allows for data types to be specified.  In that respect, ARFF > CSV (for ML).  However, I agree that Parquet seems more appropriate nowdays.  Once they [add support](https://github.com/openml/openml-python/issues/1032) for Parquet, maybe we can use that in the benchmark instead?

PR #302, opened 2021-05-13
Ignore binary *sif files when building a docker image

franchuterivera:
**Problem Statement**

Undesired huge files get added to a docker image when there are previous singularity images available.

**Proposed solution**
To add these files to the docker ignore file


PGijsbers (approved):
I'm a bit confused about expecting two locations with `.sif` files? I'm not really familiar enough with the Singularity workflow. I don't really see a reason we would ever want a `.sif` file in the docker image so adding it is fine with me.

sebhrusen:
The question is: where do they come from so that they have to be ignored?

.dockerignore
```diff
@@ -37,3 +37,5 @@ my_*
 **/*.gz
 **/*.bz2
 **/*.png
+**/*.sif
+*.sif
```
sebhrusen, line 41:
```suggestion
**/*.sif
```

sebhrusen (approved):
Removing the second exclusion and merging, thanks.

.dockerignore
sebhrusen: @franchuterivera, for some reason, I'm not able to commit the suggestion ("no changes were made to the code" !!!).
Please try on your side, I'll merge afterwards.

franchuterivera:
that is strange @sebhrusen. Thanks for the commit suggestion. I made the change in my fork, and it seems to be reflected here.

PR #303, opened 2021-05-14
Convert single task study calls to a task call

PGijsbers:
This avoids performing a call to retrieve the study and all their dataset metadata (2 calls total) for each job.

### DIFF
--- a/amlb/runners/aws.py
+++ b/amlb/runners/aws.py
@@ -333,14 +333,23 @@ class AWSBenchmark(Benchmark):
         def _run(_self):
             try:
                 resources_root = "/custom" if rconfig().aws.use_docker else "/s3bucket/user"
+                benchmark = (self._forward_params['benchmark_name']if self.benchmark_path is None or self.benchmark_path.startswith(rconfig().root_dir)
+                                   else "{}/{}".format(resources_root, self._rel_path(self.benchmark_path)))
+
+                if benchmark.startswith('openml/s/') and len(task_names) == 1:
+                    task_id = next(task['openml_task_id'] for task in self.benchmark_def if task['name'] == task_names[0])
+                    benchmark = f'openml/t/{task_id}'
+                    _task_names = []
+                else:
+                    _task_names = task_names
+
                 _self.ext.instance_id = self._start_instance(
                     instance_def,
                     script_params="{framework} {benchmark} {constraint} {task_param} {folds_param} -Xseed={seed}".format(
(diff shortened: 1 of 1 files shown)


Innixma:
Is this a fix that will reduce the frequency of openml server errors? Does this impact runs done with pre-existing benchmark files instead of studies (such as `resources/benchmarks/medium.yaml`)?

PGijsbers:
It reduces requests only when the benchmark was specified with the `openml/s/N` format. There should be an update to the retry policy for `openml-python` later this week which hopefully alleviates the server issues more generally.

amlb/runners/aws.py
```diff
@@ -333,14 +333,23 @@ def _setup(_self):
...
                 resources_root = "/custom" if rconfig().aws.use_docker else "/s3bucket/user"
+                benchmark = (self._forward_params['benchmark_name']if self.benchmark_path is None or self.benchmark_path.startswith(rconfig().root_dir)
+                                   else "{}/{}".format(resources_root, self._rel_path(self.benchmark_path)))
+
+                if benchmark.startswith('openml/s/') and len(task_names) == 1:
+                    task_id = next(task['openml_task_id'] for task in self.benchmark_def if task['name'] == task_names[0])
+                    benchmark = f'openml/t/{task_id}'
+                    _task_names = []
+                else:
+                    _task_names = task_names
```
sebhrusen, line 344: You don't need to change task_names variable, it will work either way.

Did you try it?
I want to be sure that the folder structure generated on s3 is the same as before and that this change is not making it more difficult to retrieve results from s3 a posteriori.
Currently, s3 is the long-term storage for results and those are organized by sessions, and inside the session, each folder contains the original benchmark name and the task name, which makes it relatively easy to download only a specific result.
I think it should be fine though as `aws` mode is running benchmarks using `--session=` (which removes the session folder on the ec2 instance to avoid an additional subfolder) and this should prevent the modifed params to appear anywhere.

PGijsbers: Testing with `python runbenchmark.py constantpredictor openml/s/264 -m aws -f 0` the structure on the bucket seems the same, but the local result directory is actually different. Both have the same `aws.openml_s_264.test.all_tasks.0.constantpredictor` subdirectory with the data from that run, but the main directory of this branch does not feature `logs` and `logs.zip`.

You seem to be correct that task names don't need to be modified, though I find the `openml/t/61 -t iris` notation a bit odd to explicitly support.

Issue #306, opened 2021-05-17
AMLB app should use frozen requirements.


sebhrusen:
Rename `requirements.txt` to `requirements.in` listing only the dependencies directly used by amlb, with non-fixed versions for easier upgrades.
From a fresh build, generate `requirements.txt` using `pip freeze` or `pip-compile` (TBD).
Rebuild `requirements.txt` for each new update in `requirements.in`.

Same logic must apply to `frameworks/shared/requirements.in·txt

PR #307, opened 2021-05-18
Freeze amlb requirements

sebhrusen:
https://github.com/openml/automlbenchmark/issues/306

Using `pip-tools` to generate `requirements.txt` and `frameworks/shared/requirements.txt`

Tested by using both `docker` and `aws` mode with `randomforest` framework.


PGijsbers (approved):
Looks good! 

I was thinking if there's a danger the `requirements.txt` and `requirements.in` files may get out-of-sync. I think we should consider adding a workflow that checks that when a change is made to `requirements.in` the produced `requirements.txt` matches the `requirements.txt` in the PR. Do you think that would be a good idea? I'd be ok with just adding that as a todo on the issue tracker.

sebhrusen:
@PGijsbers Yes, it would be great to have a workflow that updates the 2 `requirements.txt` when the `requirements.in` are edited.
I'll just add `pip-tools` to requirements-dev.txt in this PR, so that everything is ready to use.

sebhrusen:
@PGijsbers FYI: https://github.com/openml/automlbenchmark/issues/308

PR #309, opened 2021-05-19
Update Auto-Sklearn Setup file

franchuterivera:
This PR enhances Auto-Sklearn setup with 2 changes:

1- In auto-sklearn's workflow, the master branch refers to the stable release. We want to make this clear via the setup.sh
2- We want to be able to provide both hash/branch names as a version, as this enables us to test (within the automlbenchmark umbrella) different auto-sklearn versions.


sebhrusen (changes requested):
@franchuterivera your changes are completely removing the current support for branches and changing the semantics.

currently when specifying a framework version in a framework definition file the expectations are as follow for all frameworks:
- version: 0.12.6    will install this release if available.
- version: stable    will install the most recent stable version (from pip) if available: for frameworks that don't provide any release, we accept 'stable' pointing to a branch.
- version:  latest    will install the latest build (not officially release), can also point to a master or dev branch.
- version: my_branch   will install from the givne branch name.

With your changes, there's no possibility anymore to specify a branch.
If you want to use master, please just create a local definition for yourself with `version: latest`.

frameworks/autosklearn/setup.sh
```diff
@@ -19,12 +19,15 @@ PIP install --no-cache-dir -r $HERE/requirements.txt
...
     PIP install --no-cache-dir -U ${PKG}==${VERSION}
 else
-#    PIP install --no-cache-dir -e git+${REPO}@${VERSION}#egg=${PKG}
     TARGET_DIR="${HERE}/lib/${PKG}"
     rm -Rf ${TARGET_DIR}
-    git clone --depth 1 --single-branch --branch ${VERSION} --recurse-submodules ${REPO} ${TARGET_DIR}
+    git clone  --recurse-submodules ${REPO} ${TARGET_DIR}
+    cd ${TARGET_DIR}
+    git checkout $VERSION
+    cd ${HERE}
```
sebhrusen, line 30: we don't need to pull the entire repo just to run `autosklearn`, why removing the `--depth`, `--single-branch` and `--branch` arguments?

sebhrusen:
@franchuterivera 
> 1- In auto-sklearn's workflow, the master branch refers to the stable release. We want to make this clear via the setup.sh

if you mean that `master` contains the same code as the latest release available from pip, then the relevant change would be:
```
if [[ "$VERSION" == "latest" ]]; then
    VERSION="dev"  # or whatever your main dev branch is
fi
```
As soon as you have a release build, it's better to install the `stable` release from pip than from `master` branch in that case`.

> 2- We want to be able to provide both hash/branch names as a version, as this enables us to test (within the automlbenchmark umbrella) different auto-sklearn versions.

Is this the reason why you removed the `--depth` ... arguments?

franchuterivera:
Hello, sorry for the delay in the reply. I have incorporated your feedback. Please let me know if you think this can be further improved, but I have checked with @mfeurer and from our side, this looks ok.

franchuterivera:
BTW, I tried changing the 'version' field in the `resources/framework.yml` yet it kept selecting stable for me.

Is there a way one can easy change the version from the command line? I was trying to search the documentation but I could not find a nice way (and ended up hardcoding the version for testing.)

PGijsbers:
I updated the PR so it saves the hash to `.installed` (if a branch is specified, the hash is first retrieved so the version is still unique). In the process I converted it to a similar workflow as the R setup mentioned. In the future we can probably use some shared functions instead. 

There is currently an issue creating docker containers (the `#` in the version creates an invalid tag). I suppose that could also be dealt with in a separate PR (since it's already present with the R frameworks, I just hadn't noticed because I had to test those on aws).

PGijsbers:
I did undo some of the optimization, I'll patch that back in.

frameworks/autosklearn/setup.sh
```diff
@@ -19,13 +19,28 @@ PIP install --no-cache-dir -r $HERE/requirements.txt
...
-#    PIP install --no-cache-dir -e git+${REPO}@${VERSION}#egg=${PKG}
     TARGET_DIR="${HERE}/lib/${PKG}"
     rm -Rf ${TARGET_DIR}
-    git clone --depth 1 --single-branch --branch ${VERSION} --recurse-submodules ${REPO} ${TARGET_DIR}
+
+    if [[ "$VERSION" =~ ^# ]]; then
+      COMMIT="${VERSION:1}"
+    else
+      # find the latest commit to the VERSION branch
+      COMMIT=$(git ls-remote "${REPO}" | grep "refs/heads/${VERSION}" | cut -f 1)
```
sebhrusen, line 32: why to we need the commit when we specify a branch?
Then it forces us to clone the entire repository when we used to need to clone only the specific branch without history... it's a bit silly to clone a full repo when you're not planning to work on it.

sebhrusen: ok, I guess this optimization was temporarily removed.

amlb/runners/container.py
```diff
@@ -38,6 +38,9 @@ def image_name(cls, framework_def, label=None, **kwargs):
         if label not in rconfig().container.ignore_labels:
             tags.append(label)
         tag = re.sub(r"([^\w.-])", '.', '-'.join(tags))
+        # Some frameworks allow specifying a version by #HASH which would lead to
+        # the tag starting with a '.' which is invalid.
+        tag = re.sub(r"^\.+", "", tag)
```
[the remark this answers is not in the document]
PGijsbers: Absolutely right 👍

PGijsbers:
Verified on AWS and Docker for both branch and specific commit.

frameworks/autosklearn/setup.sh
```diff
@@ -19,13 +19,29 @@ PIP install --no-cache-dir -r $HERE/requirements.txt
...
+    git clone  --recurse-submodules --shallow-submodules ${DEPTH} ${REPO} ${TARGET_DIR}
+    cd ${TARGET_DIR}
+    git checkout "${COMMIT}"
+    cd ${HERE}
     PIP install -U -e ${TARGET_DIR}
 fi
 
 PY -c "from autosklearn import __version__; print(__version__)" >> "${HERE}/.installed"
+if [[ -n $COMMIT ]]; then
+  truncate -s-1 "${HERE}/.installed"
```
sebhrusen, line 45: do you need this to remove the last NL?

sebhrusen (approved):
Looks good to me, thanks @franchuterivera and @PGijsbers

frameworks/autosklearn/setup.sh
PGijsbers: Exactly the purpose :)

PR #311, opened 2021-05-20
Support for execution plans

sebhrusen:
https://github.com/openml/automlbenchmark/issues/169

Execution plans will allow to run benchmarks for multiple frameworks, benchmark definitions, constraints, tasks, folds… and still benefit from full parallelism, especially in AWS mode.

An execution plan is represented  as a csv file:
```
```

### DIFF
--- a/amlb/__init__.py
+++ b/amlb/__init__.py
@@ -2,11 +2,11 @@
 amlb entrypoint package.
 """
 
+from .benchmark import SetupMode
 from .logger import app_logger as log
 from .errors import AutoMLError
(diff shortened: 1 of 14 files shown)


Innixma:
Would it make sense to add auto-retry support as an extra parameter? This way, if a failure occurs on framework A, task B, fold 3, then it is retried up-to the amount of retries specified by the user? This may be simpler than the user having to feed the failures.csv results into a follow-up run (and do it repeatedly until N retries have been performed).

PR #312, opened 2021-05-21
Upgrade openml to 0.12.2 and make use of new connectivity options

PGijsbers:
Upgrade and improvements:
 - upgrades openml-python dependency to 0.12.2
 - uses the `retry_policy` option introduced in `0.12.2`
 - use the introduced `download_qualities` parameter of `get_task` to avoid downloading `qualities.xml` altogether.

### DIFF
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -39,8 +39,8 @@ class OpenmlLoader:
         if task_id is not None:
             if dataset_id is not None:
                 log.warning("Ignoring dataset id {} as a task id {} was already provided.".format(dataset_id, task_id))
-            task = oml.tasks.get_task(task_id)
-            dataset = task.get_dataset()
+            task = oml.tasks.get_task(task_id, download_qualities=False)
+            dataset = oml.datasets.get_dataset(task.dataset_id, download_qualities=False)
             _, nfolds, _ = task.get_split_dimensions()
             if fold >= nfolds:
                 raise ValueError("OpenML task {} only accepts `fold` < {}.".format(task_id, nfolds))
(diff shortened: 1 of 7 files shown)


docs/README.md
```diff
@@ -83,7 +83,7 @@ pip3 install -r requirements.txt
 - _**NOTE**: in case of issues when installing Python requirements, you may want to try the following:_
     - _on some platforms, we need to ensure that requirements are installed sequentially:_ `xargs -L 1 pip install < requirements.txt`.
     - _enforce the `pip3` version above in your virtualenv:_ `pip3 install --upgrade pip==19.3.1`.
-
+    - _when installing on Windows using `python -m pip` is recommended, and may be required to successfully install `requirements-dev.txt`.
```
sebhrusen, line 86: You could also safely replace all occurences of `pip` ot `pip3` in the doc with `python -m pip`.
The later syntax is now recommended anyway, and this is what we use in ec2 instances and containers.

runbenchmark.py
```diff
@@ -5,6 +5,8 @@
 import shutil
 import sys
 
+import openml
+
```
sebhrusen, line 9: I don't like the idea of adding a dependency to openml in this script and making changes even before knowing if it will be used at all (it is not if loading data from files).
Can't you apply this logic directly in `amlb.datasets.openml` module?

sebhrusen (changes requested):

amlb/datasets/openml.py
```diff
@@ -34,40 +34,27 @@ def __init__(self, api_key, cache_dir=None):
         if cache_dir:
             oml.config.set_cache_directory(cache_dir)
 
+        if oml.config.retry_policy != "robot":
+            log.debug("Setting openml retry_policy from '%s' to 'robot'." % oml.config.retry_policy)
+            oml.config.set_retry_policy("robot")
```
sebhrusen, line 39: just wondering if a similar logic is not needed in `amlb/benchmarks/openml.py`

sebhrusen (approved):

amlb/datasets/openml.py
PGijsbers: I never experienced the issues myself (I don't have the permissions to spawn instances in those numbers). Though that's partially the reason I had it in `runbenchmark.py`. I'll just add it there as well to ensure robot policy is used in all stages.

PR #315, opened 2021-05-31
Testing R Workflows

PGijsbers:
This turned out into finding item after item to improve or fix. Don't review this PR, it's just open so I can debug Workflows. I'll create a new PR for the final version.

### DIFF
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -256,7 +256,15 @@ class ArffSplitter(DataSplitter[str]):
 
     def _get_categorical_values(self, col):
         feat = next((f for f in self.ds._oml_dataset.features.values() if f.name == col), None)
-        return feat.nominal_values if feat is not None else None
+        if feat is not None:
+            # openml-python converts categorical features which look boolean to
+            # boolean values, which always write values as 'True' and 'False',
+            # so we need to adapt the header accordingly.
+            # Not doing so causes an issue in the R packages.
+            if set(v.lower() for v in feat.nominal_values) == {"true", "false"}:
+                return ["True", "False"]
+            return feat.nominal_values
+        return None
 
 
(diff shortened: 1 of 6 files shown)


PGijsbers:
Status update: I asked @Coorsaa to look into/get someone to look into the failures for autoxgboost and mlr3automl on KC1.

After those issues are resolved I'll set up the new PR which cleanly adds the changes of this PR to the CI pipeline.

amlb/datasets/openml.py
```diff
@@ -256,7 +256,15 @@ def _save_split(self, df, path, name):
 
     def _get_categorical_values(self, col):
         feat = next((f for f in self.ds._oml_dataset.features.values() if f.name == col), None)
-        return feat.nominal_values if feat is not None else None
+        if feat is not None:
+            # openml-python converts categorical features which look boolean to
+            # boolean values, which always write values as 'True' and 'False',
+            # so we need to adapt the header accordingly.
+            # Not doing so causes an issue in the R packages.
```
sebhrusen, line 263: Nice catch.
This autoconversion to boolean is new, isn't it? I remember that I quickly mentioned that a couple of months ago, and I don't think it was the case by then.
Btw, `H2OAutoML` is also using the arff splits and I didn't notice any issue there recently: any specific dataset that was failing? I just want to verify the behaviour in H2O, thanks.

PGijsbers: I don't know when exactly from the top of my head, it has indeed been changed between 0.10 (previous version we used) and 0.12.2. I had to fix this because the R packages failed on KC1 because of this. The header will say `{false, true}` but the data is `{True, False}`. Other frameworks seemed to ignore this difference.

PGijsbers:
Superseded by #325 and #336 (WIP).

PR #324, opened 2021-06-14
Fix mismatch between arff header and data

PGijsbers:
Due to the conversion that openml-python does from a categorical arff column to a boolean pandas column, true/false categoricals are actually written as True/False (note capitalization) in the data but the features still describe it as true/false. This leads to a is match in the ARFF header and the actual data, which caused issues for R frameworks. 

Confirmed in #315.

### DIFF
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -256,7 +256,15 @@ class ArffSplitter(DataSplitter[str]):
 
     def _get_categorical_values(self, col):
         feat = next((f for f in self.ds._oml_dataset.features.values() if f.name == col), None)
-        return feat.nominal_values if feat is not None else None
+        if feat is not None:
+            # openml-python converts categorical features which look boolean to
+            # boolean values, which always write values as 'True' and 'False',
+            # so we need to adapt the header accordingly.
+            # Not doing so causes an issue in the R packages.
+            if set(v.lower() for v in feat.nominal_values) == {"true", "false"}:
+                return ["True", "False"]
+            return feat.nominal_values
+        return None
 
 
 class CsvSplitter(DataSplitter[str]):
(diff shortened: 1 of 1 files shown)


amlb/datasets/openml.py
```diff
@@ -256,7 +256,15 @@ def _save_split(self, df, path, name):
...
     def _get_categorical_values(self, col):
         feat = next((f for f in self.ds._oml_dataset.features.values() if f.name == col), None)
-        return feat.nominal_values if feat is not None else None
+        if feat is not None:
+            # openml-python converts categorical features which look boolean to
+            # boolean values, which always write values as 'True' and 'False',
+            # so we need to adapt the header accordingly.
+            # Not doing so causes an issue in the R packages.
+            if set(v.lower() for v in feat.nominal_values) == {"true", "false"}:
+                return ["True", "False"]
```
sebhrusen, line 265: are categories not always sorted in ARFF headers? just wondering

PGijsbers: There's nothing in the [ARFF specificiation](https://www.cs.waikato.ac.nz/ml/weka/arff.html) that specifies an order:
```
 Nominal values are defined by providing an <nominal-specification> listing the possible values: {<nominal-name1>, <nominal-name2>, <nominal-name3>, ...}

For example, the class value of the Iris dataset can be defined as follows:

    @ATTRIBUTE class        {Iris-setosa,Iris-versicolor,Iris-virginica}
   

Values that contain spaces must be quoted. 
```
But I can still change it to `return [v.lower().capitalize() for v in feat.nominal_values]` if you prefer.

sebhrusen: yes, keeping the same order with `v.capitalize()` looks nicer, thanks!

sebhrusen (approved):
thanks @PGijsbers

PR #325, opened 2021-06-14
Add R packages to 'all frameworks' CI

PGijsbers:
Requires #324 to be resolved before R frameworks work on KC1 (part of the `validation` benchmark which is used for framework specific changes only).

Adds `mlr3automl` and `autoxgboost` to the frameworks to run CI against.

### DIFF
--- a/frameworks/autoxgboost/exec.py
+++ b/frameworks/autoxgboost/exec.py
@@ -19,7 +19,7 @@ def run(dataset: Dataset, config: TaskConfig):
 
     meta_results_file = os.path.join(config.output_dir, "meta_results.csv")
     run_cmd(("Rscript --vanilla -e \""
-             "source('{script}'); "
+             ".libPaths('frameworks/autoxgboost/r-packages/'); source('{script}'); "
              "run('{train}', '{test}', target.index = {target_index}, '{type}', '{output}', {cores},"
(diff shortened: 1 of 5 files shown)


frameworks/autoxgboost/exec.py
```diff
@@ -19,7 +19,7 @@ def run(dataset: Dataset, config: TaskConfig):
 
     meta_results_file = os.path.join(config.output_dir, "meta_results.csv")
     run_cmd(("Rscript --vanilla -e \""
-             "source('{script}'); "
+             ".libPaths('frameworks/autoxgboost/r-packages/'); source('{script}'); "
```
sebhrusen, line 22: please use `here` variable instead of hardcoded path

frameworks/mlr3automl/exec.py
```diff
@@ -17,7 +17,7 @@ def run(dataset: Dataset, config: TaskConfig):
 
     meta_results_file = os.path.join(config.output_dir, "meta_results.csv")
     run_cmd(("Rscript --vanilla -e \""
-             "source('{script}'); "
+             ".libPaths('frameworks/mlr3automl/r-packages/'); source('{script}'); "
```
sebhrusen, line 20: see comment above

PR #327, opened 2021-06-18
MLNet - Add option for picking up trainers

LittleLittleCloud:
default trainers are
- lgbm
- rf
- fast tree
- sdca
- lbfgs

### DIFF
--- a/frameworks/MLNet/exec.py
+++ b/frameworks/MLNet/exec.py
@@ -41,6 +41,11 @@ def run(dataset: Dataset, config: TaskConfig):
     os.environ['MODELBUILDER_AUTOML'] = MODELBUILDER_AUTOML
 
     artifacts = config.framework_params.get('_save_artifacts', [])
+    allowedMultiClassifiers = config.framework_params.get('classifiers','LGBM;RF;FASTTREE;LBFGS;SDCA')
+    allowedRegressors = config.framework_params.get('regressors','LGBM;RF;FASTTREE;LBFGS;SDCA')
+    os.environ['AutoMLServiceMultiClassifiers'] = allowedMultiClassifiers
+    os.environ['AutoMLServiceRegressors'] = allowedRegressors
+    
     tmpdir = tempfile.mkdtemp()
     tmp_output_folder = os.path.join(tmpdir, str(config.fold))
     output_dir = output_subdir('models', config=config) if 'models' in artifacts else tmp_output_folder
--- a/resources/frameworks.yaml
+++ b/resources/frameworks.yaml
@@ -127,6 +127,12 @@ MLNet:
   version: 'latest'
   description: |
     MLNET.CLI is a automated machine learning tool implemented by ml.net.
+  params:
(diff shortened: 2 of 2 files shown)


resources/frameworks.yaml
```diff
@@ -127,6 +127,12 @@ MLNet:
   version: 'latest'
   description: |
     MLNET.CLI is a automated machine learning tool implemented by ml.net.
+  params:
+    _save_artifacts: [] # supported option: logs, models
+    classifiers: 'LGBM;RF;FASTTREE;LBFGS;SDCA' # supported trainers: LGBM, RF, FASTTREE, LBFGS, SDCA
+    regressors: 'LGBM;RF;FASTTREE;LBFGS;SDCA' # supported trainers: LGBM, RF, FASTTREE, LBFGS, SDCA
+    automl_type: 'NNI' # supported type: NNI, AutoML
+
```
sebhrusen, line 135: non-default params are not allowed in this file, if you need to try different configurations for your personal use, please add your own definitions in `~/.config/automlbenchmark/framework.yaml`
See https://github.com/openml/automlbenchmark/blob/master/docs/HOWTO.md#custom-configuration for details

frameworks/MLNet/exec.py
```diff
@@ -41,6 +41,11 @@ def run(dataset: Dataset, config: TaskConfig):
     os.environ['MODELBUILDER_AUTOML'] = MODELBUILDER_AUTOML
 
     artifacts = config.framework_params.get('_save_artifacts', [])
+    allowedMultiClassifiers = config.framework_params.get('classifiers','LGBM;RF;FASTTREE;LBFGS;SDCA')
+    allowedRegressors = config.framework_params.get('regressors','LGBM;RF;FASTTREE;LBFGS;SDCA')
+    os.environ['AutoMLServiceMultiClassifiers'] = allowedMultiClassifiers
+    os.environ['AutoMLServiceRegressors'] = allowedRegressors
```
sebhrusen, line 47: As these are not purely technical parameters but a way to configure the behaviour of MLNET, we can't allow passing non-default values here.
Also, by convention, parameters that are not passed directly to the training function (in your case, to `run_cmd(cmd)` below) must be prefixed with underscore.

Summing up these 2 concerns, we can only accept something like:
```suggestion
    allowedMultiClassifiers = config.framework_params.get('_classifiers')
    allowedRegressors = config.framework_params.get('regressors')
    if allowedMultiClassifiers is not None:
        os.environ['AutoMLServiceMultiClassifiers'] = allowedMultiClassifiers
    if allowedRegressors is not None:
        os.environ['AutoMLServiceRegressors'] = allowedRegressors
```

sebhrusen (changes requested):

PGijsbers:
If @LittleLittleCloud does not respond soon I will close this PR. It can always be re-opened if needed.

LittleLittleCloud:
This PR is also based on pretty old commit, you can close it for now.

PR #334, opened 2021-07-02
Replace hardcoded python3 by python{pyv}

mfeurer:
I observed that python3 does not work in my singularity build as it is not a symlink to the system python, but something else:

```
-rwxr-xr-x 1 root root 5482296 Jul  2 07:54 python
-rwxr-xr-x 1 root root 5482296 Jul  2 07:54 python3
lrwxrwxrwx 1 root root      18 Jul  2 07:56 python3.7 -> /usr/bin/python3.7
```

I therefore propose to use the python version that's given by the `{pyv}` variable everywhere, also in the docker file.

In addition, this PR adds some brackets to the singularity file creation and removes a `RUN` which broke my build.

### DIFF
--- a/amlb/runners/singularity.py
+++ b/amlb/runners/singularity.py
@@ -168,10 +168,9 @@ add-apt-repository -y ppa:deadsnakes/ppa
 apt-get update
 apt-get -y install python{pyv} python{pyv}-venv python{pyv}-dev python3-pip
 #update-alternatives --install /usr/bin/python3 python3 $(which python{pyv}) 1
-pip3 install -U pip wheel
 
 # aliases for the python system
-SPIP=python{pyv} -m pip
+SPIP="python{pyv} -m pip"
 SPY=python{pyv}
 
 # Enforce UTF-8 encoding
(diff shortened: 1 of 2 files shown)


mfeurer:
CC @franchuterivera

amlb/runners/singularity.py
```diff
@@ -187,9 +186,10 @@ def _generate_script(self, custom_commands):
 # packages that we need to data pre- and postprocessing without breaking it.
 $SPIP install -U pip wheel
 $SPY -m venv venv
-PIP=/bench/venv/bin/python3 -m pip
-PY=/bench/venv/bin/python3
+PIP="/bench/venv/bin/python{pyv} -m pip"
+PY="/bench/venv/bin/python{pyv} -W ignore"
 #$PIP install -U pip=={pipv} wheel
+ls -al /bench/venv/bin/
```
PGijsbers, line 192:
```suggestion
```

sebhrusen:
@mfeurer is it a specificity of singularity?
This is what I get in a existing docker image:
```
lrwxrwxrwx 1 root root    9 Apr 29 02:53 python -> python3.7
lrwxrwxrwx 1 root root    9 Apr 29 02:53 python3 -> python3.7
lrwxrwxrwx 1 root root   18 Apr 29 02:53 python3.7 -> /usr/bin/python3.7
```

sebhrusen (approved):

mfeurer:
Yes, this is specific to singularity. Docker works fine. Unfortunately, I don't know what causes this difference in behavior.

PR #336, opened 2021-07-02
Allow R-framework version to be specified by commit hash

PGijsbers:
Useful to refer to a version of mlr3automl (since there are no official releases).
Will test this Monday when I have AWS access again (I can't test it on Windows - even in docker I encounter some encoding issue with installing R frameworks).

### DIFF
--- a/frameworks/autoxgboost/setup.sh
+++ b/frameworks/autoxgboost/setup.sh
@@ -9,7 +9,7 @@ if [[ "$VERSION" == "latest" || "$VERSION" == "stable" ]]; then
 fi
 
 if [[ "$VERSION" =~ ^# ]]; then
-  VERSION="${VERSION:1:8}"
+  VERSION="${VERSION:1}"
 else
   # if VERSION is not a hash, it should be a branch (or a format which is not (officially) supported)
   COMMIT=$(git ls-remote "https://github.com/${REPO}" | grep "refs/heads/${VERSION}" | cut -f 1)
@@ -17,7 +17,7 @@ else
     echo "Could not resolve version ${VERSION}. It is not a branch on https://github.com/${REPO}."
     echo "Continuing setup, install_github will try to resolve 'ref=${VERSION}'."
   else
-    VERSION="${COMMIT:0:7}"
+    VERSION=$COMMIT
   fi
 fi
 
@@ -43,4 +43,4 @@ Rscript -e 'options(install.packages.check.source="no"); install.packages(c("rem
 Rscript -e '.libPaths("'"${HERE}/r-packages/"'"); remotes::install_github("'"${REPO}"'", ref="'"${VERSION}"'", lib="'"${HERE}/r-packages/"'")'
 
 OFFICIAL_VERSION=$(Rscript -e '.libPaths("'"${HERE}/r-packages/"'"); packageVersion("autoxgboost")' | awk '{print $2}' | sed "s/[‘’]//g")
-echo "${OFFICIAL_VERSION}#${VERSION}" >> "${HERE}/.installed"
+echo "${OFFICIAL_VERSION}#${VERSION:0:7}" >> "${HERE}/.installed"
--- a/frameworks/autoxgboost/setup.sh
+++ b/frameworks/autoxgboost/setup.sh
@@ -2,10 +2,21 @@
 HERE=$(dirname "$0")
 VERSION=${1:-"stable"}
 REPO=${2:-"ja-thomas/autoxgboost"}
-# currently both stable and latest maps to master branch
+
+# Version can be specified as 'stable', 'latest', a full 40-character commit hash or a branch
 if [[ "$VERSION" == "latest" || "$VERSION" == "stable" ]]; then
-    VERSION="master"
+  VERSION="master"
 fi
+
+if ! [[ "$VERSION" =~ ^[a-fA-F0-9]{40}$ ]]; then
+  # if VERSION is not a hash, it should be a branch (or a format which is not (officially) supported)
+  VERSION=$(git ls-remote "https://github.com/${REPO}" | grep "refs/heads/${VERSION}" | cut -f 1)
+  if [[ -z $VERSION ]]; then
+    echo "Could not resolve version ${VERSION}. It is not a branch on https://github.com/${REPO}."
+    echo "Continuing setup, install_github will try to resolve 'ref=${VERSION}'."
+  fi
+fi
+
 . ${HERE}/../shared/setup.sh "${HERE}"
 if [[ -x "$(command -v apt-get)" ]]; then
   SUDO apt-get update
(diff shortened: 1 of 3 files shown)


sebhrusen (approved):
didn't try it, but looks reasonable

PGijsbers:
I managed to give it a go, and as far as I can tell it seems to work. The only undesirable aspect is that the version is determined by:

  `Rscript -e '.libPaths("'"${HERE}/r-packages/"'"); packageVersion("mlr3automl")' | awk '{print $2}' | sed "s/[‘’]//g" >> "${HERE}/.installed"`
  
Which means that it uses the internal version, which is not generally changed between commits. This is the version that is also recorded for `version` of `results.csv`. I propose to check in `setup.sh` if the specified `VERSION` is a hash and if so record that as version instead. Autoxgboost features the same issue, so I would suggest that in this PR as well should you think it's a good idea.

PGijsbers:
We will add the commit has to the `.installed` file and generate the framework version by joining each line of the file.

PGijsbers:
Decided to go slightly differently and join the official version with the hash on the same line in the `.installed` file.
Otherwise we need to make a change that would affect all frameworks (by making the framework version depend on more than just the first line of `.installed`).

---

For `autoxgboost` and `mlr3-automl` you can now specify the version as a commit hash or branch. 
Latest and stable resolve to 'master' branch.
If a branch is specified, the most recent commit to that branch is found.

Besides the official version (which may or may not uniquely identify a version), the commit hash is also stored in `.installed`.
They are stored together on the first line of the file as `OFFICIAL#HASH`, which is also reported in the results file.
This should make it trivial to rerun experiments with the exact same framework version.

Notes:
 - The commit hash might still at some point become invalid, if it's on a branch which is later deleted (e.g. squash merged with another branch).
 - I decided to keep the 'official' version around since that's more easily identifiable for humans (only for context, since it's not unique).

frameworks/autoxgboost/setup.sh
```diff
@@ -2,10 +2,21 @@
...
 REPO=${2:-"ja-thomas/autoxgboost"}
-# currently both stable and latest maps to master branch
+
+# Version can be specified as 'stable', 'latest', a full 40-character commit hash or a branch
 if [[ "$VERSION" == "latest" || "$VERSION" == "stable" ]]; then
-    VERSION="master"
+  VERSION="master"
+fi
+
+if ! [[ "$VERSION" =~ ^[a-fA-F0-9]{40}$ ]]; then
```
sebhrusen, line 11: asking for the full 40 chars is unusual given that most git commands accept and print only the first few chars of the hash.
It would be more user-friendly to simply use a specific syntax for hashes: `#def123` interpreted as hash `def123`.
Besides this is what was recommended and implemented for autosklearn: https://github.com/openml/automlbenchmark/pull/309

PGijsbers: I wasn't sure about shorter hashes since over time they wouldn't necessarily have to be unique.
After thinking some more, should that ever happen, we know we should use the first hash in the commit history that matches it, so it's not really a problem. Will change.

PGijsbers: Done.

frameworks/autoxgboost/setup.sh
```diff
@@ -2,10 +2,25 @@
...
-# currently both stable and latest maps to master branch
+
+# Version can be specified as 'stable', 'latest', a branch or a commit hash (indicated by starting with '#')
 if [[ "$VERSION" == "latest" || "$VERSION" == "stable" ]]; then
-    VERSION="master"
+  VERSION="master"
+fi
+
+if [[ "$VERSION" =~ ^# ]]; then
+  VERSION="${VERSION:1:8}"
```
sebhrusen, line 12:
```suggestion
  VERSION="${VERSION:1}"
```
well, my point was not to use only 7 chars if user is providing more, I just didn't want to force the user to provide the full hash when the "default" 7 chars is almost always enough.

frameworks/autoxgboost/setup.sh
```diff
@@ -2,10 +2,25 @@
...
+else
+  # if VERSION is not a hash, it should be a branch (or a format which is not (officially) supported)
+  COMMIT=$(git ls-remote "https://github.com/${REPO}" | grep "refs/heads/${VERSION}" | cut -f 1)
+  if [[ -z $COMMIT ]]; then
+    echo "Could not resolve version ${VERSION}. It is not a branch on https://github.com/${REPO}."
+    echo "Continuing setup, install_github will try to resolve 'ref=${VERSION}'."
+  else
+    VERSION="${COMMIT:0:7}"
```
sebhrusen, line 20:
```suggestion
    VERSION=$COMMIT
```

frameworks/autoxgboost/setup.sh
```diff
@@ -27,4 +42,5 @@ mkdir "${HERE}/r-packages/"
 Rscript -e 'options(install.packages.check.source="no"); install.packages(c("remotes", "mlr", "mlrMBO", "mlrCPO", "farff", "GenSA", "rgenoud", "xgboost"), repos="https://cloud.r-project.org/", lib="'"${HERE}/r-packages/"'")'
 Rscript -e '.libPaths("'"${HERE}/r-packages/"'"); remotes::install_github("'"${REPO}"'", ref="'"${VERSION}"'", lib="'"${HERE}/r-packages/"'")'
 
-Rscript -e '.libPaths("'"${HERE}/r-packages/"'"); packageVersion("autoxgboost")' | awk '{print $2}' | sed "s/[‘’]//g" >> "${HERE}/.installed"
+OFFICIAL_VERSION=$(Rscript -e '.libPaths("'"${HERE}/r-packages/"'"); packageVersion("autoxgboost")' | awk '{print $2}' | sed "s/[‘’]//g")
+echo "${OFFICIAL_VERSION}#${VERSION}" >> "${HERE}/.installed"
```
sebhrusen, line 46:
```suggestion
echo "${OFFICIAL_VERSION}#${VERSION:0:7}" >> "${HERE}/.installed"
```
for results file, we can now limit to 7 chars.

frameworks/mlr3automl/setup.sh
```diff
@@ -1,9 +1,27 @@
...
+  # if VERSION is not a hash, it should be a branch (or a format which is not (officially) supported)
+  COMMIT=$(git ls-remote "https://github.com/${REPO}" | grep "refs/heads/${VERSION}" | cut -f 1)
+  if [[ -z $COMMIT ]]; then
+    echo "Could not resolve version ${VERSION}. It is not a branch on https://github.com/${REPO}."
+    echo "Continuing setup, install_github will try to resolve 'ref=${VERSION}'."
+  else
+    VERSION="${COMMIT:0:7}"
+  fi
+fi
+
```
sebhrusen, line 24: same as above

frameworks/autoxgboost/setup.sh
PGijsbers: Gotcha. I thought you also weren't satisfied by how noisy the framework version looks with 40 character hash.

PGijsbers: Nvm, I see you addressed that when writing to the results file.

PR #339, opened 2021-07-07
Add requirements on the version of pandas in setup.sh of the framework flaml

qingyun-wu:


PGijsbers:
Can you give an indication which issue you are trying to resolve?
I assume you're running into [this pandas issue](https://github.com/pandas-dev/pandas/issues/42345)?

In which case we need to ensure that we fix our library versions better across all frameworks (and hopefully soon there will be a new pandas release which fixes this as well).
To remedy this issue in the mean time, clearing your cache should probably work (as the issue arises when trying to load a pandas <1.3 dataframe with pandas >1.3).

If it is not related to that issue, please explain.

PGijsbers:
I guess the issue here is that the framework has a fixed version to 1.2.4. We don't care for that version specifically, but we do want to only explicitly bump versions. Meanwhile the framework venv will install the latest version (or whichever version specified), which happens to be pandas 1.3 for some frameworks. This means any dataframe pickled in the benchmark process and unpickled in the framework process raises an error.

@sebhrusen I propose we raise our requirement to 1.3 now, which requires clearing the cache but otherwise allows all new runs to work. And then potentially bumping it again next week if they release a hotfix, so users that missed the upgrade to 1.3 can keep their old cache.

qingyun-wu:
> Can you give an indication which issue you are trying to resolve?
> I assume you're running into [this pandas issue](https://github.com/pandas-dev/pandas/issues/42345)?
> 
> In which case we need to ensure that we fix our library versions better across all frameworks (and hopefully soon there will be a new pandas release which fixes this as well).
> To remedy this issue in the mean time, clearing your cache should probably work (as the issue arises when trying to load a pandas <1.3 dataframe with pandas >1.3).
> 
> If it is not related to that issue, please explain.

Hi @PGijsbers , thanks for your reply. This PR is indeed related to the pickle issues caused by pandas version (although the error is not exactly the same as what was described in the issue). Attached below are the error msg and the traceback I got.

```
  File "/Users/qingyunwu/Git/Github/automlbenchmark/amlb/benchmark.py", line 511, in run
    meta_result = self.benchmark.framework_module.run(self._dataset, task_config)
  File "/Users/qingyunwu/Git/Github/automlbenchmark/frameworks/flaml/__init__.py", line 24, in run
  File "/Users/qingyunwu/Git/Github/automlbenchmark/frameworks/shared/caller.py", line 143, in run_in_venv
    res[name] = deserialize_data(res[name]) if res[name] is not None else None
  File "/Users/qingyunwu/Git/Github/automlbenchmark/amlb/utils/process.py", line 702, in profiler
    return fn(*args, **kwargs)
  File "/Users/qingyunwu/Git/Github/automlbenchmark/frameworks/shared/serialization.py", line 74, in deserialize_data
    return pd.read_pickle(path, compression=ser_config.pandas_compression)
  File "/Users/qingyunwu/Git/Github/automlbenchmark/venv/lib/python3.7/site-packages/pandas/io/pickle.py", line 208, in read_pickle
    return pc.load(handles.handle, encoding=None)
  File "/Users/qingyunwu/Git/Github/automlbenchmark/venv/lib/python3.7/site-packages/pandas/compat/pickle_compat.py", line 249, in load
    return up.load()
  File "/Users/qingyunwu/miniconda3/lib/python3.7/pickle.py", line 1088, in load
    dispatch[key[0]](self)
  File "/Users/qingyunwu/miniconda3/lib/python3.7/pickle.py", line 1385, in load_stack_global
    self.append(self.find_class(module, name))
  File "/Users/qingyunwu/Git/Github/automlbenchmark/venv/lib/python3.7/site-packages/pandas/compat/pickle_compat.py", line 189, in find_class
    return super().find_class(module, name)
  File "/Users/qingyunwu/miniconda3/lib/python3.7/pickle.py", line 1426, in find_class
    __import__(module, level=0)
ModuleNotFoundError: No module named 'pandas._libs.arrays'
```

And you are right about the reason: Our framework will install the latest version, which happens to be pandas 1.3 (this may be true for some other frameworks as well). 

Please feel free to let me know if you need more information. Thank you!

Best,
Qingyun

sebhrusen:
@qingyun-wu pandas version mismatch can be used by… not pickling pandas dataframes to exchange data between the amlb process and `flaml` process.
See for example: https://github.com/openml/automlbenchmark/blob/master/frameworks/GAMA/__init__.py#L13

Please try it, and if it works, preferrably use this solution, thanks.

sebhrusen:
@PGijsbers maybe we could actually use `parquet` as the default serialization format, to (almost completely) avoid version mismatch issues with pandas, wdyt?
I think it's just slightly slower afair.

PGijsbers:
That would probably be the best solution as it should also be robust going forward.

PGijsbers:
Because #345 made the default serialization format parquet, the error caused by the pickle incompatibility no longer occurs. FLAML works on master (verified with `python runbenchmark.py flaml`). If you still want to make changes to the FLAML requirements feel free to re-open or create a new PR.

Issue #337, opened 2021-07-06
TunedRandomForest baseline not working?


mfeurer:
Hi together,

I was just trying the TunedRandomForest baseline and I think it is not working properly for all datasets, see for example the output of `python3 runbenchmark.py TunedRandomForest test -s force`:
```
Running benchmark `TunedRandomForest` on `test` framework in `local` mode.
Loading frameworks definitions from ['/home/feurerm/sync_dir/projects/openml/automlbenchmark/resources/frameworks.yaml'].
Loading benchmark constraint definitions from ['/home/feurerm/sync_dir/projects/openml/automlbenchmark/resources/constraints.yaml'].
Loading benchmark definitions from /home/feurerm/sync_dir/projects/openml/automlbenchmark/resources/benchmarks/test.yaml.
Setting up framework TunedRandomForest.
Running cmd `/home/feurerm/sync_dir/projects/openml/automlbenchmark/frameworks/TunedRandomForest/setup.sh stable`
shared/setup.sh /home/feurerm/sync_dir/projects/openml/automlbenchmark/frameworks/TunedRandomForest true
Collecting pip
  Downloading https://files.pythonhosted.org/packages/47/ca/f0d790b6e18b3a6f3bd5e80c2ee4edbb5807286c21cdd0862ca933f751dd/pip-21.1.3-py3-none-any.whl (1.5MB)
Collecting wheel
  Using cached https://files.pythonhosted.org/packages/65/63/39d04c74222770ed1589c0eaba06c05891801219272420b40311cd60c880/wheel-0.36.2-py2.py3-none-any.whl
Installing collected packages: pip, wheel
  Found existing installation: pip 19.0.3
    Uninstalling pip-19.0.3:
      Successfully uninstalled pip-19.0.3
Successfully installed pip-21.1.3 wheel-0.36.2
PY=/home/feurerm/sync_dir/projects/openml/automlbenchmark/frameworks/TunedRandomForest/venv/bin/python -W ignore
PIP=/home/feurerm/sync_dir/projects/openml/automlbenchmark/frameworks/TunedRandomForest/venv/bin/python -m pip
Collecting psutil==5.8.0
  Downloading psutil-5.8.0-cp37-cp37m-manylinux2010_x86_64.whl (296 kB)
Collecting ruamel.yaml.clib==0.2.2
  Downloading ruamel.yaml.clib-0.2.2-cp37-cp37m-manylinux1_x86_64.whl (547 kB)
[...]

PR #340, opened 2021-07-07
Update the input to the evaluation

PGijsbers:
Updates the old `X_enc` references (part of #337)


sebhrusen (approved):
wasn't TRF part of the workflow? I'm surprised this was not detected earlier

PGijsbers:
The final fit and predict used the correct variables, so those did not give an error - trf produced results as expected. The `X_enc` error happened within a `try/except` block. I'm not exactly sure what exceptions we were expecting so it's probably worth testing how it functions without. I created #341

PR #342, opened 2021-07-08
Sort categoricals in alphabetical order

PGijsbers:
This is in particular important for the target column in classification tasks, where AutoWEKA preserves the order of those variables in the column-order for their class probabilities. And we expect those to be alphabetical. This approach seemed more sensible to me than either reading the ARFF header or the WEKA predictions file to figure out the order WEKA used.

I am not sure how to handle custom ARFF input in this scenario. Automatic splits for it aren't supported yet ([I think?](https://github.com/openml/automlbenchmark/blob/9e5b66d11a804b19595d602f3f3e7e54e544e201/amlb/datasets/file.py#L224)), but an explicit `train` and `test` file would simply have to enforce this order itself.

**Edit: Modifies the CI script to detect changed files in subdirectories too.**

### DIFF
--- a/.github/workflows/run_all_frameworks.yml
+++ b/.github/workflows/run_all_frameworks.yml
@@ -39,8 +39,6 @@ jobs:
           echo Common files: $common_files
           echo Changed files: $changed_files
 
-          shopt -u globstar
-
           is_common=1
           for f in $changed_files
           do
--- a/.github/workflows/run_all_frameworks.yml
+++ b/.github/workflows/run_all_frameworks.yml
@@ -25,6 +25,10 @@ jobs:
         name: Detect Common Changes  # detect if any changes occurred that should trigger all frameworks.
         run: |
           changed_files=$(git diff --name-only HEAD..$GITHUB_BASE_REF)
+
+          # globs should be interpreted recursively to find all files
+          shopt -s globstar
+
           common_files="amlb/** "\
           "resources/** "\
           "frameworks/shared/** "\
@@ -34,6 +38,8 @@ jobs:
           "requirements.txt"
           echo Common files: $common_files
           echo Changed files: $changed_files
+
+          shopt -u globstar
 
           is_common=1
           for f in $changed_files
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -262,8 +262,8 @@ class ArffSplitter(DataSplitter[str]):
             # so we need to adapt the header accordingly.
             # Not doing so causes an issue in the R packages.
             if set(v.lower() for v in feat.nominal_values) == {"true", "false"}:
-                return [v.lower().capitalize() for v in feat.nominal_values]
-            return feat.nominal_values
+                return sorted(v.lower().capitalize() for v in feat.nominal_values)
+            return sorted(feat.nominal_values)
         return None
 
 


PGijsbers:
Looks like the CI script doesn't detect common files in subdirectories. Will add it to this PR today.

PGijsbers:
Note that you have to empty your cache (or at least the `dataset_train/test_X.arff` files), since they need to be re-generated.

.github/workflows/run_all_frameworks.yml
```diff
@@ -35,6 +39,8 @@ jobs:
           echo Common files: $common_files
           echo Changed files: $changed_files
 
+          shopt -u globstar
```
sebhrusen, line 42: not sure you need to turn it off, the enabling should apply only to current shell

amlb/datasets/openml.py
```diff
@@ -262,8 +262,8 @@ def _get_categorical_values(self, col):
             # so we need to adapt the header accordingly.
             # Not doing so causes an issue in the R packages.
             if set(v.lower() for v in feat.nominal_values) == {"true", "false"}:
-                return [v.lower().capitalize() for v in feat.nominal_values]
-            return feat.nominal_values
+                return sorted(v.lower().capitalize() for v in feat.nominal_values)
+            return sorted(feat.nominal_values)
```
sebhrusen, line 266: interesting, was I doing this sorting somewhere else when rewriting the ARFF files in previous logic — meaning before changing the way those arff files are loaded / rewritten.

PGijsbers: I can't really find any evidence of it. It looks like the bug already existed in 97a4ac2739b446734f9aa59c5a10a244bbfd9438.

PGijsbers: I'm not inclined to investigate further on when it was introduced if that's okay. Sometime between the workshop paper (AutoWEKA had AUC>0.5 so it worked) and just before the data loading rework.

sebhrusen (approved):

amlb/datasets/openml.py
sebhrusen: I'm a bit annoyed that we have to transform the original ARFF file header for the benefit of one framework: it looks like a bug in AutoWeka imo.

sebhrusen: > AutoWEKA preserves the order of those variables in the column-order for their class probabilities.  And we expect those to be alphabetical.

Is not the `probabilities_labels` in `exec.py` that simply don't match the columns produced by autoweka?

PGijsbers: I guess it's a matter of perspective. 

 1. Auto-WEKA preserves class order as defined in the ARFF file.
 2. In the dataset meta-data we use OpenML metadata*.
 3. In Auto-WEKA exec.py there is now no longer a way to see the original order as in the ARFF file.

Makes that I don't see a way to easily set `probabilities_labels = dataset.target.values` correctly.

\* Unfortunately OpenML meta-data doesn't match the ARFF header either: [OpenML#1114](https://github.com/openml/OpenML/issues/1114).

Should #1114 be resolved s.t. the feature data matches the ARFF data I expect the bug to continue without further action.
Since without this PR the ARFF header remains unsorted, yet we also [sort values in the metadata](https://github.com/openml/automlbenchmark/blob/9e5b66d11a804b19595d602f3f3e7e54e544e201/amlb/datasets/openml.py#L106). So we would still have to change at least that.

So (I think) it came down to either keeping additional meta-data about the original ARFF header order or just sorting the header when writing to it. I went with the latter. I am not completely happy with it, but it seemed least invasive.

PGijsbers: At some point in the future when parquet files are the primary source of data, we have to generate the ARFF files ourselves have control over category order (except for custom ARFF files that are already split, I don't know if it would be problematic to just not support those).

sebhrusen (changes requested):
needs to be fixed at AutoWeka level

PGijsbers:
Re-opened PR. We noticed that the definition of the features as provided by the openml server are not consistent. That means as long as we rely on them (and that issue isn't fixed), the ARFF files we produced are not deterministic (specifically the order of nominal values in the header). This PR fixes that.

PGijsbers:
Closes #347

sebhrusen (approved):
this ensures that the order of nominal values in generated ARFF files is deterministic

Issue #344, opened 2021-07-08
Old Pandas ver in requirements.txt


drheavy:
In requirements.txt there is a Pandas version 1.2.4, but with this version I get an error on each test ("AttributeError: Can't get attribute 'new_block' on <module 'pandas.core.internals.blocks' from '.../projects/automlbenchmark/venv/lib/python3.8/site-packages/pandas/core/internals/blocks.py")
After Pandas updating to the last version 1.3.0 it works well. I think requirements.txt should be updated.

PGijsbers:
Per the discussion in #339 we're planning to bypass this problem altogether by avoiding working with pickled dataframes and instead prefer to use parquet.

PGijsbers:
I'm leaving this issue open to acknowledge there currently is an issue with the pandas version incompatibility.

PGijsbers:
There should no longer be a need to upgrade pandas since we no longer use pickle as the default serialization format (#345).

PR #345, opened 2021-07-09
Make Parquet the default serialization format

PGijsbers:
As discussed in #339, the pandas version discrepancy causes issues for frameworks. Specifically there currently is a bug that prevents pickled dataframes to compatible between pandas 1.2.4 and 1.3.0. Moving forward using `parquet` as the serialized format between main and subprocesses should still offer good performance while being less likely to cause issues with mismatched versions.

Note: this is not only important for frameworks that use serialized input data, but their output (the reason AutoGluon currently fails despite already having Parquet input).

### DIFF
--- a/frameworks/shared/requirements.txt
+++ b/frameworks/shared/requirements.txt
@@ -4,7 +4,11 @@
 #
 #    pip-compile frameworks/shared/requirements.in
 #
+numpy==1.21.0
+    # via pyarrow
 psutil==5.8.0
+    # via -r frameworks/shared/requirements.in
+pyarrow==4.0.1
     # via -r frameworks/shared/requirements.in
 ruamel.yaml.clib==0.2.2
     # via ruamel.yaml
--- a/frameworks/GAMA/__init__.py
+++ b/frameworks/GAMA/__init__.py
@@ -9,8 +9,6 @@ def setup(*args, **kwargs):
 
 def run(dataset: Dataset, config: TaskConfig):
     from frameworks.shared.caller import run_in_venv
-    from frameworks.shared.serialization import ser_config
-    ser_config.pandas_serializer = 'parquet'  # can't use pickle due to pandas version mismatch
 
     data = dict(
         target=dataset.target.name,
--- a/frameworks/GAMA/requirements.txt
+++ b/frameworks/GAMA/requirements.txt
@@ -1,2 +1,1 @@
 packaging
-pyarrow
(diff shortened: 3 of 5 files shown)


PGijsbers:
closes #344 and #339.

sebhrusen:
you'll also need to add `pyarrow` to the `shared/requirements`

PGijsbers:
> you'll also need to add pyarrow to the shared/requirements

Thanks, you're right. In that same vein, shouldn't `pandas` actually also be there?

frameworks/shared/requirements.txt
```diff
@@ -4,8 +4,12 @@
 #
 #    pip-compile frameworks/shared/requirements.in
 #
+numpy==1.21.0
+    # via pyarrow
```
sebhrusen, line 8: I hope this is not introducing new version conflicts, the main advantage of `pickle` is that it was avoiding this : frameworks are supposed to install their dependencies after this one so they can always install another `numpy` version, hopefully still compatible with `pyarrow`… and in any case they can switch to alternatives (pickle, hdf, and json in the worst case)

sebhrusen (approved):

sebhrusen:
> In that same vein, shouldn't pandas actually also be there? I realize that all current frameworks already install it as a dependency, but (except for GAMA? For now?) the same is true for pyarrow. Then again, it's probably better to impose versions at least as possible for framework venvs.

well, the idea of venv is to let the framework pick the versions it needs.
Besides, not every framework uses the pandas dataframe (H2O, MLPlan don't) so we should avoid imposing unnecessary dependencies otherwise we well end up with the same issues than before creating venvs.
Also, pandas comes up with its own set of dependencies and again we sohuld really keep the shared dependencies to a minimum: `pickle` was allowing this. Here with `pyarrow` we're also introducing `numpy` as a dependency that may conflict with the version required by the framework, so in my opinion, it's not fully satisfactory compared with the original goal of those venvs.

PGijsbers:
> it's not fully satisfactory compared with the original goal of those venvs

Partially agree. Absolutely right for those frameworks that don't use the dataframe altogether, but for those that do (as we see) `pickle` still has dependencies, but makes them implicit. Would be happy to revisit this in the future.

frameworks/shared/requirements.txt
PGijsbers: Since `pyarrow` only interfaces with `numpy` through its public functions (I should hope), it might be even _more_ compatible with various `numpy` versions?

PR #346, opened 2021-07-09
extract labels from Weka predictions

sebhrusen:
WEKA annoyingly doesn't provide the labels in any of its predictions file (tried various formats), making the predictions files interpretables only in relation of the input file, which is bad design imo.

This hack is still able to collect all labels in most cases.

### DIFF
--- a/frameworks/AutoWEKA/exec.py
+++ b/frameworks/AutoWEKA/exec.py
@@ -78,6 +78,15 @@ def run(dataset: Dataset, config: TaskConfig):
     probabilities_labels = dataset.target.values
     if not os.path.exists(weka_file):
         raise NoResultError("AutoWEKA failed producing any prediction.")
+
+    collected_labels = [None] * len(probabilities_labels)
+
+    def extract_pred(pred):
+        idx, val = pred.split(':')
+        idx = int(idx)
+        collected_labels[idx-1] = val
+        return val
+
     with open(weka_file, 'r') as weka_file:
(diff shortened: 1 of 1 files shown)


frameworks/AutoWEKA/exec.py
```diff
@@ -78,19 +78,37 @@ def run(dataset: Dataset, config: TaskConfig):
...
-            _, tru = actual.split(':')
+            pred = extract_pred(predicted)
+            tru = extract_pred(actual)
             probabilities.append(pred_probabilities)
             predictions.append(pred)
             truth.append(tru)
 
+    collab = set(collected_labels)
+    unused_labels = [l for l in probabilities_labels if l not in collab]
+    if len(unused_labels) > 1:
```
PGijsbers, line 105: What unused label are we expecting (that makes it `>1` instead of `>0`)?

sebhrusen: we only collect labels from predictions and truth values so for a multiclass problem (let's say 10 classes), we can imagine one fold for which we have only 7 distinct values in the target feature (and same, more or less, in the predictions), in which case we won't see all values.
If missing only 1, then we know it for sure, but not if missing 2 or more.

PGijsbers: > If missing only 1, then we know which one it is for sure

That was the part I missed. I should have stopped to think 😅

PR #349, opened 2021-07-14
Fix AW permissions

mfeurer:
Unfortunately, the previous change to `chmod` to be independent of the used SMAC version didn't work. This here works for me with singularity.


PGijsbers (approved):
Tested on Docker (Windows Host), AWS (for local mode), both work (not that I expected anything to break).

PR #358, opened 2021-07-23
Allow specific H2O version to be installed

PGijsbers:


sebhrusen (approved):
Thanks for that fix @PGijsbers

Issue #330, opened 2021-06-28
PR from a fork fails CI


PGijsbers:
It looks like CI fails if the branch is on a fork, [example](https://github.com/openml/automlbenchmark/pull/329/checks?check_run_id=2883590478). CI actually should probably just check out the PR head and test that. We might also opt for running both pr and push as is conventional, but considering the workload that feels like a waste of time and resources.
Maybe:
 - run unit tests on both `push`  and `pull_request`
 - run the appropriate validation or test benchmark on ` pull_request` 
 
 edit: A related issue (that could get tackled at the same time?) is that the detected changes currently detect changes from the current head of master and the head of the PR. This means that if master had changes after the branch branched off, those changed files will also influence the decision to run CI for a specific framework or all frameworks. This decision should purely be based off which files are changed in the PR (since the integration test is proposed to run on `pull_request`, meaning the other changes in master are included anyway).

PR #359, opened 2021-07-23
Install opencv-python to avoid error on import

PGijsbers:
With the current setup, an error is produced on first import. This means the final statement in the setup script exits with an error, which prevents the benchmark from running on light automl (see [lama#53](https://github.com/sberbank-ai-lab/LightAutoML/issues/53) for more information).

edit: I still noticed an error running the framework (at least on iris and APSFailure). But it's unrelated. Looks like it probably has to do with changes to the input data (encoding), I'll look into it separately.

### DIFF
--- a/frameworks/lightautoml/setup.sh
+++ b/frameworks/lightautoml/setup.sh
@@ -10,6 +10,7 @@ fi
 # creating local venv
 . ${HERE}/../shared/setup.sh ${HERE} true
 
+apt-get install -y python3-opencv
 
 if [[ "$VERSION" == "stable" ]]; then
     PIP install --no-cache-dir -U ${PKG}


sebhrusen (approved):
Changes look good.
Just got worried by an error in `lightautoml` on th e1st checks:
```
ValueError: invalid literal for int() with base 10: 'neg'
```
Tried to rerun the checks but it didn't rerun the same frameworks :(

PGijsbers:
That's because there's an error in the CI script that detects the changes between the current HEAD of master instead of where it branched off (it's mentioned in #330).

PGijsbers:
But yes unfortunately `lightautoml` needs some more work :(

Issue #357, opened 2021-07-22
NoResultError: parquet must have string column names


Innixma:
Running AutoGluon on latest AutoMLBenchmark mainline, I'm seeing the following error for `kc1`, `MiniBooNE` and `pc4` datasets:

```
NoResultError: parquet must have string column names
```

Looking deeper, it appears that AutoGluon successfully trained on all of the datasets, but at the very end this error occurred.

An example log output of `kc1`:

```
[INFO] [amlb.print:20:51:29.770]                       model  score_test  score_val  pred_time_test  pred_time_val     fit_time  pred_time_test_marginal  pred_time_val_marginal  fit_time_marginal  stack_level  can_infer  fit_order
[INFO] [amlb.print:20:51:30.091] 0   RandomForestGini_BAG_L2    0.866271   0.876279       30.125159      27.074952  1748.063022                 0.112795                0.125331           0.604155            2       True         19
[INFO] [amlb.print:20:51:30.091] 1           CatBoost_BAG_L2    0.865136   0.869832       30.318362      27.270582  1932.345979                 0.305997                0.320961         184.887112            2       True         21
[INFO] [amlb.print:20:51:30.091] 2    NeuralNetFastAI_BAG_L2    0.864438   0.855177       36.343328      32.203115  2255.920542                 6.330963                5.253495         508.461675            2       True         24
[INFO] [amlb.print:20:51:30.091] 3     ExtraTreesGini_BAG_L2    0.861819   0.883482       30.128556      27.078614  1748.063162                 0.116192                0.128994           0.604295            2       True         22
[INFO] [amlb.print:20:51:30.091] 4   RandomForestEntr_BAG_L2    0.860335   0.872031       30.126674      27.074091  1748.162348                 0.114310                0.124470           0.703481            2       True         20
[INFO] [amlb.print:20:51:30.091] 5   RandomForestEntr_BAG_L1    0.860161   0.823200        0.114205       0.122945     0.703191                 0.114205                0.122945           0.703191            1       True          6
[INFO] [amlb.print:20:51:30.091] 6       WeightedEnsemble_L3    0.858153   0.885198       48.502545      46.556897  2479.222977                 0.002307                0.001428           1.611111            3       True         28
[INFO] [amlb.print:20:51:30.092] 7     ExtraTreesEntr_BAG_L1    0.856931   0.830755        0.116041       0.126892     0.606815                 0.116041                0.126892           0.606815            1       True          9
[...]

PR #360, opened 2021-07-23
Column names of serialized pandas dataframes must be strings

sebhrusen:
https://github.com/openml/automlbenchmark/issues/357


sebhrusen:
@PGijsbers maybe we'll need to add `kc1` to the validation workflow.

PGijsbers:
It is included in the validation set [here](https://github.com/openml/automlbenchmark/blob/7d478ebabe15c3ac7348bd729465f07eb2b5d8a5/.github/workflows/run_all_frameworks.yml#L60), which means that it is executed whenever a change is made to a specific framework (as opposed to any shared components).

PGijsbers (approved):

PR #361, opened 2021-07-24
Fix LAMA - don't return `truth`

PGijsbers:
When processing the results, if `truth` is returned it is processed based on `is_encoded` - regardless of if only the predictions are encoded or also truth ([here](https://github.com/openml/automlbenchmark/blob/c6eb3784ddde723fc4c7a3230886b1a67d142cb3/amlb/results.py#L290-L298)).

I'm not entirely sure what the original reason was, but it seems to me that there is no good reason for allowing the frameworks return `truth` (or even pass labels of the test data in the first place)? I'd be curious to know why it's a thing (and consequently see if perhaps we should rework that in the future).

### DIFF
--- a/frameworks/lightautoml/exec.py
+++ b/frameworks/lightautoml/exec.py
@@ -65,7 +65,6 @@ def run(dataset, config):
         output_file=config.output_predictions_file,
         probabilities=probabilities,
         predictions=predictions,
-        truth=y_test,
         target_is_encoded=is_classification,
         training_duration=training.duration,
         predict_duration=predict.duration,
--- a/frameworks/lightautoml/exec.py
+++ b/frameworks/lightautoml/exec.py
@@ -53,6 +53,8 @@ def run(dataset, config):
 
         predictions = np.argmax(probabilities, axis=1)
         class_map = automl.outer_pipes[0].ml_algos[0].models[0][0].reader.class_mapping
+        if class_map is None and df_train[label].dtype == bool:
+            class_map = {False: 0, True: 1}
         if class_map:
             column_to_class = {col: class_ for class_, col in class_map.items()}
             predictions = list(map(column_to_class.get, predictions))
--- a/frameworks/lightautoml/exec.py
+++ b/frameworks/lightautoml/exec.py
@@ -42,6 +42,7 @@ def run(dataset, config):
     with Timer() as predict:
         preds = automl.predict(X_test).data
 
+    probabilities_labels = None
     if is_classification:
         probabilities = preds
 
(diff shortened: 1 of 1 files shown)


sebhrusen (approved):
OK, I didn't get that lightsautoml was returning encoded predictions even when the input target is not encoded.
I just hope that they're encoded the same way as we usually do it so that we can decode them correctly: based on `iris` it looks ok, I hope there's no edge case.

PGijsbers:
You're right, I didn't look closely enough at the rest of the script and just put on a band-aid. I opened an issue [in their repo](https://github.com/sberbank-ai-lab/LightAutoML/issues/57) to clarify. I'll hold off merging this for now.

sebhrusen:
```
for TabularUtilizedAutoML (as it is a composition of TabularAutoMLs) it is a little bit harder - automl.outer_pipes[0].ml_algos[0].models[0].reader.class_mapping, as we go to the inner automl model and take its Reader property.
```
it looks like lightautoml integration needs a fix

PGijsbers:
Agreed. The `probability_labels` should be passed and the integers converted to labels.

It still leaves the original question of why `y_test` is available to the frameworks to begin with, though that is outside the scope of this PR.

PGijsbers:
Improved, predictions are now always converted back to their original domain and `probabilities_labels` is given. We might still want to wait for a response on https://github.com/sberbank-ai-lab/LightAutoML/issues/49 before merging.

frameworks/lightautoml/exec.py
```diff
@@ -51,7 +52,13 @@ def run(dataset, config):
...
 
         predictions = np.argmax(probabilities, axis=1)
-
+        class_map = automl.outer_pipes[0].ml_algos[0].models[0][0].reader.class_mapping
+        if class_map is None and df_train[label].dtype == bool:
+            class_map = {False: 0, True: 1}
+        if class_map:
+            column_to_class = {col: class_ for class_, col in class_map.items()}
+            predictions = list(map(column_to_class.get, predictions))
+            probabilities_labels = list(class_map)
```
sebhrusen, line 61: If I understand that `col` is always an integer, then the following is the only way to guarantee that `probabilities_labels` will be sorted correctly:
```suggestion
            probabilities_labels = [class_map[k] for k in sorted(class_map)]
```

PGijsbers: `class_map` is a dictionary from class to column, e.g. `{"b": 0, "a": 1}` sorted in order of columns.
Should `probabilities_labels` not match the order of the columns of `probabilities`? `sorted(class_map)` will just put the labels in alphabetical order. If you mean that you don't want to rely on the order of `class_map` then I suppose you'd want `[column_to_class[col] for col in sorted(column_to_class)]`?

sebhrusen: > class_map is a dictionary from class to column, e.g. {"b": 0, "a": 1} sorted in order of columns.

is it guaranteed by contract? as hidden implementation detail?

> [column_to_class[col] for col in sorted(column_to_class)]?

yes, that's what I meant, sorry.

PGijsbers: I can't find `class_mapping` in their documentation at all so I will assume it's an implementation detail and explicitly ensure the order.

sebhrusen: thanks

PGijsbers:
```suggestion
            probabilities_labels = [column_to_class[col] for col in sorted(column_to_class)]
```

sebhrusen (approved):

Issue #370, opened 2021-07-28
Scipy sparse matrices not handled correctly by TPOT and autosklearn


sebhrusen:
- serialization of sparse matrices was not applied correctly.
- once fixed, the frameworks still fail with the following errors:

```text
# TPOT
  File "/Users/seb/repos/ml/automlbenchmark/frameworks/TPOT/venv/lib/python3.7/site-packages/tpot/base.py", line 1359, in _check_dataset
    self.config_dict
ValueError: Not all operators in None supports sparse matrix. Please use "TPOT sparse" for sparse matrix.
```

```text
#autosklearn
  File "/Users/seb/repos/ml/automlbenchmark/frameworks/autosklearn/venv/lib/python3.7/site-packages/sklearn/utils/multiclass.py", line 288, in type_of_target
    if y.ndim > 2 or (y.dtype == object and len(y) and
TypeError: len() of unsized object
```
We'll improve support for sparse data in a future version: for now, we can simply deserialize the sparse matrices as dense matrices for the frameworks that don't use pandas.

mfeurer:
Just checking - are these sparse target matrices `y`? We might indeed not have tests for that.

CC @eddiebergman

sebhrusen:
@mfeurer in this case both `X` and `y` are indeed sparse, not sure this makes sense for `y`.
I currently fixed this by turning both into arrays as I thought the problem was `X`, but it's very possible that for some frameworks, it's only necessary to do this for `y`.

mfeurer:
Thanks for the clarification. Auto-sklearn should support sparse `X`, but we'll check, and will also check what the behavior for sparse `y` values is.

sebhrusen:
@mfeurer for autosklearn, sparse `X` with dense `y` seems to work fine (and faster), meaning that in your case, sparse `y` was the issue.
Thanks for noticing this: ideally we'd like to have frameworks using sparse data whenever possible, so I'll probably just make the `y`s dense by default, and see individually for each framework regarding `X`.
cc: @PGijsbers

eddiebergman:
@sebhrusen It's probably in the interest of `autosklearn` to handle spare `y` correctly, I'll have a look into it

sebhrusen:
@eddiebergman Sure, just mentioning that we have a workaround on our side for now that also seems to work for other frameworks. Thanks for fixing it on your side too.

eddiebergman:
Hi @sebhrusen,

[...]

PR #371, opened 2021-07-28
Issue 370: ensure that sparse data are handled by frameworks that don't support them

sebhrusen:
https://github.com/openml/automlbenchmark/issues/370

I ended up trying to provide the best strategy regarding sparse data handling for each framework.

By default, sparse data (scipy matrices or pandas DataFrame with sparse columns) retrieved from the datasourse (openml or parsed from file) remain sparse.
Some frameworks (`autosklearn`, `TPOT`) don't work if `y` is sparse: for those, we make `y` dense.
Some other frameworks (`flaml`, `GAMA`, `lightautoml`, `mljar`) fail with sparse dataframes: for those, both `X` and `y` are made dense at the last possible moment (when deserializing data in the framework's process).
Some other (`Autogluon`) consume parquet files directly and we can't make those sparse as `pyarrow` fails saving dataframes with sparse columns, so the parquet files are always dense.
Same for frameworks consuming Arff or CSV files (`AutoWEKA`, `autoxgboost`, `H2OAutoML`, `MLPlan`, `MLNet`, `mlr3automl`) that we can't save to sparse format (`liac_arff` library converts sparse data to dense when saving arff files).

Note that this strategy is now easily configurable, and we can easily decide to change it.

### DIFF
--- a/amlb/utils/core.py
+++ b/amlb/utils/core.py
@@ -9,9 +9,24 @@ import json
 import logging
 import pprint
 import re
+import sys
 import threading
+import types
 
 log = logging.getLogger(__name__)
+
+
+def register_module(module_name):
+    if module_name not in sys.modules:
+        mod = types.ModuleType(module_name)
+        sys.modules[module_name] = mod
+    return sys.modules[module_name]
+
+
+def register_submodule(mod, name):
+    fullname = '.'.join([mod.__name__, name])
+    module = register_module(fullname)
+    setattr(mod, name, module)
 
 
 class Namespace:
@@ -148,7 +163,7 @@ class Namespace:
             self.__dict__.update(other)
         return self
 
-    def __mod__(self, other):
+    def __or__(self, other):
         """extends self with other (adds only missing keys)"""
         if other is not None:
             for k, v in other:
@@ -195,7 +210,7 @@ class Namespace:
         return isinstance(other, Namespace) and self.__dict__ == other.__dict__
 
     def __hash__(self):
-        return hash(self.__dict__)
+        return hash(repr(self))
 
     def __str__(self):
         return Namespace.printer.pformat(Namespace.dict(self))
--- a/amlb/utils/core.py
+++ b/amlb/utils/core.py
@@ -65,7 +65,7 @@ class Namespace:
             if ns is None:
                 continue
             if not deep:
-                merged + ns
+                merged += ns
             else:
                 for k, v in ns:
                     if isinstance(v, Namespace):
@@ -158,12 +158,24 @@ class Namespace:
         self.__dict__.update(dict(*args, **kwargs))
 
     def __add__(self, other):
+        res = Namespace()
+        res += self
+        res += other
+        return res
+
+    def __iadd__(self, other):
         """extends self with other (always overrides)"""
         if other is not None:
             self.__dict__.update(other)
         return self
 
     def __or__(self, other):
+        res = Namespace()
+        res |= self
+        res |= other
+        return res
+
+    def __ior__(self, other):
         """extends self with other (adds only missing keys)"""
         if other is not None:
             for k, v in other:
--- a/frameworks/autosklearn/__init__.py
+++ b/frameworks/autosklearn/__init__.py
@@ -1,6 +1,6 @@
 from amlb.benchmark import TaskConfig
 from amlb.data import Dataset
-from amlb.utils import call_script_in_same_dir
+from amlb.utils import call_script_in_same_dir, unsparsify
 
 
 def setup(*args, **kwargs):
@@ -10,14 +10,16 @@ def setup(*args, **kwargs):
 def run(dataset: Dataset, config: TaskConfig):
     from frameworks.shared.caller import run_in_venv
 
+    X_train, X_test = dataset.train.X_enc, dataset.test.X_enc
+    y_train, y_test = unsparsify(dataset.train.y_enc, dataset.test.y_enc)
     data = dict(
         train=dict(
-            X=dataset.train.X_enc,
-            y=dataset.train.y_enc
+            X=X_train,
+            y=y_train
         ),
         test=dict(
-            X=dataset.test.X_enc,
-            y=dataset.test.y_enc
+            X=X_test,
+            y=y_test
         ),
         predictors_type=['Numerical' if p.is_numerical() else 'Categorical' for p in dataset.predictors]
     )
--- a/frameworks/autosklearn/__init__.py
+++ b/frameworks/autosklearn/__init__.py
@@ -21,12 +21,7 @@ def run(dataset: Dataset, config: TaskConfig):
         ),
         predictors_type=['Numerical' if p.is_numerical() else 'Categorical' for p in dataset.predictors]
     )
-    # options = dict(serialization=dict(
-    #     sparse_matrix_deserialized_format='array'
-    # ))
 
     return run_in_venv(__file__, "exec.py",
-                       input_data=data, dataset=dataset, config=config,
-                       # options=options
-                       )
+                       input_data=data, dataset=dataset, config=config)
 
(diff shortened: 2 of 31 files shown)


frameworks/shared/serialization.py
```diff
@@ -28,16 +32,71 @@ def _import_data_libraries():
     pandas_compression=os.environ.get('AMLB_SER_PD_COMPR') or 'infer',
     # 'snappy', 'gzip', 'brotli', None ?
     pandas_parquet_compression=os.environ.get('AMLB_SER_PD_PQT_COMPR') or None,
+    # if numpy can use pickle to serialize ndarrays,
+    numpy_pickle=True,
+    # if sparse matrices should be compressed during serialization.
+    sparse_matrix_compression=True,
+    # if sparse matrices should be deserialized to some specific format:
+    #  None (no change), 'array' (numpy), 'dense' (dense matrix).
+    sparse_matrix_deserialized_format='array',
```
PGijsbers, line 41: Looks good to me, but if I understand correctly then deserialization in a sparse format is opt-in instead of opt-out (since the default is to deserialize to array)? Maybe we should make it opt-out instead, i.e. by default deserialize in sparse format unless specified otherwise in the `__init__.py`? Because I would argue that the default behavior should be that we can throw anything at the automl frameworks.

frameworks/shared/serialization.py
```diff
@@ -53,36 +112,45 @@ def serialize_data(data, path):
             data.to_hdf(path, os.path.basename(path), mode='w', format='table')
         elif ser == 'json':
             data.to_json(path, compression=ser_config.pandas_compression)
-    elif np:
-        root, ext = os.path.splitext(path)
-        path = f"{root}.npy"
-        np.save(path, data, allow_pickle=True)
     else:
-        raise ImportError("Numpy or Pandas are required to serialize data between processes.")
+        raise ImportError(f"Numpy or Pandas are required to serialize data between processes to {path}.")
```
PGijsbers, line 116: I would suggest making this error message more specific based on data type:

 > Package <numpy/pandas/scipy> is required to serialize data of type {type(data)} to {path}.

If not adding the specific package, then for developers at least the data type will help deduce which package was missing. Otherwise it is not immediately clear [what went wrong](https://github.com/openml/automlbenchmark/pull/371/checks?check_run_id=3186600314#step:7:741).

frameworks/shared/serialization.py
```diff
@@ -28,16 +32,71 @@ def _import_data_libraries():
...
+    else:
+        return data
+
+
 @profile(log)
 def serialize_data(data, path):
-    np, pd = _import_data_libraries()
-    if pd and isinstance(data, (pd.DataFrame, pd.Series)):
+    np, pd, sp = _import_data_libraries()
+    if np and isinstance(data, np.ndarray):
```
PGijsbers, line 89: The added check on arrays makes MLPlan fails which procudes `list` predictions (and probabilities).

PGijsbers (changes requested):
You removed support for `list` predictions/probabilities (previously saved with `np.save`), this breaks support for MLPlan.

frameworks/shared/serialization.py
sebhrusen: I'm still working on this but Python frameworks accepting pandas DataFrame (instead of numpy arrays) also fail with sparse dataframes, so I'd rather make the default to "dense" or "array" for now ("dense" looks actually slightly better).
When we'll have full support for sparse matrices then  we may change the default, but for now it's compromised; for example there are problems with sparse dataframes and parquet format, pandas/arrows fails saving those to parquet format (https://github.com/apache/arrow/issues/1894).

frameworks/shared/serialization.py
sebhrusen: serialization module was never supposed to serialize lists… that's why the methods are named `serialized_data`, accepting only data formats.
Lists go with the blob in json format… I'll probably just fix MLPlan to convert predictions to array.

frameworks/shared/serialization.py
PGijsbers: That looks like a good reason to have the default as you suggested. 👍

frameworks/shared/serialization.py
PGijsbers: Since `result` does not have a signature I wasn't sure whether list format predictions/probabilities were officially supported but assumed they were. If the support was only accidental then changing MLPlan integration is the correct call 👍

frameworks/shared/serialization.py
sebhrusen: well, if I have a object of type `ndarray` then I obviously have `numpy` loaded, same with pandas and so on… so the only use-case is when there's nothing… which should never happen.
anyway, I decided to fallback to `pickle` or `json`, which will be able to serialize only simple data structures without additional dependencies.

sebhrusen:
I'm working on tests for `serialization` and `caller/callee` and it looks like basic frameworks are failing, so you can safely wait before reviewing.

PGijsbers:
Seeing as all checks passed, should I review now?

amlb/utils/core.py
```diff
@@ -148,7 +163,7 @@ def __add__(self, other):
             self.__dict__.update(other)
         return self
 
-    def __mod__(self, other):
+    def __or__(self, other):
```
PGijsbers, line 166: I appreciate this change. Is there a specific reason you did not change it to `__ior__`?

frameworks/autosklearn/__init__.py
```diff
@@ -21,7 +21,12 @@ def run(dataset: Dataset, config: TaskConfig):
...
     )
+    # options = dict(serialization=dict(
+    #     sparse_matrix_deserialized_format='array'
+    # ))
 
     return run_in_venv(__file__, "exec.py",
-                       input_data=data, dataset=dataset, config=config)
+                       input_data=data, dataset=dataset, config=config,
+                       # options=options
+                       )
```
PGijsbers, line 31: Please remove comments

PGijsbers (approved):
Looks good! Two minor remarks (the `__ior__` is just a question/consideration).

sebhrusen:
@PGijsbers I like the `ior` idea, reflects perfectly the current semantic, will change it, thanks.

frameworks/shared/serialization.py
sebhrusen: resolved: serialization now also supports basic datastructures

frameworks/shared/serialization.py
sebhrusen: changed my mind, set the default to `None` (sparse remains sparse) and explicitly set the sparse strategy per frameowkr if they need one.

PGijsbers (approved):
Based on the code changes it looks good. Haven't had time to verify it actually works (though I assume you have).

PGijsbers:
Had a closer look (and a few tests), two more remarks:
 1. According to `liac-arff` documentation, it is possible to [save sparse arff files](https://pythonhosted.org/liac-arff/#dumping-sparse-data). Did you try this?
 2. Could you add a small bit about this in the framework integration documentation?

sebhrusen:
> According to liac-arff documentation, it is possible to save sparse arff files. 

Yes, I tried it on Sat, and serialization works but then some of the frameworks consuming Arff files didn't support the format (H2O being one of them)… which is like creating a new issue.
I'm not sure I want to go down this rabbit hole now (would have to change the file format for H2O for example, Arff has never been a format of choice and it's creating several issues in this integration which makes paradoxically — given that wrote a great part of it — H2O integration one of the less satisfying).

> Could you add a small bit about this in the framework integration documentation?

I'll add some intructions about the serialization options.

PGijsbers:
> … which is like creating a new issue.

Hmm, I'm not sure how I feel about not providing the choice to frameworks if they use ARFF. But at the same time we are planning to move away from ARFF anyway. I guess either way it could be tackled in a separate PR should we choose to support sparse ARFF.

sebhrusen:
> Hmm, I'm not sure how I feel about not providing the choice to frameworks if they use ARFF

But we have the same problem with AG using parquet currently.
As I say, the rabbit hole is right in front of us, do you want to jump in?

PGijsbers:
I suppose not right now.

sebhrusen:
https://github.com/openml/automlbenchmark/issues/375

PR #372, opened 2021-07-28
Disable imputation of missing values for TPOT framework

sebhrusen:
TPOT seems to work fine without imputation: it was probably not the case in the past, but it works fine now, so as it is the only non-trivial framework for which we do imputation, let's remove this.


PGijsbers (approved):
Slightly annoying that TPOT doesn't adhere to time constraints well enough to pass CI. But the change looks fine.

PR #373, opened 2021-07-28
X_NUM_THREADS env variables should not be needed anymore

sebhrusen:
Frameworks like `autogluon`, `mljar`, `mlplan` use sklearn with numpy and scipy on multiple CPUs without having to set the following env variables on their process:
```python
os.environ['OMP_NUM_THREADS'] = '1'
os.environ['OPENBLAS_NUM_THREADS'] = '1'
os.environ['MKL_NUM_THREADS'] = '1'
```

However, if the benchmark explicitly passed the number of cpus to `mlplan`, `autosklearn` and `TPOT` (respectively through the `ncpus`, `n_jobs` (bis) parameters, this is not used by `autogluon` and `mljar`, which could explain why they don't need those env variables. (?)

I suggest to remove those "probably legacy" variables from all framework integrations, after getting approval from the framework's authors.

### DIFF
--- a/frameworks/GAMA/exec.py
+++ b/frameworks/GAMA/exec.py
@@ -3,13 +3,12 @@ import os
 import sys
 import tempfile as tmp
 
+os.environ['JOBLIB_TEMP_FOLDER'] = tmp.gettempdir()
 if sys.platform == 'darwin':
     os.environ['OBJC_DISABLE_INITIALIZE_FORK_SAFETY'] = 'YES'
-os.environ['JOBLIB_TEMP_FOLDER'] = tmp.gettempdir()
-os.environ['OMP_NUM_THREADS'] = '1'
-os.environ['OPENBLAS_NUM_THREADS'] = '1'
-os.environ['MKL_NUM_THREADS'] = '1'
-
+    os.environ['OMP_NUM_THREADS'] = '1'
+    os.environ['OPENBLAS_NUM_THREADS'] = '1'
+    os.environ['MKL_NUM_THREADS'] = '1'
 
 import category_encoders
 from packaging import version
--- a/frameworks/TPOT/exec.py
+++ b/frameworks/TPOT/exec.py
@@ -4,12 +4,12 @@ import pprint
 import sys
 import tempfile as tmp
 
+os.environ['JOBLIB_TEMP_FOLDER'] = tmp.gettempdir()
 if sys.platform == 'darwin':
     os.environ['OBJC_DISABLE_INITIALIZE_FORK_SAFETY'] = 'YES'
-os.environ['JOBLIB_TEMP_FOLDER'] = tmp.gettempdir()
-os.environ['OMP_NUM_THREADS'] = '1'
-os.environ['OPENBLAS_NUM_THREADS'] = '1'
-os.environ['MKL_NUM_THREADS'] = '1'
+    os.environ['OMP_NUM_THREADS'] = '1'
+    os.environ['OPENBLAS_NUM_THREADS'] = '1'
+    os.environ['MKL_NUM_THREADS'] = '1'
 
 from tpot import TPOTClassifier, TPOTRegressor, __version__
 
--- a/frameworks/autosklearn/exec.py
+++ b/frameworks/autosklearn/exec.py
@@ -2,13 +2,16 @@ import logging
 import math
 import os
 import shutil
+import sys
 import tempfile as tmp
 import warnings
 
 os.environ['JOBLIB_TEMP_FOLDER'] = tmp.gettempdir()
-os.environ['OMP_NUM_THREADS'] = '1'
-os.environ['OPENBLAS_NUM_THREADS'] = '1'
-os.environ['MKL_NUM_THREADS'] = '1'
+if sys.platform == 'darwin':
+    os.environ['OMP_NUM_THREADS'] = '1'
+    os.environ['OPENBLAS_NUM_THREADS'] = '1'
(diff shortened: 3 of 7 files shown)


frameworks/GAMA/exec.py
```diff
@@ -3,13 +3,12 @@
...
 if sys.platform == 'darwin':
     os.environ['OBJC_DISABLE_INITIALIZE_FORK_SAFETY'] = 'YES'
-os.environ['JOBLIB_TEMP_FOLDER'] = tmp.gettempdir()
-os.environ['OMP_NUM_THREADS'] = '1'
-os.environ['OPENBLAS_NUM_THREADS'] = '1'
-os.environ['MKL_NUM_THREADS'] = '1'
-
+    os.environ['OMP_NUM_THREADS'] = '1'
+    os.environ['OPENBLAS_NUM_THREADS'] = '1'
+    os.environ['MKL_NUM_THREADS'] = '1'
```
sebhrusen, line 11: @PGijsbers need approval from author

frameworks/autosklearn/exec.py
```diff
@@ -2,13 +2,16 @@
...
 
 os.environ['JOBLIB_TEMP_FOLDER'] = tmp.gettempdir()
-os.environ['OMP_NUM_THREADS'] = '1'
-os.environ['OPENBLAS_NUM_THREADS'] = '1'
-os.environ['MKL_NUM_THREADS'] = '1'
+if sys.platform == 'darwin':
+    os.environ['OMP_NUM_THREADS'] = '1'
+    os.environ['OPENBLAS_NUM_THREADS'] = '1'
+    os.environ['MKL_NUM_THREADS'] = '1'
+
```
sebhrusen, line 14: @mfeurer needs approval from author

frameworks/hyperoptsklearn/exec.py
```diff
@@ -3,12 +3,15 @@
...
 
 os.environ['JOBLIB_TEMP_FOLDER'] = tmp.gettempdir()
-os.environ['OMP_NUM_THREADS'] = '1'
-os.environ['OPENBLAS_NUM_THREADS'] = '1'
-os.environ['MKL_NUM_THREADS'] = '1'
+if sys.platform == 'darwin':
+    os.environ['OMP_NUM_THREADS'] = '1'
+    os.environ['OPENBLAS_NUM_THREADS'] = '1'
+    os.environ['MKL_NUM_THREADS'] = '1'
+
```
sebhrusen, line 14: @PGijsbers hyperoptsklearn is not included in https://github.com/openml/automlbenchmark/pull/235 and I think it's dead, isn't it?

frameworks/TPOT/exec.py
```diff
@@ -4,12 +4,12 @@
...
+os.environ['JOBLIB_TEMP_FOLDER'] = tmp.gettempdir()
 if sys.platform == 'darwin':
     os.environ['OBJC_DISABLE_INITIALIZE_FORK_SAFETY'] = 'YES'
-os.environ['JOBLIB_TEMP_FOLDER'] = tmp.gettempdir()
-os.environ['OMP_NUM_THREADS'] = '1'
-os.environ['OPENBLAS_NUM_THREADS'] = '1'
-os.environ['MKL_NUM_THREADS'] = '1'
+    os.environ['OMP_NUM_THREADS'] = '1'
+    os.environ['OPENBLAS_NUM_THREADS'] = '1'
+    os.environ['MKL_NUM_THREADS'] = '1'
```
sebhrusen, line 12: @weixuanfu, I need approval from author, thanks

mfeurer:
> However, if the benchmark explicitly passed the number of cpus to mlplan, autosklearn and TPOT (respectively through the ncpus, n_jobs (bis) parameters, this is not used by autogluon and mljar, which could explain why they don't need those env variables. (?)

I believe there is a difference between the resource allocation between tools like auto-sklearn on the one side and autogluon on the other. Auto-sklearn runs multiple machine learning pipelines in parallel and therefore needs to restrict each of them to use only a single core, which is what these variables achieve. autogluon runs one machine learning pipeline at a time, so it doesn't need this.

Having said this, auto-sklearn does this by itself starting from version 0.13, so you could instead add an if statement wrt to the version number.

frameworks/hyperoptsklearn/exec.py
PGijsbers: I'm not sure. There doesn't seem to be any active development anymore and it doesn't completely conform to guidelines (e.g. no time-based limit). I'm also not sure how relevant the package still is (it's seems great at what it does, but I haven't seen an AutoML paper use it their comparison in a while).

frameworks/TPOT/exec.py
PGijsbers: Weixuanfu can correct me if I'm wrong, but with [TPOT parallelization](https://github.com/EpistasisLab/tpot/blob/6448bdb71ba08b4a0447c640d2f05a05e1affc21/tpot/base.py#L1545-L1574) the reason for setting this is the same as with auto-sklearn (this parallelization does not control further parallelization of underlying packages which may lead to using a number of cores exceeding the specified limit with `n_jobs`).

frameworks/GAMA/exec.py
PGijsbers: For similar reasons as auto-sklearn, I think to guarantee no accidental violation of cpu constraints would need to be used for GAMA.

sebhrusen:
Given your feedbacks, I'll postpone those changes for after v2. Thank you.

Issue #375, opened 2021-08-02
Add support for sparse data in Arff and parquet formats.


sebhrusen:
Frameworks that mainly support Arff or Parquet format (instead of pandas DF) should be allowed to consume sparse data.

PR #376, opened 2021-08-02
Write splits of sparse data into sparse arff format

sebhrusen:
https://github.com/openml/automlbenchmark/issues/375

Still need to ensure that all frameworks reading arff files support this sparce format:
- H2O : no, might consider consuming parquet files instead.
- AutoWeka: relies on WEKA (?), so it should support sparse data.
- MLPlan: relies on WEKA (?)
- autoxgboost: relies on `farff:readARFF` (see below).
- mlr3automl: relies on `mlr3oml::read_arff`, https://rdrr.io/github/mlr-org/mlr3oml/man/read_arff.html
- ranger (actually just a R impl of RF): relies on `farff::readARFF` that doesn't support sparse Arff files, https://rdrr.io/cran/farff/man/readARFF.html

### DIFF
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -18,7 +18,7 @@ import scipy.sparse as sp
 
 from ..data import AM, DF, Dataset, DatasetType, Datasplit, Feature
 from ..resources import config as rconfig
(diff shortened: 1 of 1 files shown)


PGijsbers:
I think that's reasonable, helps us get towards full sparse format support even if there's currently a mismatch for some frameworks.

PR #381, opened 2021-08-16
Ensure that framework's extra result columns always appear at the end.

sebhrusen:
Some frameworks like AutoGluon expose additional columns to the results file.
Those should always appear after the metrics.


ledell (approved):
Great, thanks!

PR #385, opened 2021-08-30
Fix MLNet on AWS

PGijsbers:
It seems that for some reason not having a `HOME` directory causes an issue with `mlnet` (even if `{DOTNET, MLNET}_CLI_HOME` are configured). This leads to problems in AWS mode where `HOME` is not configured. This patch ensures a `HOME` variable exists to avoid the errors. Additionally it cleans up the `version` registration to `.installed` (it would only match the last character), the `grep` is required because `mlnet --version` produces two different outputs depending on if its the first `mlnet` call or not.


PGijsbers:
@LittleLittleCloud Any idea what's happening or why setting the respective `HOME` variables doesn't fix the issue?

LittleLittleCloud:
@PGijsbers 
What's the error looks like. And is the errror from dotnet install script, or mlnet cli itself.

PGijsbers:
It's from the `mlnet` cli tool. As far as I can tell it happens whenever the tool is used and there is no `HOME` environment variable. Based on the message I assume that some part of the tool assumes that `HOME` is always set, but when it isn't that leads to an unexpected `null` value which crashes the application:

```
[  361.111154] cloud-init[1494]: Running cmd `/repo/frameworks/MLNet/lib/mlnet classification --dataset /s3bucket/input/org/openml/www/datasets/61/dataset_train_0.csv --test-dataset /s3bucket/input/org/openml/www/datasets/61/dataset_test_0.csv --train-time 60 --label-col 4 --output /tmp/tmpssjbguqy --name 0 --verbosity q --log-file-path /tmp/tmpssjbguqy/0/log.txt`
[  361.705643] cloud-init[1494]: Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'path1')
[  361.707696] cloud-init[1494]:    at System.IO.Path.Combine(String path1, String path2)
[  361.709225] cloud-init[1494]:    at Microsoft.ML.CLI.Commands.MLCommand..ctor() in /_/src/mlnet/Commands/MLCommand.cs:line 39
[  361.709358] cloud-init[1494]:    at Microsoft.ML.CLI.Commands.AutoMLCommand..ctor() in /_/src/mlnet/Commands/AutoMLCommand.cs:line 35
[  361.709445] cloud-init[1494]:    at Microsoft.ML.CLI.Commands.ClassificationCommand..ctor() in /_/src/mlnet/Commands/ClassificationCommand.cs:line 25
[  361.709520] cloud-init[1494]:    at Microsoft.ML.CLI.Program.Main(String[] args) in /_/src/mlnet/Program.cs:line 217
[  361.709600] cloud-init[1494]: Aborted (core dumped)
```

Note that this also happens even if `MLNET_CLI_HOME` is set.

LittleLittleCloud:
@PGijsbers 
Got it, I will push a fix for that later today, the new version will be `16.7.4`

PGijsbers:
Great :) a month ago you said:

>  For ML.Net, we are using 16.5.26 to run benchmarks

Could we use `16.7.4` instead? Or should we use `16.5.26`?

LittleLittleCloud:
@PGijsbers 
Sure, let me know if that fixes the error.

PGijsbers:
Thanks, looks like it works. I also managed to remove `MLNET_CLI_HOME` without fail, is that correct or should we keep it in the environment?

LittleLittleCloud:
@PGijsbers 
You can remove `MLNET_CLI_HOME` since it's not used by mlnet.cli

frameworks/MLNet/setup.sh
```diff
@@ -2,32 +2,32 @@
 HERE=$(dirname "$0")
 MLNET='mlnet'
 VERSION=${1:-"latest"}
+
 DOTNET_INSTALL_DIR="$HERE/lib"
 MLNET="$DOTNET_INSTALL_DIR/mlnet"
 DOTNET="$DOTNET_INSTALL_DIR/dotnet"
 SOURCE="https://mlnetcli.blob.core.windows.net/mlnetcli/index.json"
 
+export DOTNET_CLI_HOME="$DOTNET_INSTALL_DIR"
```
sebhrusen, line 11: if you export it, I guess it's used by `dotnet_install.sh`.
But what about the
```export DOTNET_ROOT="$DOTNET_INSTALL_DIR"```
is it needed?

sebhrusen (approved):
minor comment, otherwise LGTM, thanks!

frameworks/MLNet/setup.sh
PGijsbers: Yes, without it MLNet will not work: `A fatal error occurred. The required library libhostfxr.so could not be found.`

PR #388, opened 2021-09-02
AWS: allow user to specify availability zone for ec2 instances

sebhrusen:


PGijsbers (approved):
As far as I can tell this works as intended. And when specifying an invalid zone it's relatively easy to see the cause of the error too.

sebhrusen:
@PGijsbers thanks a lot for testing it!

PR #389, opened 2021-09-02
AMLB failed retrying on Spot forced termination

sebhrusen:
App was expectiong message `Server.SpotInstanceTermination` but was `Server.SpotInstanceTermination: Spot instance termination`
Therefore we need to revert the search string logic.

### DIFF
--- a/amlb/utils/core.py
+++ b/amlb/utils/core.py
@@ -234,8 +234,9 @@ class Namespace:
         return Namespace.dict(self)
 
 
-def repr_def(obj):
-    return "{clazz}({attributes})".format(clazz=type(obj).__name__, attributes=', '.join(("{}={!r}".format(k, v) for k, v in obj.__dict__.items())))
+def repr_def(obj, show_private=False):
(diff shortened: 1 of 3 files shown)


amlb/utils/core.py
```diff
@@ -234,8 +234,9 @@ def __json__(self):
         return Namespace.dict(self)
 
 
-def repr_def(obj):
-    return "{clazz}({attributes})".format(clazz=type(obj).__name__, attributes=', '.join(("{}={!r}".format(k, v) for k, v in obj.__dict__.items())))
+def repr_def(obj, show_private=False):
+    return "{cls}({attrs!r})".format(cls=type(obj).__name__,
+                                     attrs={k: v for k, v in vars(obj).items() if show_private or not k.startswith('_')})
```
PGijsbers, line 239: Not asking you to change it now, but for readability of the code and (in this case in particular) the diffs, using multiple lines of code is nicer.

PGijsbers (approved):

PR #390, opened 2021-09-03
Rename r-packages to lib

PGijsbers:
Since it is the defined folder for framework dependencies, similar to #377


sebhrusen (approved):
thanks, I missed that one

PGijsbers:
Getting an error `there is no package called ‘mlr3learners’` in CI, was not present in my test run on AWS. But it seems unrelated to this PR as it was also present in the [retry logic PR](https://github.com/openml/automlbenchmark/runs/3499152666#step:7:18384). For that reason I'll go ahead and merge this and create a new issue for it.

Issue #321, opened 2021-06-11
Benchmark does not use OpenML cache directory by default


PGijsbers:
`runbenchmark.py` explicitly specifies the input directory, which is used as cache:
```python
default_dirs = ns(
    input_dir="~/.openml/cache",
    ...
)
```
but the default is normally `~/.openml/` ([specifics](https://openml.github.io/openml-python/develop/usage.html#configuration)), but can also be configured by the user (in the openml configuration).
Wouldn't it be better to only use `input_dir` if it is set in the benchmark configuration, and make the default of the benchmark fall back on whatever OpenML decides the openml cache directory is?

Or was this on purpose because we add additional files there (i.e. split files)? If so, wouldn't it be better to have a mirror directory with _just_ those extra files, and not the entire dataset/metadata files as well?

sebhrusen:
Default openml-client cache folder used to be "~/.openml/cache" (at least on Linux/macOS), hence this default.
Btw, the link provided above mentions:
```text
cachedir:
        if not given, will default to ~/.openml/cache
```

I'm not convinced that supporting multiple `input` folders would be useful: it will just add complexity and its share of bugs as `input_dir` is used to share data in `container` mode, and to potentially upload data to ec2 instances.
Actually, I tend to think that defaulting to the `openml` cache folder is not even that useful, and we could as well point default `input_dir` to `~/.config/automlbenchmark/data`

PGijsbers:
> Btw, the link provided above mentions:

Hm, I'll take it up with `openml-python` devs to see what we do here. Thanks for pointing that out 🤦‍♂️ 
But either way I think we could avoid a hard-coded path, since I think the user's openml configuration file should be leading when deciding where the openml cache folder is (imo). The `openml` section in `config.yaml` can then be used to overrule any options from the `.openml/config` file.

> I'm not convinced that supporting multiple input folders would be useful...

I think it makes most sense to have an input directory (for custom datasets) and be able to share the openml cache directory in addition. Perhaps we can load from the openml cache but write any fold-files to the amlb input directory? You shouldn't need to access the openml cache directly, just through openml. The benefit of using the openml cache is mostly for local mode, where a user could first use `openml` to see what tasks/studies they find interesting (already downloading them for that purpose), and then run the benchmark (without redownloading the datasets they just downloaded).

sebhrusen:
> The openml section in config.yaml can then be used to overrule any options from the .openml/config file.

[...]

PR #394, opened 2021-09-09
Option to resume benchmark or rerun missing results based on previous results

sebhrusen:
By using the `--resume` option, the benchmark app will skip the duplicated jobs (ie. the jobs for which there's already an entry in `output_dir/results.csv`).

This is especially useful in AWS mode when there can be failures due to instances unavailability: in this case, an entry is created in `failures.csv` but not in `results.csv`, so re-rerunning the same benchmark later with the `--resume` option will run only the missing jobs.

### DIFF
--- a/resources/config.yaml
+++ b/resources/config.yaml
@@ -12,6 +12,7 @@ root_dir:    # app root dir: set by caller (runbenchmark.py)
 script:     # calling script: set by caller (runbenchmark.py)
 run_mode:   # target run mode (local, docker, aws): set by caller (runbenchmark.py)
 sid:        # session id: set by caller (runbenchmark.py)
+job_history:    # file containing the list of jobs already executed: set by caller (runbenchmark.py)
 
 test_mode: false        # if set to true, some additional checks are executed at runtime.
 seed: auto              # default global seed (used if not set in task definition), can be one of:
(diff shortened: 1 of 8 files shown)


resources/config.yaml
```diff
@@ -61,7 +62,7 @@ benchmarks:                     # configuration namespace for the benchmarks def
     min_vol_size_mb: -1         # default minimum amount of free space required on the volume. If <= 0, skips verification.
 
 job_scheduler:           # configuration namespace
-  stop_on_job_failure:   # if true, the entire run will be aborted on the first job failure (mainly used for testing) : set by caller (runbenchmark.py)
+  exit_on_job_failure:   # if true, the entire run will be aborted on the first job failure (mainly used for testing) : set by caller (runbenchmark.py)
```
sebhrusen, line 65: `stop_on_job_failure` was some kind of typo as the code uses the name above

runbenchmark.py
```diff
@@ -8,13 +8,15 @@
...
+from openml.config import get_cache_directory
+
 import amlb
 from amlb.utils import Namespace as ns, config_load, datetime_iso, str2bool, str_sanitize, zip_path
 from amlb import log, AutoMLError
 
 
 default_dirs = ns(
-    input_dir="~/.openml/cache",
+    input_dir=get_cache_directory(),
```
sebhrusen, line 19: should be good enough to fix `https://github.com/openml/automlbenchmark/issues/321` and I don't see any negative impact as in AWS mode, the input_dir is always overriden.

sebhrusen:
@PGijsbers please have a look at this (will be offline for the next 10 days).
Works locally, still need to test it in AWS mode, but I think it should be a nice addition for v2

runbenchmark.py
PGijsbers: I am a bit confused, that seems to be what I proposed in the original post of the issue. Why are you in favor now?

PGijsbers (approved):
I tested it and it seems to work so far (I simply ran with -f 0 first and later without it and it simply added the remainder).

runbenchmark.py
sebhrusen: was it? I was confused myself when making this change, so I checked the ticket to recall my arguments then, and it looks like I was mainly against having 2 folders: one for input and one for the openml cache…
Maybe there was a mistunderstanding, idk, I've been searching what could go wrong with this fix, and now I don't see anything.

PGijsbers: I think the issue got derailed. The original proposal was exactly this. However I then asked why it wasn't this already, and proposed that perhaps it is due to our own split files, in which case I said maybe a second folder could be an option if that's why it was separated. Then the discussion continued about the second directory as opposed to changing the default directory 😅 

Still, I do think there's something to be said against creating files in the openml cache directory itself, but for me that wouldn't be a strong enough incentive not to share the cache directory.

sebhrusen: Ok, let's do this for now: it should make the majority happy.

PR #404, opened 2021-09-27
Fix/docker tag

### LINKED ISSUE #403: Standardize docker image tags (on `stable`)
With a clean`stable` repository running `python runbenchmark.py FRAMEWORK -m docker -s only` generates docker images with inconsistent tags:
 - most frameworks indicate just the version `VERSION` (e.g. `0.10.7`)
 - mlnet is set to `latest`
 - baselines are set to `stable`
 
On the stable branch we want to generate a unique tag for the image so that we can retrieve the image from docker hub if it's already created. The proposed format is `FRAMEWORK_VERSION-BENCHMARK_VERSION`, e.g. `0.10.7-v2`.

PGijsbers:
Previously the format would be `framework_version` when creating an image on the default branch (normally `stable`). 
In case the local environment would deviate from the online latest version, `-dev` would be appended to the tag (e.g. `0.10.6-dev`).
Now in similar fashion, if everything is clean and latest on the default branch and a version tag happens to be present, it will be added to the tag e.g. `0.10.6-v2.0.1`.

Closes #403

### DIFF
--- a/amlb/runners/container.py
+++ b/amlb/runners/container.py
@@ -182,7 +182,8 @@ Do you still want to build the container image? (y/[n]) """).lower() or 'n'
                 image = self._container_image_name(dev)
 
         if not image:
-            image = self._container_image_name()
+            version_tag = next((t for t in tags if t.startswith("v")), None)
+            image = self._container_image_name(version_tag)
         self._run_container_build_command(image, cache)
         return image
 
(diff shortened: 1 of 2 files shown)


sebhrusen (approved):
Thanks @PGijsbers

PR #405, opened 2021-09-28
[v2.0.1] Fix docker version tags

PGijsbers:
I'm confused as to why some commits which were already in `stable-v2` are marked as new (e.g. `stable-tag`  workflow).

### DIFF
--- a/.github/workflows/stable-tag.yml
+++ b/.github/workflows/stable-tag.yml
@@ -1,0 +1,24 @@
+name:  Tag with Stable
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  tag-if-latest-stable:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Register git credentials
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+      - name: Update stable if latest
+        run: |
+          MOST_RECENT_STABLE=$(git ls-remote --tags origin | tail -n 1 | sed -E 's/.*\s//')
+          echo "New: '""${GITHUB_REF}""', most recent: '""${MOST_RECENT_STABLE}""'"
+          if [ "$GITHUB_REF" == "$MOST_RECENT_STABLE" ]; then
+            git tag stable
+            git push --tags -f
+          fi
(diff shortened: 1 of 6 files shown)


sebhrusen:
Oh, you merged https://github.com/openml/automlbenchmark/pull/404 into `master`?
I would not recommend this in the future: `stable-vX` are branched out from `master` so we should avoid any merge from `master` into `stable-vX` — that's what the current PR is doing — as it makes history very hard to read and there's a risk of bringing unwanted changes into `stable-vX`.
I'd rather apply the fix on `stable-vX` and then merge the latter back into `master`: this keeps the history linear.

For this PR, I'd even rather cherry pick commits and apply them into `stable-v2` than merge `master` into `stable-v2`.

PGijsbers:
> Oh, you merged #404 into master?

Yes, sorry. That's how I understood the workflow. I thought only in case updates needed to be applied on an old `stable-v*` branch would we merge directly into those.

What you say implies that we should work from the `stable-v2`? What then is the function of the master branch? And does that mean any future PR into `stable-v2` should also be tagged with a new version, or will it be ok for merges to happen without a new version tag?

I can restore the `fix/docker-tag` branch and merge it into `stable-v2` but I am not entirely sure how bad that will make future synchronization.

sebhrusen:
> What you say implies that we should work from the stable-v2? What then is the function of the master branch? 

No, usual dev goes into `master`, but fixes that should be in already created stable branch (e.g. `stable-v2`) are implemented in that branch, and then merged back into `master`: to me this is the case with the docker image fix, as it was a fix for `v2`.

Basically, at the end, `master` always contains fixes from `stable-vX`, but the opposite is not true. So we should always merge `stable-vX` into `master`, not the other way round.

> And does that mean any future PR into stable-v2 should also be tagged with a new version, or will it be ok for merges to happen without a new version tag?

once you have a pack of fixes in `stable-v2`, I think we can make a new release `v2.0.1` for example as it's just a version tag and this should also trigger the `stable` tag if I'm correct.

> I can restore the fix/docker-tag branch and merge it into stable-v2 but I am not entirely sure how bad that will make future synchronization.

in worst case, there'll be a conflict when merging `stable-v2` into `master`, but this should be easy to resolve.

PR #412, opened 2021-10-03
Included lightautoml in frameworks_stable

eddiebergman:
Sorry, I'm doing these pull requests small and piecemeal as I diagnose some issues on our benchmarking tool.
My expectation is that any framework can be used with the tags 'stable' or 'latest' but perhaps this is incorrect.

* `frameworks_stable.yaml` didnt include 'lightautoml' but it's supported by it's `setup.sh` so I included it in.
* `frameworks_{stable,latest}.yaml` didn't include MLNet. Looking at the `setup.sh`, I included it in `frameworks_latest.yaml`. I'm unsure what you'd like to do about the 'stable' version. The simplest solution is convert both 'stable' and 'latest' to the empty string as defined in it's `setup.sh` so that it can be used with both tags.

### DIFF
--- a/frameworks/mlr3automl/setup.sh
+++ b/frameworks/mlr3automl/setup.sh
@@ -24,7 +24,7 @@ fi
 
 . $HERE/../shared/setup.sh "$HERE"
 if [[ -x "$(command -v apt-get)" ]]; then
-SUDO apt-get update
+  SUDO apt-get update
   SUDO apt-get install -y software-properties-common dirmngr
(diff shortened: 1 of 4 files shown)


PGijsbers:
Thanks! As you suggested, MLNet's `stable` should probably be converted to the empty string as well. Best I can tell, it looks like they install the latest published version, which we consider stable (equivalent to a PyPI release). 

Piecemeal PRs are preferred, it makes it easier to process quickly (I realize the irony) and they are less likely to be outdated or result in merge conflicts if they are not resolved quickly (say, nearly a year after it's opened 🥲).

PR #413, opened 2021-11-03
Improv/aws meta

PGijsbers:
Adds `volume_type`  and `volume_size` (in gb) to the `meta_info` available of AWS instances.
Currently hardcoded to assume only one volume is used, as we do not allow for multiple volumes in the first place (and simply aggregating does not make sense if the different volumes are of a different volume type).

### DIFF
--- a/amlb/runners/aws.py
+++ b/amlb/runners/aws.py
@@ -456,6 +456,7 @@ class AWSBenchmark(Benchmark):
 
                 if state_code == 16:
                     if inst_desc['meta_info'] is None:
+                        instance_volume = list(instance.volumes.all())[0]
                         meta_info = dict(
                             instance_type=instance.instance_type,
                             launch_time=str(instance.launch_time),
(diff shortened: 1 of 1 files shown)


amlb/runners/aws.py
```diff
@@ -456,6 +456,7 @@ def log_console():
 
                 if state_code == 16:
                     if inst_desc['meta_info'] is None:
+                        instance_volume = list(instance.volumes.all())[0]
```
sebhrusen, line 459: if `instance.volumes.all()` returns an iterator, then the safest way (just in case?) is to use
```python
instance_volume = next(instance.volumes.all(), ns(size=math.nan, volume_type='NA'))
```

sebhrusen (approved):

amlb/runners/aws.py
```diff
@@ -465,6 +466,8 @@ def log_console():
                             private_ip=instance.private_ip_address,
                             availability_zone=instance.placement['AvailabilityZone'],
                             subnet_id=instance.subnet_id,
+                            volume_size_gb=instance_volume.size,
+                            volume_type=instance_volume.volume_type,
```
sebhrusen, line 470: why not directly considering multiple volumes now:
```python
instance_volumes = list(instance.volumes.all())
…
volumes=[{'type': v.type, 'size_gb': v.size} for v in instance_volumes]
```

sebhrusen (commented):
I'd rather have one property for the volume(s), no?

PR #415, opened 2021-11-08
Add constraints for v2 benchmark

PGijsbers:
For ease of reproducibility, we want to include our experimental setup in the constraints file. For our experiments we increase the volume size to 100gb and require gp3 volumes (general purpose SSD).

### DIFF
--- a/resources/constraints.yaml
+++ b/resources/constraints.yaml
@@ -28,16 +28,16 @@ test:
   cores: 8
   min_vol_size_mb: 131072
 
-ssd1h8c:
+1h8c_gp3:
   folds: 10
   max_runtime_seconds: 3600
   cores: 8
   min_vol_size_mb: 100000
-  ec2_volume_type: gp2
+  ec2_volume_type: gp3
 
-ssd4h8c:
+4h8c_gp3:
   folds: 10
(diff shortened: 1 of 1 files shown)


resources/constraints.yaml
```diff
@@ -27,3 +27,17 @@ test:
   max_runtime_seconds: 28800
   cores: 8
   min_vol_size_mb: 131072
+
+ssd1h8c:
```
sebhrusen, line 31: do you really want to put the volume type first as if it were the most important?
I'd rather stick with our current approach: {duration}{cores}{whatever comes next}

PGijsbers: I don't care either way, but `{whatever comes next}` was (as far as I was aware) never specified to be after the other fields (or before, for that matter). I went with a prefix because I figured it was less likely to lead to mistakes for me personally (I'm using to typing `1h8c`, I can't forget about `ssd` if I prefix it whereas muscle memory may lead me to skip it the other way around). Nevertheless I don't mind changing it.

PGijsbers: done

sebhrusen (approved):
thanks, good luck! :)

PR #420, opened 2021-11-18
Accept datasets stored in s3

yinweisu:
Currently amlb benchmark yaml file doesn't accept s3 url. This is useful when users want to benchmark on AWS machines with their own private datasets.
This PR adds the ability to parse s3 url and download the file. Users will need to have AWS credentials setup on the given machine to access the resource if it's not publicly available.

### DIFF
--- a/amlb/datasets/fileutils.py
+++ b/amlb/datasets/fileutils.py
@@ -70,7 +70,8 @@ scheme_handlers = dict(
     http=HttpHandler(),
     https=HttpHandler(),
     s3=S3Handler(),
-    s3a=S3Handler
+    s3a=S3Handler(),
+    s3n=S3Handler(),
 )
 
 SUPPORTED_SCHEMES = list(scheme_handlers.keys())
--- a/amlb/datasets/fileutils.py
+++ b/amlb/datasets/fileutils.py
@@ -22,11 +22,7 @@ def s3_path_to_bucket_prefix(s3_path):
 
 
 def is_s3_url(path):
-    if type(path) != str:
-        return False
-    if (path[:2] == 's3') and ('://' in path[:6]):
-        return True
-    return False
+    return isinstance(path, str) and path.lower().startswith("s3://")
 
 
 def is_valid_url(url):
--- a/amlb/datasets/fileutils.py
+++ b/amlb/datasets/fileutils.py
@@ -11,7 +11,22 @@ from ..utils import touch
 
 log = logging.getLogger(__name__)
 
-SUPPORTED_SCHEMES = ("http", "https")
+SUPPORTED_SCHEMES = ("http", "https", "s3")
+
+
+def s3_path_to_bucket_prefix(s3_path):
+    s3_path_cleaned = s3_path.split('://', 1)[1]
+    bucket, prefix = s3_path_cleaned.split('/', 1)
+
+    return bucket, prefix
+
+
+def is_s3_url(path):
+    if type(path) != str:
+        return False
+    if (path[:2] == 's3') and ('://' in path[:6]):
+        return True
+    return False
 
 
 def is_valid_url(url):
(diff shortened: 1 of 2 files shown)


amlb/datasets/fileutils.py
```diff
@@ -11,7 +11,22 @@
...
+    s3_path_cleaned = s3_path.split('://', 1)[1]
+    bucket, prefix = s3_path_cleaned.split('/', 1)
+
+    return bucket, prefix
+
+
+def is_s3_url(path):
+    if type(path) != str:
+        return False
+    if (path[:2] == 's3') and ('://' in path[:6]):
```
PGijsbers, line 27: What is the one optional char we except to be able to be present in `path[2:6]`? If none, then please consider my suggestion.

amlb/datasets/fileutils.py
```diff
@@ -11,7 +11,22 @@
...
+        return True
+    return False
```
PGijsbers, line 29:
```suggestion
def is_s3_url(path):
    return isinstance(path, str) and path.lower().startswith("s3://")
```

PGijsbers (commented):
Haven't tested it yet, in principle I agree S3 url support is nice. Just had a minor remark and hope @sebhrusen can weigh in on the general implementation.

amlb/datasets/fileutils.py
```diff
@@ -21,21 +32,45 @@ def is_valid_url(url):
...
     if not is_valid_url(url):
         return False
-    head_req = Request(url, method='HEAD')
-    try:
-        with urlopen(head_req) as test:
-            return test.status == 200
-    except URLError as e:
-        log.error(f"Cannot access url %s: %s", url, e)
-        return False
+    if not is_s3_url(url):
```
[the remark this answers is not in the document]
sebhrusen: also, as the logic is now significantly more complex, I'd suggest a more OO approach for both clarity and to make it easier to add support for new schemes.
```python

class FileHandler:
    def exists(self, url): pass
    def download(self, url, dest_path): pass


class HttpHandler(FileHandler):
    def exists(self, url):
        head_req = Request(url, method='HEAD')
        try:
            with urlopen(head_req) as test:
                return test.status == 200
        except URLError as e:
            log.error(f"Cannot access url %s: %s", url, e)
            return False
    
    def download(self, url, dest_path):
        with urlopen(url) as resp, open(dest_path, 'wb') as dest:
            shutil.copyfileobj(resp, dest)


class S3Handler(FileHandler):
    def exists(self, url):
        # your logic
        
    def download(self, url, dest_path):
        # your logic
        
    def _s3_path_to_bucket_prefix(self, path):
        # your logic


scheme_handlers = dict(
    http: HttpHandler(),
    https: HttpHandler(),
    s3: S3Handler(),
)


SUPPORTED_SCHEMES = list(scheme_handlers.keys())


def get_file_handler(url):
    return scheme_handlers[urlparse(url).scheme]
```
then in `file.py`:
```python
        elif is_valid_url(dataset):
            cached_file = os.path.join(self._cache_dir, os.path.basename(dataset))
            if not os.path.exists(cached_file):  # don't download if previously done
                handler = get_file_handler(dataset)
                assert handler.exists(dataset), f"Invalid path/url: {dataset}"
                handler.download(dataset, dest_path=cached_file)
            return self._extract_train_test_paths(cached_file)
```

amlb/datasets/fileutils.py
```diff
@@ -21,21 +32,45 @@ def is_valid_url(url):
...
+        head_req = Request(url, method='HEAD')
+        try:
+            with urlopen(head_req) as test:
+                return test.status == 200
+        except URLError as e:
+            log.error(f"Cannot access url %s: %s", url, e)
+            return False
+    else:
+        import boto3
+        from botocore.errorfactory import ClientError
```
sebhrusen, line 45: feel free to import boto3 in the top of the file, this is an optional requirement in `amlb`.

sebhrusen (changes requested):
thanks for your contribution @yinweisu , making some suggestions to improve the overall logic to handle new schemes.

amlb/datasets/fileutils.py
PGijsbers: I was considering suggesting something similar 👍

Innixma: +1, this will make things much easier to extend to new file formats

amlb/datasets/fileutils.py
Innixma: Wouldn't that cause a crash if it is at top of file and is optional if the user hasn't installed boto3? If it is an optional requirement then we wouldn't want to import it until we are sure we need to use it.

amlb/datasets/fileutils.py
Innixma: The optional character is to support both `s3://` and `s3a://` , where a stands for 'accelerated'. It is a different file loading mechanism, with different performance tradeoffs.

amlb/datasets/fileutils.py
yinweisu: Awesome suggestion! Thanks!

amlb/datasets/fileutils.py
Innixma: @yinweisu I'd recommend undoing this change so we can continue to support s3a. I'm unsure how relevant it is anymore since I believe s3a was integrated into s3 by default now, but probably good to still have it.

Innixma: There are also other formats beyond s3a, including s3n: https://stackoverflow.com/questions/33356041/technically-what-is-the-difference-between-s3n-s3a-and-s3

yinweisu: With the OO approach we are going to use, we might simply have something like this:
`scheme_handlers = dict(
    http: HttpHandler(),
    https: HttpHandler(),
    s3: S3Handler(),
    s3a: S3Handler(),
)`
This way we don't need `is_s3_url()` anymore

Innixma: How do we detect if it is an s3 url in the first place then?

for example, you could have a local path called `'s3a/local/data'`.

yinweisu: The code is using urlparse to extract the scheme. I think for a local path `s3a/local/data`, the scheme would be empty

yinweisu:
@sebhrusen Hi, can you review the changes I've made? Thanks!

amlb/datasets/fileutils.py
sebhrusen: I don't know why I wrote `optional`, I meant the exact opposite… probalby skipped the `not`.

amlb/datasets/fileutils.py
```diff
@@ -4,73 +4,85 @@
...
+        s3_path_cleaned = s3_path.split('://', 1)[1]
+        bucket, prefix = s3_path_cleaned.split('/', 1)
+        return bucket, prefix
+
+
+scheme_handlers = dict(
+    http=HttpHandler(),
+    https=HttpHandler(),
+    s3=S3Handler(),
+    s3a=S3Handler
```
sebhrusen, line 73:
```suggestion
    s3a=S3Handler(),
```

didn't you want to support `s3n` as well?

sebhrusen (commented):
looks good to me, except for the small mistake with the extended `s3` schemes

amlb/datasets/fileutils.py
yinweisu: Yes, you are correct. Thanks for pointing that out!

sebhrusen (approved):
Thanks a lot for this @yinweisu , I'm merging it.

PR #421, opened 2021-11-22
Fix mlr3automl

ja-thomas:
Fixed the bug in mlr3automl, unfortunately I don't have write access and would need for a-hanf to react, so if we need to run now we can just use my fork.

### DIFF
--- a/frameworks/mlr3automl/setup.sh
+++ b/frameworks/mlr3automl/setup.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 HERE=$(dirname "$0")
 VERSION=${1:-"stable"}
-REPO=${2:-"a-hanf/mlr3automl"}
+REPO=${2:-"ja-thomas/mlr3automl"}
 MLR_REPO=${3:-"https://github.com/mlr-org"}
 
 # Version can be specified as 'stable', 'latest', a branch or a commit hash (indicated by starting with '#')


frameworks/mlr3automl/setup.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 HERE=$(dirname "$0")
 VERSION=${1:-"stable"}
-REPO=${2:-"a-hanf/mlr3automl"}
+REPO=${2:-"ja-thomas/mlr3automl"}
```
sebhrusen, line 4: no need to change the default repo, I'd prefer to leave the official repo as the default.
On our side, we can easily change this directly in the framework definition:
```yaml
mlr3automl:
  setup_args: ["master", "ja-thomas/mlr3automl"]  # or use "#commit-hash" instead of "master"
  version: 'ja-thomas-fix'
```
@PGijsbers wdyt?

PGijsbers: I agree. I prefer to give @a-hanf a few days to see if Janek's fixes can be merged with the main mlr3automl repo, but otherwise we will tackle this through the framework definition.

a-hanf: Hi everyone and thanks for looking into this.
I merged Janek's changes to the mlr3automl main branch and tried a local test run of the benchmark. The set-up went through without errors and the code was running, but the results show some errors.
I'll look into this more in the following days, hopefully we can get it resolved soon.
Thanks!

ja-thomas:
fixed in #424

PR #422, opened 2021-11-23
Fix AWS random cancel issue

sebhrusen:
Letting the job runner handle the rescheduling logic to ensure that the job can't be acted upon by its current worker after being rescheduled.

### DIFF
--- a/amlb/job.py
+++ b/amlb/job.py
@@ -355,7 +355,6 @@ class MultiThreadingJobRunner(JobRunner):
                     if job.state is State.rescheduling:
                         self.reschedule(job)
                     else:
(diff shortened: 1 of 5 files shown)


sebhrusen:
> I think that should do the trick (at the very least for the scenario I encountered).

I think it's more than a trick  :) It's effectively preventing any modification/state change on the rescheduled job until it gets back in queue

PGijsbers:
Indeed. "It does the trick" for me does not have the connotation of it being something hacky. I certainly didn't mean to imply it's a half-baked solution, I think it's the proper approach :)

PR #423, opened 2021-11-24
Add support for the OpenML test server

mfeurer:
We (me, @louquinze, @eddiebergman) are currently facing https://github.com/openml/openml-python/issues/1123 which prevents us from benchmarking string datasets with the AutoML benchmark because we cannot upload the datasets to the server. As a remedy, we uploaded them to the testserver (after removing the columns in question) and would now like to use the with the AutoML benchmark.

This PR adds support for the domain `"openmltestserver"` that basically reuses the OpenML logic.

### DIFF
--- a/amlb/benchmarks/openml.py
+++ b/amlb/benchmarks/openml.py
@@ -16,7 +16,7 @@ def is_openml_benchmark(benchmark: str) -> bool:
         supported_types = ['s', 't']
 
         if oml_id.isdecimal():
-            return domain == "openml" and oml_type in supported_types
+            return domain in ("openml", "openmltestserver") and oml_type in supported_types
     return False
 
 
@@ -25,6 +25,10 @@ def load_oml_benchmark(benchmark: str) -> Tuple[str, Optional[str], List[Namespa
     domain, oml_type, oml_id = benchmark.split('/')
     path = None  # benchmark file does not exist on disk
     name = benchmark  # name is later passed as cli input again for containers, it needs to remain parsable
+
+    if domain == "openmltestserver":
+        log.debug("Setting openml server to the test server.")
+        openml.config.server = "https://test.openml.org/api/v1/xml"
 
     if openml.config.retry_policy != "robot":
         log.debug(
@@ -38,7 +42,8 @@ def load_oml_benchmark(benchmark: str) -> Tuple[str, Optional[str], List[Namespa
         data = openml.datasets.get_dataset(t.dataset_id, download_data=False, download_qualities=False)
         tasks = [Namespace(name=str_sanitize(data.name),
                            description=data.description,
-                           openml_task_id=t.id)]
+                           openml_task_id=t.id,
+                           id="{}/t/{}".format(domain, t.id))]
     elif oml_type == 's':
         log.info("Loading openml suite %s.", oml_id)
         suite = openml.study.get_suite(oml_id)
@@ -50,7 +55,8 @@ def load_oml_benchmark(benchmark: str) -> Tuple[str, Optional[str], List[Namespa
         for tid, did in zip(suite.tasks, suite.data):
             tasks.append(Namespace(name=str_sanitize(datasets.loc[did]['name']),
                                    description=f"{openml.config.server.replace('/api/v1/xml', '')}/d/{did}",
-                                   openml_task_id=tid))
+                                   openml_task_id=tid,
+                                   id="{}/t/{}".format(domain, tid)))
     else:
         raise ValueError(f"The oml_type is {oml_type} but must be 's' or 't'")
     return name, path, tasks
--- a/amlb/benchmarks/openml.py
+++ b/amlb/benchmarks/openml.py
@@ -16,7 +16,7 @@ def is_openml_benchmark(benchmark: str) -> bool:
         supported_types = ['s', 't']
 
         if oml_id.isdecimal():
-            return domain in ("openml", "openmltestserver") and oml_type in supported_types
+            return domain in ("openml", "test.openml") and oml_type in supported_types
     return False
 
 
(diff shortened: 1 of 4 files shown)


docs/README.md
```diff
@@ -127,7 +127,8 @@ positional arguments:
                         benchmark description file, or an openml suite or
                         task. OpenML references should be formatted as
                         'openml/s/X' and 'openml/t/Y', for studies and tasks
-                        respectively. Defaults to `test`.
+                        respectively. Use 'test.openml/s/X' for the test
```
PGijsbers, line 130: You changed this to `openmltestserver`, as far as I can tell. I would personally prefer `test.openml` but if that gave issues this works for me (but the docs need to be updated).

sebhrusen (approved):
looks good to me, thanks @mfeurer

docs/README.md
sebhrusen: you mean it should be changed in `openml.py`, right? the doc looks good to me, I also prefer `test.openml` over `openmltestserver` as it's easier to convert into a url:

`test.openml/t/4242` -> `test.openml.org/t/4242`

sebhrusen (commented):
apart from this domain change

docs/README.md
mfeurer: Changed.

runbenchmark.py
```diff
@@ -31,7 +31,8 @@
                     help="The benchmark type to run as defined by default in resources/benchmarks/{benchmark}.yaml,"
                          "\na path to a benchmark description file, or an openml suite or task."
                          "\nOpenML references should be formatted as 'openml/s/X' and 'openml/t/Y',"
-                         "\nfor studies and tasks respectively."
+                         "\nfor studies and tasks respectively. Use 'test.openml/s/X' for the test "
```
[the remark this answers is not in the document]
mfeurer: Done.

mfeurer:
There seems to be an issue with the resource verification. The fact that for the call string `python runbenchmark.py constantpredictor -t iris -f 0 -e` there is the error
```
ERROR:
task definition must contain an ID or the property 'dataset' to create an ID, but task definition is { 'cores': 1,
  'ec2_instance_type': 'm5.large',
  'enabled': False,
  'folds': 1,
  'id': None,
  'max_mem_size_mb': -1,
  'max_runtime_seconds': 600,
  'metric': [ None, 'acc', 'auc', 'logloss', 'f1', None, 'mae', 'mse', 'rmse',
              'rmsle', 'r2'],
  'min_vol_size_mb': -1,
  'name': '__dummy-task',
  'openml_task_id': 0}
```
which contains an `openml_task_id` of zero but no dataset name hints to me that there is an issue that is outside of this PR.

amlb/resources.py
```diff
@@ -210,14 +210,14 @@ def _validate_task(self, task, lenient=False):
 
         conf = 'id'
         if task[conf] is None:
-            task[conf] = ("openml.org/t/{}".format(task.openml_task_id) if task['openml_task_id'] is not None
-                          else "openml.org/d/{}".format(task.openml_dataset_id) if task['openml_dataset_id'] is not None
```
sebhrusen, line 214: @mfeurer the deletion of this part is causing the new error.
This is to support the legacy format that we're using in all `yml` files, especially the `test.yaml` one.
I agree that ideally, this should not be here but in `openml.py` module, but right now, I don't have time to look into this, so you can just restore this to get the test wroking again.

amlb/benchmarks/openml.py
```diff
@@ -38,7 +42,8 @@ def load_oml_benchmark(benchmark: str) -> Tuple[str, Optional[str], List[Namespa
         data = openml.datasets.get_dataset(t.dataset_id, download_data=False, download_qualities=False)
         tasks = [Namespace(name=str_sanitize(data.name),
                            description=data.description,
-                           openml_task_id=t.id)]
+                           openml_task_id=t.id,
+                           id="{}/t/{}".format(domain, t.id))]
```
sebhrusen, line 46:
```suggestion
                           id="{}.org/t/{}".format(domain, t.id))]
```
originally, for openml, we wanted to have the `id` == short url to the task/dataset, that's why in the validation, the task_id `42` gives a unique `id=openml.org/t/42`, making it easy for users to look at the task description,  and also for the `amlb_report` module to load metadata.
So it is actually quite important to keep this id as a proper url.

amlb/benchmarks/openml.py
```diff
@@ -50,7 +55,8 @@ def load_oml_benchmark(benchmark: str) -> Tuple[str, Optional[str], List[Namespa
         for tid, did in zip(suite.tasks, suite.data):
             tasks.append(Namespace(name=str_sanitize(datasets.loc[did]['name']),
                                    description=f"{openml.config.server.replace('/api/v1/xml', '')}/d/{did}",
-                                   openml_task_id=tid))
+                                   openml_task_id=tid,
+                                   id="{}/t/{}".format(domain, tid)))
```
sebhrusen, line 59:
```suggestion
                                   id="{}.org/t/{}".format(domain, tid)))
```

amlb/resources.py
```diff
@@ -210,14 +210,14 @@ def _validate_task(self, task, lenient=False):
...
-                          else ((task.dataset['id'] if isinstance(task.dataset, (dict, Namespace))
-                                 else task.dataset if isinstance(task.dataset, str)
-                                 else None) or task.name) if task['dataset'] is not None
+            task[conf] = (((task.dataset['id'] if isinstance(task.dataset, (dict, Namespace))
+                            else task.dataset if isinstance(task.dataset, str)
+                            else None) or task.name) if task['dataset'] is not None
```
sebhrusen, line 215:
```suggestion
            task[conf] = ("openml.org/t/{}".format(task.openml_task_id) if task['openml_task_id'] is not None
                          else "openml.org/d/{}".format(task.openml_dataset_id) if task['openml_dataset_id'] is not None
                          else ((task.dataset['id'] if isinstance(task.dataset, (dict, Namespace))
                                 else task.dataset if isinstance(task.dataset, str)
                                 else None) or task.name) if task['dataset'] is not None
                          else None)
```

sebhrusen (changes requested):

amlb/benchmarks/openml.py
mfeurer: Sorry, this was plain oversight.

mfeurer:
I reverted the changes to `resources.py` while keeping the (in my eyes) more helpful error message. All test are passing except for the mlr3automl test on iris, not sure what is going on there.

sebhrusen (approved):
thanks @mfeurer, the `mlr3automl` issue is being fixed separately.

PR #424, opened 2021-11-26
fix mlr3automl setup

a-hanf:
Updating to dev versions of bbotk and mlr3hyperband

### DIFF
--- a/frameworks/mlr3automl/setup.sh
+++ b/frameworks/mlr3automl/setup.sh
@@ -41,7 +41,7 @@ fi
 LIB="${HERE}/lib/"
 mkdir "${LIB}"
 
-Rscript -e 'options(install.packages.check.source="no"); install.packages(c("mlr3", "mlr3pipelines", "mlr3misc", "mlr3oml", "mlr3tuning", "paradox"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
+Rscript -e 'options(install.packages.check.source="no"); install.packages(c("mlr3", "mlr3pipelines", "mlr3misc", "mlr3oml", "mlr3tuning", "mlr3learners", "paradox"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
 Rscript -e 'options(install.packages.check.source="no"); install.packages(c("remotes", "checkmate", "R6", "xgboost", "ranger", "LiblineaR", "emoa", "e1071", "glmnet"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
 Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${MLR_REPO}"'/mlr3extralearners", lib="'"${LIB}"'")'
 Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${MLR_REPO}"'/mlr3hyperband", lib="'"${LIB}"'")'
--- a/frameworks/mlr3automl/setup.sh
+++ b/frameworks/mlr3automl/setup.sh
@@ -41,9 +41,11 @@ fi
 LIB="${HERE}/lib/"
 mkdir "${LIB}"
 
-Rscript -e 'options(install.packages.check.source="no"); install.packages(c("mlr3", "mlr3pipelines", "mlr3misc", "mlr3oml", "mlr3hyperband", "mlr3tuning", "paradox"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
+Rscript -e 'options(install.packages.check.source="no"); install.packages(c("mlr3", "mlr3pipelines", "mlr3misc", "mlr3oml", "mlr3tuning", "paradox"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
 Rscript -e 'options(install.packages.check.source="no"); install.packages(c("remotes", "checkmate", "R6", "xgboost", "ranger", "LiblineaR", "emoa", "e1071", "glmnet"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
 Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${MLR_REPO}"'/mlr3extralearners", lib="'"${LIB}"'")'
+Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${MLR_REPO}"'/mlr3hyperband", lib="'"${LIB}"'")'
+Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${MLR_REPO}"'/bbotk", lib="'"${LIB}"'")'
 Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${REPO}"'", ref="'"${VERSION}"'", lib="'"${LIB}"'")'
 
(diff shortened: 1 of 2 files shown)


a-hanf:
@ja-thomas I think this should fix the setup issues. Looks like the following change was incompatible with older versions of mlr3hyperband:
https://github.com/mlr-org/bbotk/pull/157

If you find some time, please test and let me know if it works, I can't test on docker / AWS at the moment

frameworks/mlr3automl/setup.sh
```diff
@@ -41,9 +41,11 @@ fi
 LIB="${HERE}/lib/"
 mkdir "${LIB}"
 
-Rscript -e 'options(install.packages.check.source="no"); install.packages(c("mlr3", "mlr3pipelines", "mlr3misc", "mlr3oml", "mlr3hyperband", "mlr3tuning", "paradox"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
+Rscript -e 'options(install.packages.check.source="no"); install.packages(c("mlr3", "mlr3pipelines", "mlr3misc", "mlr3oml", "mlr3tuning", "mlr3learners", "paradox"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
```
sebhrusen, line 44: some thing I tried in the past and worked well for me was to completely remove those explicits imports and use instead only:
```bash
Rscript -e 'options(install.packages.check.source="no"); install.packages(c("remotes"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${REPO}"'", ref="'"${VERSION}"'", lib="'"${LIB}"'", dependencies=TRUE)'
Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${MLR_REPO}"'/mlr3extralearners", lib="'"${LIB}"'")'
```
I can confirm that it also install automatically `mlr3tuning`, `mlr3learners` and everything needed.

of course, if `mlr3hyperband` and `bbotk` needs to be installed from a custom repo, you can install them at the end using `upgrade='always'` to ensure it overrides the default library.
```bash
Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${MLR_REPO}"'/mlr3extralearners", lib="'"${LIB}"'", upgrade="'"always"'")'
Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${MLR_REPO}"'/mlr3hyperband", lib="'"${LIB}"'", upgrade="'"always"'")'
Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${MLR_REPO}"'/bbotk", lib="'"${LIB}"'", upgrade="'"always"'")'
```
If we're able to get rid of the listing of all dependencies and rely on the resolution provided by the package itself, it should make this setup much more robust. wdyt?

a-hanf: This is a good idea.
mlr3automl currently links to the CRAN version of the dependencies, I'll have to make it so the dev versions are installed from github instead. I'll have a look tonight

a-hanf: Done

PGijsbers:
I rebased this locally to `stable-v2`  but it applies 817e0ee461bc372cd2efc2ce2a630773e89c93fd. Otherwise it seems to work fine. It's actually not an issue though (I think), since for a `v2.0.3` release it's good to revert the version change again so that the [versioning workflow](https://github.com/openml/automlbenchmark/blob/master/.github/workflows/versioning.yml) actually runs correctly. It results in the following diff (I manually removed the extra context lines):
``` 
diff --git a/amlb/__version__.py b/amlb/__version__.py
index 07bcb06..d3c8ad7 100644
--- a/amlb/__version__.py
+++ b/amlb/__version__.py
@@ -1,3 +1,3 @@
-__version__ = "2.0.2"
+__version__ = _dev_version

diff --git a/frameworks/mlr3automl/setup.sh b/frameworks/mlr3automl/setup.sh
index 05fa759..b94bcfa 100755
--- a/frameworks/mlr3automl/setup.sh
+++ b/frameworks/mlr3automl/setup.sh
@@ -41,10 +41,8 @@ fi
-Rscript -e 'options(install.packages.check.source="no"); install.packages(c("mlr3", "mlr3pipelines", "mlr3misc", "mlr3oml", "mlr3hyperband", "mlr3tuning", "paradox"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
-Rscript -e 'options(install.packages.check.source="no"); install.packages(c("remotes", "checkmate", "R6", "xgboost", "ranger", "LiblineaR", "emoa", "e1071", "glmnet"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
-Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${MLR_REPO}"'/mlr3extralearners", lib="'"${LIB}"'")'
-Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${REPO}"'", ref="'"${VERSION}"'", lib="'"${LIB}"'")'
+Rscript -e 'options(install.packages.check.source="no"); install.packages(c("remotes"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
+Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${REPO}"'", lib="'"${LIB}"'", dependencies=TRUE)'
``` 
I pushed it as a [new branch](https://github.com/openml/automlbenchmark/tree/pr424rebase) since I wanted to test if it works (on AWS, I can't do local tests with r frameworks)(test is now running).

a-hanf:
The setup change downloads the dev versions of the mlr3 packages to prevent some incompatibilities in the CRAN versions, but by doing so I guess we hit the limit for downloads from Github.
I can make it so that most of the packages are downloaded from CRAN and only a few from Github, I suppose this would solve the issue. What do you think?

PGijsbers:
That would help. Short term that hopefully works, as long as we can build a docker image without hitting the limits we should be fine (and this might already be the case? but I can't test it). Long term I suppose we can look at a way to set up a PAT, but I don't think it's necessary to dive into that now.

a-hanf:
I switched to the CRAN version of some of the dependencies. Let me know if that helps

PGijsbers:
I managed to do a successful run with `python runbenchmark mlr3automl test test -t iris -f 0 -m aws` (from the [rebased branch](https://github.com/openml/automlbenchmark/tree/pr424rebase)). @sebhrusen could you build a docker image and test it works across the validation datasets? I fear that running the full test through amazon will again encounter rate limits. If all looks good we can merge this and #433 and do a `v2.0.4`  release (if no other required changes pop up).

sebhrusen:
@PGijsbers I tried to build a docker image from your `pr424rebase` branch: the image builds fine but at runtime I get:
```text
Running cmd `Rscript --vanilla -e ".libPaths('/bench/frameworks/mlr3automl/lib'); source('/bench/frameworks/mlr3automl/exec.R'); run('/input/org/openml/www/datasets/61/dataset_train_0.arff', '/input/org/openml/www/datasets/61/dataset_test_0.arff', target.index = 5, 'classification', '/output/predictions/iris/0/predictions.csv', 2, time.budget = 60, meta_results_file='/output/meta_results.csv', seed='288827007', name='iris')"`
Error in library(mlr3extralearners) : 

  there is no package called ‘mlr3extralearners’

Calls: source -> withVisible -> eval -> eval -> library

Execution halted

```

PGijsbers:
That confuses me. How could the install work fine both times, and one time give an error-free environment and the other time lack a dependency?

PGijsbers:
Superseded by #443

PR #425, opened 2021-11-29
Unsparsify target variables for (Tuned)RF

PGijsbers:
Sparse targets are not supported in scikit-learn 0.24.2, and are used with tasks 360932 and 360933 (QSAR) in the benchmark.
Unfortunately I can not test the change since I am unable to generate docker images (a connection issue locally, and the `Dockerfile not found`  error on the server). @sebhrusen can you please test? :(


sebhrusen (approved):
tested and approved (using less ntrees that in our default config, as I just wanted to check that it produces a result).
Thanks @PGijsbers

PR #427, opened 2021-12-01
Openml must be configured when loading the tasks

sebhrusen:
Also use `openml.cache_directory` instead of `openml.get_cache_directory()` to read the default cache folder, as the latter doesn't point to the root of the cache folder.

### DIFF
--- a/amlb/benchmark.py
+++ b/amlb/benchmark.py
@@ -72,6 +72,9 @@ class Benchmark:
             return
 
         self._forward_params = locals()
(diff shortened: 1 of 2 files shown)


PGijsbers (approved):
Thanks for spotting this! I also opened an issue on [openml-python#1124](https://github.com/openml/openml-python/issues/1124), so that we eventually may have a clearer interface 👍 
(but it's clear this should be fixed in `stable-v2` now regardless)

PR #433, opened 2021-12-06
fixes #432 add precision to runtimes in results.csv

### LINKED ISSUE #432: Not enough precision in predict_duration in results.csv
We are [rounding](https://github.com/openml/automlbenchmark/blob/aee089164e8236e581672732c8bdc5ee6393d0ac/amlb/results.py#L172) to the first decimal for `predict_duration` so a lot of the results are getting rounded to 0.

Here, `"{:.1f}".format`, we need .3g instead of .1f.

ledell:
Fixes: https://github.com/openml/automlbenchmark/issues/432

### DIFF
--- a/amlb/results.py
+++ b/amlb/results.py
@@ -169,7 +169,7 @@ class Scoreboard:
         for col in nanable_int_cols:
             df[col] = df[col].astype(np.object).map(int_print).astype(np.str)
         for col in low_precision_float_cols:
-            df[col] = df[col].astype(np.float).map(partial(num_print, "{:.1f}".format)).astype(np.float)
+            df[col] = df[col].astype(np.float).map(partial(num_print, "{:.3g}".format)).astype(np.float)
         for col in high_precision_float_cols:
             df[col] = df[col].map(partial(num_print, "{:.6g}".format)).astype(np.float)
 
--- a/amlb/results.py
+++ b/amlb/results.py
@@ -169,7 +169,8 @@ class Scoreboard:
         for col in nanable_int_cols:
             df[col] = df[col].astype(np.object).map(int_print).astype(np.str)
         for col in low_precision_float_cols:
-            df[col] = df[col].astype(np.float).map(partial(num_print, "{:.3g}".format)).astype(np.float)
+            float_format = lambda f: ("{:.1g}" if f < 1 else "{:.1f}").format(f)
+            df[col] = df[col].astype(np.float).map(partial(num_print, float_format)).astype(np.float)
         for col in high_precision_float_cols:
             df[col] = df[col].map(partial(num_print, "{:.6g}".format)).astype(np.float)
 


PGijsbers (approved):
LGTM.

amlb/results.py
```diff
@@ -169,7 +169,7 @@ def as_printable_data_frame(self, verbosity=3):
         for col in nanable_int_cols:
             df[col] = df[col].astype(np.object).map(int_print).astype(np.str)
         for col in low_precision_float_cols:
-            df[col] = df[col].astype(np.float).map(partial(num_print, "{:.1f}".format)).astype(np.float)
+            df[col] = df[col].astype(np.float).map(partial(num_print, "{:.3g}".format)).astype(np.float)
```
sebhrusen, line 172: hold on, will come up with a better formatting:
```python
> "{:0.3g}".format(12345)
'1.23e+04'
```
which I don't like, I think we lose readability if all numbers are not printed the same way
`g` is great for small numbers but `f` is better for large ones…

sebhrusen (changes requested):

amlb/results.py
sebhrusen: Also, for those duration measurements, I don't think we can guarantee a precision better than 0.1s (and it's already very optimistic given that some frameworks get the data loaded before the timer, and others during…).
I think the main goal here is just to avoid having `0` printed for very low values, like some prediction times on some very small datasets (for example with `flaml`).

PGijsbers: I think when visualizing the results we have to separate packages which load data internally and those which get passed a pandas object.

sebhrusen:
```suggestion
            float_format = lambda f: ("{:.1g}" if f < 1 else "{:.1f}").format(f)
            df[col] = df[col].astype(np.float).map(partial(num_print, float_format)).astype(np.float)
```

sebhrusen: @PGijsbers it's not only a question of data loading: as soon as we provide a measure, users will blindly do some ranking on it, and  due to the inherent imprecision in time measurement for such a complex feature, if for example you have 2 predict durations with `20.311` and `20.317`, I don't think it's fair to rank one higher than the other.
I don't see any way to completely prevent this, but I think we should at least mitigate those readings by trying to offer meaningful numbers.

PGijsbers: But that's also why `g` might be good, since it would be `20.3` if only 3 significant digits are shown both times. That said, I understand that there's also a minimum measurable duration, and I don't really know what that is in our application. Normally measuring with milliseconds is OK (I think?), though I'm not 100% sure if that works well enough here (and such short tests are normally repeated, whereas here it's a one-off).

sebhrusen: As said before, we're mainly concerned with not displaying `0.0` for small prediction times which occur *very rarely*: (too) small datasets + simple models (e.g. flaml), so for the great majority of durations measured `f` is perfectly fine as it renders *all* digits with the requested precision (0.1). If you use `g` everywhere then by limiting the number of significant digits, the precision also varies depending on how large the number is: this is fine for a metric like `mse` but for a duration, the precision is fixed, it just depends on how the timer is applied. 

On top of this, `g` will use scientific notation for most durations, which doesn't make it friendly when looking directly at the results in the console or in the csv file. I do that constantly, and for me `3600` = 1h instantly, but `3.6e+03`…takes about a second more, not even mentioning `1.44e+04`…

Now the question of precision. Sure, the timer is accurate at least to the millisecond, but what we're timing is not.
> though I'm not 100% sure if that works well enough here (and such short tests are normally repeated, whereas here it's a one-off)

Exactly, microbenchmarks for example rely on a large amount of repetitions to attain this level of accuracy.
In our case, on top of this, the way the timer is applied is not fully consistent (check the framework integrations): some frameworks systematically provide the probabilities, some don't. For sklearn-based APIs, I was at first timing only the call to `predict`, but some frameworks changed this to measure `predict_proba`, a framework like `MLPlan` returns its own prediction time whereas at the opposite for `MLNet` we measure the lifetime of an entire subprocess… that's why I'm saying that providing a `.1` precision is already generous, and I'm not comfortable pretending we can provide millis.

However, accuracy is improved by considering prediction per row when analyzing the results: the larger the dataset, the more accurate the `prediction_duration/nrows`, and it allows us to neglect some of those integration differences mentioned above. If you have a precision of .1s for the entire prediction duration, then for 100 rows, you already have a precision around millis.

With my suggestion above, all numbers are rendered "plain" with the same precision, except for the very small ones for which we're just avoiding the `0.0`:
```python
In [1]: float_format = lambda f: ("{:.1g}" if f < 1 else "{:.1f}").format(f)

In [2]: float_format(3634.12345)
Out[2]: '3634.1'

In [3]: float_format(14489.3456)
Out[3]: '14489.3'

In [4]: float_format(123.987)
Out[4]: '124.0'

In [5]: float_format(0.5678)
Out[5]: '0.6'

In [6]: float_format(0.00345)
Out[6]: '0.003'
```

Finally right now, when we look at prediction time, `.1` already allows us to clearly separate frameworks, so I don't see the need to add more digits that we can't justify clearly.

ledell:
@PGijsbers @sebhrusen I committed Seb's suggested change.  Then I changed the base branch from `master` to `stable-v2`.  I tried to rebase my branch on top of `stable-v2`, but I am having conflicts related to the version number in `amlb/__version__.py`.  I manually fixed it (my local was saying 2.0, so i changed to 2.0.3).  However, when i `git rebase --continue` it's applying the "Update version to 2.0" commit, so that creates a conflict again and it seems like I am stuck in a loop.  Could one of you please give it a try to see if you can get the rebase working?  `git merge` works ok but I think we want to rebase, not merge.  I think I am making a mistake here...


```
me ~/code/github-shared/backup_automlbenchmark $ open amlb/__version__.py
me ~/code/github-shared/backup_automlbenchmark $ git add amlb/__version__.py
me ~/code/github-shared/backup_automlbenchmark $ git rebase --continue
Applying: Update version to 2.0
No changes - did you forget to use 'git add'?
If there is nothing left to stage, chances are that something else
already introduced the same changes; you might want to skip this patch.

Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".
```

sebhrusen:
@ledell this is not the right way to rebase: when you create your branch on top of branch A (in your case `issue/432` was created on top of `master` and then rebase on top of branch B (in your case `stable-v2`), you're not only bringing your own commits on top of B, but all the commits of B on top of A followed by your own commits…

what you should have done is `git rebase --onto stable-v2 master` : this would add on top of `stable-v2` all the commits you previously added on top of `master`.

now to fix this, maybe you can try `git rebase --onto stable-v2 HEAD~2`: it should rebase on top of `stable-v2` only your last 2 commits, ignoring the rest

sebhrusen:
@ledell FYI, I did the rebase as mentioned above so that I can approve this.

sebhrusen (approved):

PR #437, opened 2021-12-07
Failed to build docker images from linux

sebhrusen:
@PGijsbers I guess it was the same issue you were getting on Windows, right?
```
cannot locate specified Dockerfile: frameworks/constantpredictor/.setup/Dockerfile
```
after cleaning up the `.dockerignore` it seems to work.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,13 +8,10 @@ results
 **/models
 **/predictions
 **/scores
-**/.setup
-**/.installed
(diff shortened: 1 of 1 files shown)


PGijsbers (approved):
Yes, it was. Strangely enough the problem disappeared for me. But I can confirm that it still works with this PR for me, and either way it looks like an improvement :)

PR #439, opened 2021-12-11
Iteratively build the forest to honor constraints

PGijsbers:
In particular depending on the dataset size either memory or time constraints can become a problem which makes it unreliable as a baseline. Gradually growing the forest sidesteps both issues.

Tested on `test` and `Higgs`, the latter which times out on all folds in `v2.0.3` but returned a result every time now (with a forest of 1490 to 1500 trees). 
The only thing I am not sure about is whether or not to set a cap on the number of trees. Small tasks, like blood-transfusion-service-center, can grow tens of thousands of trees and may incur longer inference time (I tested on one fold and it did not really seem a concern there, but I don't know if that generalizes to other tasks/larger time budgets).
That said, while the inference time may be long it should not be so significant as to exceed any limits.

### DIFF
--- a/frameworks/RandomForest/exec.py
+++ b/frameworks/RandomForest/exec.py
@@ -28,7 +28,7 @@ def run(dataset, config):
     log.info(f"\n**** Random Forest [sklearn v{sklearn.__version__}] ****\n")
 
     is_classification = config.type == 'classification'
-    this_process = psutil.Process(os.getpid())  # Does it cover memory usage of `n_jobs`?
+    this_process = psutil.Process(os.getpid())
 
     encode = config.framework_params.get('_encode', True)
     X_train, X_test = dataset.train.X, dataset.test.X
@@ -39,7 +39,7 @@ def run(dataset, config):
         if not (k.startswith('_') or k == "n_estimators")
     }
     n_jobs = config.framework_params.get('_n_jobs', config.cores)  # useful to disable multicore, regardless of the dataset config
-    step_size = config.framework_params.get('step_size', 10)
+    step_size = config.framework_params.get('_step_size', 10)
     final_forest_size = config.framework_params.get('n_estimators', 2000)
 
     # Default margins are conservative, because robustness is paramount for a baseline.
@@ -69,7 +69,7 @@ def run(dataset, config):
 
             will_run_out_of_memory = extrapolate_with_worst_case(memory_usage) >= config.max_mem_size_mb * memory_margin
             will_run_out_of_time = extrapolate_with_worst_case(training_times) >= config.max_runtime_seconds * time_margin
-            if rf.n_estimators == final_forest_size:
+            if rf.n_estimators >= final_forest_size:
                 log.info("Stop training because desired forest size has been reached.")
                 break
             elif will_run_out_of_time:
--- a/frameworks/RandomForest/exec.py
+++ b/frameworks/RandomForest/exec.py
@@ -1,12 +1,14 @@
 import logging
 import os
 import tempfile as tmp
+from typing import List
 
 os.environ['JOBLIB_TEMP_FOLDER'] = tmp.gettempdir()
 os.environ['OMP_NUM_THREADS'] = '1'
 os.environ['OPENBLAS_NUM_THREADS'] = '1'
 os.environ['MKL_NUM_THREADS'] = '1'
 
+import psutil
 import sklearn
 from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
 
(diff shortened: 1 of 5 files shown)


mfeurer:
> The only thing I am not sure about is whether or not to set a cap on the number of trees. 

A brief suggestion on this: if you leave the argument n_estimators as it is (2000) and hard-code the increasing steps (10) you would have an upper bound and also wouldn't have to touch 4 benchmark definition files.

PGijsbers:
Thanks for the suggestion, I'm in favor.

PGijsbers:
Updated. It's still possible to get the old behavior (by setting `_step_size=n_estimators=2000`), but most of the script is now parametrized.

frameworks/RandomForest/exec.py
```diff
@@ -16,29 +18,69 @@
 log = logging.getLogger(os.path.basename(__file__))
 
 
+def extrapolate_with_worst_case(values: List[float], n: int = 5) -> float:
```
joaquinvanschoren, line 21: The mean slope would probably give a better estimate, but the maximum also sounds ok. It's not quite the worst case but possibly close enough.

frameworks/RandomForest/exec.py
```diff
@@ -16,29 +18,69 @@
...
 
     log.info("Running RandomForest with a maximum time of {}s on {} cores.".format(config.max_runtime_seconds, n_jobs))
-    log.warning("We completely ignore the requirement to stay within the time limit.")
     log.warning("We completely ignore the advice to optimize towards metric: {}.".format(config.metric))
 
     estimator = RandomForestClassifier if is_classification else RandomForestRegressor
     rf = estimator(n_jobs=n_jobs,
                    random_state=config.seed,
+                   n_estimators=step_size,
+                   warm_start=True,
```
joaquinvanschoren, line 56: If you warm-start with the same random seed, won't you just add the same trees over and over again? From looking at the code, it seems you would add the same 10 trees over and over again to the ensemble?

joaquinvanschoren (approved):
Looks ok, but please see the note on warm-starting and random seeds.

frameworks/RandomForest/exec.py
PGijsbers: There is not much penalty for stopping a bit too early before the end of the budget, while exceeding the budget can result, in the case of memory, a failed run. For this reason I decided to use max instead of mean (which I actually started with originally). It's true that it's not the real worst case, but I feel it's a simple approximation.

frameworks/RandomForest/exec.py
PGijsbers: Sharp eye. I actually had to test this myself since I could not find clarification on this in the documentation:
```
from sklearn.datasets import make_classification
from sklearn.ensemble import RandomForestClassifier

x, y = make_classification(random_state=0)
rf = RandomForestClassifier(n_estimators=1, warm_start=True, random_state=0)

rf.fit(x, y)
rf.n_estimators += 1
rf.fit(x, y)

print(rf.estimators_)
```
> [DecisionTreeClassifier(max_features='auto', random_state=209652396), DecisionTreeClassifier(max_features='auto', random_state=398764591)]

Repeating this script more than once gives identical output (which in turn is identical to just fitting the forest with `n_estimators=2` in a single fit). You can further look at the trees to see that they are indeed distinct:
```
from sklearn.tree import export_text
for tree in rf.estimators_:
    print(export_text(tree))
```
So it seems that the usage here is correct and does not require any modification.

joaquinvanschoren: OK, so warm-starting overrules the random state? 
Or does it internally start another RNG using random_state=0? You you get seeds 209652396 and 398764591 if you run it again?
That is probably good but something that should be documented :).

PGijsbers: A `random_state` results in the same forest (with unique trees) whether or not `warm_start` is set, even when `warm_start=True` and the forest is created through multiple `fit` calls. This of course contradicts that every `fit` call will yield identical results.

ledell (approved):
I didn't try to run the code, but reading it, it makes sense to me and I think this is a good solution to the out of memory problem.

frameworks/RandomForest/exec.py
```diff
@@ -16,29 +18,69 @@
...
+    this_process = psutil.Process(os.getpid())
 
     encode = config.framework_params.get('_encode', True)
     X_train, X_test = dataset.train.X, dataset.test.X
     y_train, y_test = dataset.train.y, dataset.test.y
 
-    training_params = {k: v for k, v in config.framework_params.items() if not k.startswith('_')}
+    training_params = {
+        k: v for k, v in config.framework_params.items()
+        if not (k.startswith('_') or k == "n_estimators")
```
sebhrusen, line 39: why not using `_n_estimators` instead? or `_max_n_estimators`?
This looks very confusing to me, because non-prefixed params are always pass-through parameters, so when we look at the `RandomForest` definition now, it looks like we're always using `n_estimators`, when it's not actually the case.

sebhrusen (commented):
NB: I prepared a review before my December holidays, but unfortunately forgot to submit it.

PR #441, opened 2021-12-15
Iterative fit for TunedRandomForest to meet memory and time constraints

PGijsbers:
Specifically for each value of `max_features` to try, an equal time budget is allotted, with one additional budget being reserved for the final fit. This does mean that different `max_features` can lead to different number of trees, but it keeps it simple. Tested from docker (`test test`) and aws (`openml/s/269 -t Airlines_DepDelay_10M`).

I realize it's very similar to the RandomForest at this point, the RandomForest baseline is just a special case were no values for `max_features`  are tried. If that's alright, I would like to move forward as is for now. In the future it can be refactored so that `RandomForest` and `TunedRandomForest` use the same code but with a different configuration (unless there's a good reason to keep them separate).

### DIFF
--- a/frameworks/TunedRandomForest/exec.py
+++ b/frameworks/TunedRandomForest/exec.py
@@ -17,12 +17,13 @@ os.environ['OPENBLAS_NUM_THREADS'] = '1'
 os.environ['MKL_NUM_THREADS'] = '1'
 
 import psutil
+import pandas as pd
 import sklearn
 from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
-from sklearn.model_selection import cross_val_score
 
 from frameworks.shared.callee import call_run, result
 from frameworks.shared.utils import Timer
+from custom_validate import cross_validate
 
 log = logging.getLogger(__name__)
 
@@ -105,29 +106,32 @@ def run(dataset, config):
             value = max_features_values.pop(0)
             log.info(f"Evaluating max_features={value} in {time_per_value} seconds.")
 
-            random_forest = estimator(
-                n_jobs=n_jobs,
-                random_state=config.seed,
-                max_features=value,
-                n_estimators=step_size,
-                warm_start=True,
-                **training_params
-            )
+            random_forests = [
+                estimator(
+                    n_jobs=n_jobs,
+                    random_state=config.seed,
+                    max_features=value,
+                    n_estimators=step_size,
+                    warm_start=True,
+                    **training_params
+                ) for _ in range(k_folds)
+            ]
 
             training_times = [training.duration]
             memory_usage = [this_process.memory_info()[0] / (2 ** 20)]
 
             while True:
                 try:
-                    scores = cross_val_score(
-                        estimator=random_forest,
+                    cv_result = cross_validate(
+                        estimators=random_forests,
                         X=X_train,
                         y=y_train,
                         scoring=metric,
                         error_score='raise',
                         cv=k_folds,
+                        return_estimator=True,
                     )
-                    max_feature_scores[value].append(statistics.mean(scores))
+                    max_feature_scores[value].append(statistics.mean(cv_result["test_score"]))
                 except Exception as e:
                     log.error("Failed CV scoring for max_features=%s :\n%s", value, e)
                     log.debug("Exception:", exc_info=True)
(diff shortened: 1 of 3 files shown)


PGijsbers:
I'm adding a fix which allows for stopping tuning if time exceeds and not all `max_features` values are tried. Currently having issues with building the docker image due to unrelated matter. No need to review until I leave a new ping, apologies.

PGijsbers:
There are still a lot of `KeyErrors` but this seems to be a [known issue that does not affect performance](https://github.com/joblib/joblib/issues/1243).

frameworks/TunedRandomForest/exec.py
```diff
@@ -33,14 +35,28 @@ def pick_values_uniform(start: int, end: int, length: int):
...
-    tuning_params = config.framework_params.get('_tuning', training_params)
+    training_params = {
+        k: v for k, v in config.framework_params.items()
+        if not (k.startswith('_') or k == "n_estimators")
+    }
+    if "max_features" in training_params:
+        raise ValueError("`max_features` may not be specified for Tuned Random Forest.")
+
     n_jobs = config.framework_params.get('_n_jobs', config.cores)  # useful to disable multicore, regardless of the dataset config
+    step_size = config.framework_params.get('_step_size', 10)
```
joaquinvanschoren, line 57: I assume that the '10' here is a fallback value, in case the step size is not defined in the config?

frameworks/TunedRandomForest/exec.py
```diff
@@ -64,60 +80,112 @@ def run(dataset, config):
...
-                                             y=y_train,
-                                             scoring=metric,
-                                             error_score='raise',
-                                             cv=5)
-                    max_feature_scores.append((statistics.mean(scores), max_features_value))
-                except stopit.utils.TimeoutException as toe:
-                    log.error("Failed CV scoring for max_features=%s : Timeout", max_features_value)
-                    tuning_durations.append((max_features_value, cv_scoring.duration))
-                    raise toe
+                    scores = cross_val_score(
```
joaquinvanschoren, line 123: What's the behavior here? Will cross_val_score simply call random_forest.fit(fold_X, fold_y)? Because in that case, because of warm-starting, will it train fold 2 on the ensemble pre-trained on fold 1? Or is it smart enough to build 5 fresh models? If so, and you run it in a loop (as done here), will it correctly warm-start the 5 individual models? Or start fresh again?

frameworks/TunedRandomForest/exec.py
```diff
@@ -64,60 +80,112 @@ def run(dataset, config):
...
+            tuning_durations[value] = training_times
+            memory_usage_by[value] = memory_usage
+            last_initial_fit_time = training_times[1] - training_times[0]
+
+        # TODO: Transform the dictionaries into a pandas dataframe (value, trees, score, duration, total_duration, memory, d_memory)
+        log.info("Tuning scores:\n%s", max_feature_scores)
+        log.info("Tuning durations:\n%s", tuning_durations)
+        log.info("Tuning memory:\n%s", memory_usage_by)
+
+        _, best_value = max((scores[-1], value) for value, scores in max_feature_scores.items())
```
joaquinvanschoren, line 162: Add some documentation string that this retrains the best model on the entire training set?

frameworks/TunedRandomForest/exec.py
```diff
@@ -64,60 +80,112 @@ def run(dataset, config):
...
+        log.info("Training final model with `max_features=%s`.", best_value)
+        rf = estimator(n_jobs=n_jobs,
+                       random_state=config.seed,
+                       max_features=best_value,
+                       warm_start=True,
```
joaquinvanschoren, line 167: Same question about warm starting with fixed random state.

joaquinvanschoren (approved):

frameworks/TunedRandomForest/exec.py
PGijsbers: It is. And for now we will keep the framework definitions identical from the previous iteration, meaning that this default value will be the one used in tests. I did some small scale testing and it seemed that fitting `10` or `100` trees at once did not make a meaningful difference to the end result, so I decided to keep it small (one could argue that maybe it should be `1` instead but even `Higgs` fit hundreds of trees).

frameworks/TunedRandomForest/exec.py
PGijsbers: > Or is it smart enough to build 5 fresh models? 

[Yes](https://github.com/scikit-learn/scikit-learn/blob/0d378913be6d7e485b792ea36e9268be31ed52d0/sklearn/model_selection/_validation.py#L269), but (as you point out) it should not smart enough to use warm-start in a loop (for the same reason). I will write something to verify this. I'm not actually sure that there's a way to achieve this within scikit-learn `cross_validate`-like functions.

PGijsbers: It seems that `warm_start`  and `cross_validate` are indeed [incompatible](https://github.com/scikit-learn/scikit-learn/discussions/22042#discussioncomment-1852489) and forests are trained from scratch each time. Will have to change the code to instead perform our own cross validation. I opened [an issue](https://github.com/scikit-learn/scikit-learn/issues/22044) hoping that functionality will be added in the future.

PGijsbers:
After a somewhat painful process of delving into `cross_validate` **and** `_forest` code, I have a patch which seems to work as expected: provide an estimator per fold, which can use warm-starting. There are currently two open issues with scikit-learn ([#22087](https://github.com/scikit-learn/scikit-learn/issues/22087) and [#22044](https://github.com/scikit-learn/scikit-learn/issues/22044)), which should both be resolved in some manner before taking out this adaptation.

I tested it locally through docker (on `test` and `APSFailure` with `test` constraints) and on aws (on `test` and `validation` on `test` constraints and `KDDCup99` on `1h8c_gp3` constraints). Unfortunately `KDDCup99` still fails, this seems to stem from a memory error (which I don't know if it is avoidable without introducing additional complexity like subsampling for large datasets), and [an old known scikit-learn error](https://github.com/scikit-learn/scikit-learn/issues/11777). When the last error arises we could simply forgo tuning (or use results that managed to be computed) and fit only the final model (with default). Since the latter is (roughly) equivalent to imputing with `randomforest` results I think it might be a suitable substitute where required.

frameworks/TunedRandomForest/exec.py
PGijsbers: I added an additional module which contains modified scikit-learn code which now enables using `warm_start` together with `cross_validate` and updated the `tunedrandomforest` code accordingly.

PR #442, opened 2021-12-17
Respect TMP, TMPDIR, TEMP

eddiebergman:
Ello,
CC: @mfeurer 

This PR copies over the environment variables `TMP`, `TMPDIR` and `TEMP` into the environment in which a framework is run. This allows users to specify where automlbenchmark and any underlying framework should place there tmpfiles.

**context**
As part of our automated benchmarking with auto-sklearn, we were filling up the `/tmp` directory of some of our nodes. This would also cause larger runs to eventually crash due to memory issues, as well as causing issues for other users of our cluster.

Debugging why it was not going into the unlimited `TMPDIR=/tmp/slurm_xxxx` folder provided by our slurm cluster, we found that `automlbenchmark` would remove any env variables that were set before running a framework.

We can not pass an argument to `autosklearn` to delete the temporary directory upon finishing as we require `automlbenchmark`'s functionality to zip up the artifacts. Hence, we rely on slurm's own cleanup of `/tmp/slurm_xxxx` once the job has finished and all artifacts zipped up.

As far as I could find, there is no way to pass env variables through automlbenchmark to the framework that is being run.

* [[link](https://github.com/openml/automlbenchmark/blob/master/frameworks/shared/caller.py#L83)] Option to add env variables through to `run_in_venv` which runs the framwork in an environment.
* [[link](https://github.com/openml/automlbenchmark/blob/master/frameworks/autosklearn/__init__.py#L27)] Autosklearn doesn't pass any options to `run_in_venv`, some other frameworks do
* We also run other frameworks to compare autosklearn so modifying just `autosklearn`'s `__init__.py` in the point above would not be sufficient.

### DIFF
--- a/frameworks/shared/caller.py
+++ b/frameworks/shared/caller.py
@@ -3,7 +3,7 @@ import logging
 import os
 import re
 from tempfile import TemporaryDirectory, mktemp
-from typing import Optional, Union
+from typing import List, Optional, Union
 
 import numpy as np
 
@@ -83,7 +83,7 @@ def run_in_venv(caller_file, script_file: str, *args,
                 options: Union[None, dict, ns] = None,
                 process_results=None,
                 python_exec=None,
-                retain_tmp_env: bool = True):
+                retained_env_vars: Optional[List[str]] = ['TMP', 'TEMP', 'TMPDIR']):
     here = dir_of(caller_file)
     if python_exec is None:  # use local virtual env by default
         python_exec = venv_python_exec(here)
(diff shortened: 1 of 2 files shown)


frameworks/shared/caller.py
```diff
@@ -82,7 +82,8 @@ def run_in_venv(caller_file, script_file: str, *args,
                 input_data: Union[dict, ns], dataset: Dataset, config: TaskConfig,
                 options: Union[None, dict, ns] = None,
                 process_results=None,
-                python_exec=None):
+                python_exec=None,
+                retain_tmp_env: bool = True):
```
sebhrusen, line 86: I like the idea, especially for TMP properties, but I'd suggest to make it more generic then in case frameworks want to propagate some other variables:
```suggestion
                retained_env_vars: List[str] = ['TMP', 'TEMP', 'TMPDIR']):
```

frameworks/shared/caller.py
```diff
@@ -93,6 +94,12 @@ def run_in_venv(caller_file, script_file: str, *args,
     ser_config = options['serialization']
     env = options['env'] or ns()
 
+    # To honour any arguments about tmpdir, we copy over
+    # the TMP, TEMP and TMPDIR environment variables
+    if retain_tmp_env:
+        for env_var in ["TMP", "TEMP", "TMPDIR"]:
+            env[env_var] = os.environ.get(env_var)
```
sebhrusen, line 101: adapt according to suggestion above.

Also, having an env variable with `None` value is not exactly the same as having no env variable, so I'd rather suggest to use:
```python
if retained_env_vars:
    for env_var in retained_env_vars:
        env_val = os.environ.get(env_var)
        if env_val is not None:
            env[env_var] = env_val
```

sebhrusen (commented):
@eddiebergman Thanks for the great suggestion, let's use this occasion to make the whole framework env easier to configure.

eddiebergman:
Made the suggested changes :)

sebhrusen (approved):
thanks @eddiebergman , merging this.

PR #443, opened 2021-12-21
fix for mlr3automl

Coorsaa:
makes PR #424 obsolete

### DIFF
--- a/frameworks/mlr3automl/setup.sh
+++ b/frameworks/mlr3automl/setup.sh
@@ -41,10 +41,11 @@ fi
 LIB="${HERE}/lib/"
 mkdir "${LIB}"
 
-Rscript -e 'options(install.packages.check.source="no"); install.packages(c("mlr3", "mlr3pipelines", "mlr3misc", "mlr3oml", "mlr3hyperband", "mlr3tuning", "paradox"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
-Rscript -e 'options(install.packages.check.source="no"); install.packages(c("remotes", "checkmate", "R6", "xgboost", "ranger", "LiblineaR", "emoa", "e1071", "glmnet"), repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
-Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${MLR_REPO}"'/mlr3extralearners", lib="'"${LIB}"'")'
-Rscript -e '.libPaths("'"${LIB}"'"); remotes::install_github("'"${REPO}"'", ref="'"${VERSION}"'", lib="'"${LIB}"'")'
+Rscript -e 'options(install.packages.check.source="no"); install.packages("remotes", repos="https://cloud.r-project.org/", lib="'"${LIB}"'")'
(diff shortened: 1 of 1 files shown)


PGijsbers:
Thanks! I can confirm it builds. It passes all validation tests except `census-income`, where it gives the following error:
```
[DEBUG] [amlb.utils.process:20:56:19.849] Running cmd `Rscript --vanilla -e ".libPaths('/bench/frameworks/mlr3automl/lib'); source('/bench/frameworks/mlr3automl/exec.R'); run('/input/org/openml/www/datasets/4535
/dataset_train_1.arff', '/input/org/openml/www/datasets/4535/dataset_test_1.arff', target.index = 42, 'classification', '/output/predictions/census-income/1/predictions.csv', 4, time.budget = 600, meta_results_f
ile='/output/meta_results.csv', seed='710346650', name='census-income')"` with input: None
[INFO] [amlb.print:20:56:20.497]
[INFO] [amlb.print:20:56:20.498] Attaching package: ‘mlr3extralearners’
[INFO] [amlb.print:20:56:20.498]
[INFO] [amlb.print:20:56:20.498] The following objects are masked from ‘package:mlr3’:
[INFO] [amlb.print:20:56:20.502]
[INFO] [amlb.print:20:56:20.503]     lrn, lrns
[INFO] [amlb.print:20:56:20.503]
[INFO] [amlb.print:20:56:20.503] Loading required package: paradox
[INFO] [amlb.print:20:56:31.487] [1] "Finished loading data after  9.65436553955078  seconds"
[INFO] [amlb.print:20:56:31.488] [1] "remaining budget:  590  seconds"
[INFO] [amlb.print:20:56:32.175] Error: Target column 'V42' must have at least two levels
[INFO] [amlb.print:20:56:32.176] Timing stopped at: 0.485 0.007 0.37
[INFO] [amlb.print:20:56:32.431]
[INFO] [amlb.print:20:56:32.431] Execution halted
[INFO] [amlb.print:20:56:32.431]
[INFO] [amlb.print:20:56:32.431]
[INFO] [amlb.print:20:56:32.432]
[DEBUG] [amlb.utils.process:20:56:32.432] [1] "Finished loading data after  9.65436553955078  seconds"
```

My best guess is that this is some data loading issue, could you still have a closer look?

Coorsaa:
thx for checking @PGijsbers, it's probably a glmnet specific error which occurs in some ege case situations. @a-hanf will have a look after his christmas holidays.

a-hanf:
I had a look today and it seems unrelated to `mlr3automl` or `glmnet`.
I executed the benchmark locally and had a look at the data. After reading the training data here:
https://github.com/openml/automlbenchmark/blob/beef02403389809d6722dbc81dd4612910f24861/frameworks/mlr3automl/exec.R#L14
the training data set has only missing values in the labels column. I could imagine two reasons:
- AutoML benchmark saving the data as .arff is messing up
- the .arff parser is reading it incorrectly


For the first point - do other non-R frameworks read this from the .arff correctly? That would mean the second point is the critical one.
For the second point - do other R frameworks work correctly here? We should be able to load the data the same way as other frameworks then.

I don't know enough about .arff parsing / how the benchmark handles these data sets internally to debug this. @Coorsaa can you investigate further?

PGijsbers:
Seems like it works with the latest changes to `stable-v2`! 🎉

PGijsbers (approved):

PR #447, opened 2022-01-06
Signal to encode predictions as proba now works

PGijsbers:
In a previous iteration it was encoded as a numpy file, but now it's
serialized to JSON which means that results.probabilities is simply a
string if imputation is required.


sebhrusen (approved):
LGTM, thx @PGijsbers

Issue #264, opened 2021-03-12
Schedule for the 2021 benchmark


dev-rinchin:
Hi! 
Do you have a schedule for the 2021 benchmark run? When are you planing to get the results?

dev-rinchin:
Where can I find a list of fixed framework versions for the 2021 benchmark?
@PGijsbers I found the list here:
https://github.com/openml/automlbenchmark/blob/fw_2020q1/resources/frameworks_2021Q1.yaml
Do you plan to add LightAutoML?

PGijsbers:
Sorry. Yes, we are planning to add LightAutoML! I can't promise an exact timeline - things have not been going according to plan - but we are doing it as soon as possible. I'm still hopeful we can start experiments within a month.

alexmryzhkov:
Hello @PGijsbers!

I would like to ask you how is it going with the benchmark? Are there any expectations about the moment it will be finished and the final results will be announced? 

Alex

PGijsbers:
Sorry for the slow reply. We're looking to start next week (but won't immediately queue all jobs). I am still hopeful results will be available before the end of the month, but I've underestimated the amount of work before 😞.

alexmryzhkov:
@PGijsbers Thanks for info, nice to hear. 104 datasets will take enormous resources power ;)

dev-rinchin:
@PGijsbers hi! any updates?

PGijsbers:
Unfortunately I wasn't able to work much on the benchmark recently, but we're almost there. I've been very poor with estimating work left, but at this point I think there is almost nothing that _needs_ to be done before we can start with a `stable` release. I'll run validation tests on the latest versions of all the frameworks next week and contact authors who released versions since last reporting in #235 (speaking of- can you confirm that we should fix LAMA to 0.2.15?). I think I'm done giving any estimates of when results are in, burnt myself enough on that in the past, but soon™.

PGijsbers:
We just did a stable release, which is in principle the version we will use (we will only patch if required). We will now start preparing for the experiments by building new docker images and doing a few trials. Hopefully we can start on a large scale next week. I still expect it to take a few weeks before we have results as there is a significant number of compute hours involved. Unfortunately our AWS permissions are limited to fewer instances than we would like and spot instances seem rather unavailable lately, but we are doing our best.

dev-rinchin:
Hi @PGijsbers! have you finished the experiment? Where I can found the results for the experiments?
[...]

PR #449, opened 2022-01-10
update lightautoml to latest

dev-rinchin:


sebhrusen:
@PGijsbers is this upgrade the result of a discussion with you?
Surprised to see this PR in the definitions file for the current experiment and targetting `stable-v2`.

PGijsbers:
I suppose it's based off https://github.com/openml/automlbenchmark/issues/264. But no, we did not discuss bumping the `lightautoml` version for the experiments.

@dev-rinchin the experiments are already underway. To keep things clear we keep framework versions fixed through the experiments (unless there are technical issues, in which case we allow an update which specifically only targets those technical issues). We do regret it's taking a few months instead of a few weeks (versions were fixed end of September), but still feel we're in a fair window. For this reason I will close the PR.

Note that it's still possible to benchmark any `lightautoml` version with the tool by using custom frameworks, or the most recent versions by using `latest` (from repo's `master`) or `stable` (from pypi).

dev-rinchin:
I didn't discuss it with Pieter, just discovered this PR: https://github.com/openml/automlbenchmark/commit/31f18e8ce9d161d8cbb773d1c384737b964c24a6
so I thought we could still update the version and we can discuss it here. As I see now, it's wrong, thank you for the answer.

PGijsbers:
Sorry for the confusion. That version bump was indeed to fix technical difficulties (in this case, the framework wouldn't install). The only changes made after September to that framework were to fix a naming issue for a function and change the way dependencies were installed. This allowed us to install and evaluate `mlr3automl`. The core optimization is identical to the version released in September.

PR #455, opened 2022-03-23
Update to support AutoGluon v0.4

Innixma:
PR to add support for AutoGluon v0.4

- Updates install from source logic to align with v0.4
- Added `AutoGluon_hq` and `AutoGluon_gq` presets that represent mid-way points between the quality of `AutoGluon_bestquality` and the inference speed of `AutoGluon`.
- Added loading test data prior to inference for a more genuine inference latency (previously loading data from file was included in inference time)
- Added `predictor.persist_models('best')` call to avoid loading models from disk during inference (for more genuine inference latency estimates).

### DIFF
--- a/frameworks/AutoGluon/setup.sh
+++ b/frameworks/AutoGluon/setup.sh
@@ -9,16 +9,17 @@ fi
 
 # creating local venv
 . ${HERE}/../shared/setup.sh ${HERE} true
-#if [[ -x "$(command -v apt-get)" ]]; then
-#    SUDO apt-get install -y libomp-dev
+
+# Below fixes seg fault on MacOS due to bug in libomp: https://github.com/awslabs/autogluon/issues/1442
 if [[ -x "$(command -v brew)" ]]; then
-    brew install libomp
+    wget https://raw.githubusercontent.com/Homebrew/homebrew-core/fb8323f2b170bd4ae97e1bac9bf3e2983af3fdb0/Formula/libomp.rb
+    brew install libomp.rb
+    rm libomp.rb
 fi
 
 PIP install --upgrade pip
 PIP install --upgrade setuptools wheel
-PIP install "mxnet<2.0.0"
-PIP install "scikit-learn-intelex<2021.3"
+PIP install "scikit-learn-intelex<2021.6"
 
 if [[ "$VERSION" == "stable" ]]; then
     PIP install --no-cache-dir -U ${PKG}
@@ -29,11 +30,15 @@ else
     rm -Rf ${TARGET_DIR}
     git clone --depth 1 --single-branch --branch ${VERSION} --recurse-submodules ${REPO} ${TARGET_DIR}
     cd ${TARGET_DIR}
-    PIP install -e core/
+    # Note: Normally we would just call `./full_install.sh` but because `pip` and `PIP` are not the same,
+    # the script does not work here. Therefore, we have to explicitly install each submodule below.
+    # This has the downside that source installs of old versions of the package might not follow the below steps.
+    # It is recommended to instead use pip install of older AG versions to ensure it works correctly.
+    # https://github.com/awslabs/autogluon/blob/master/full_install.sh
+    PIP install -e common/
     PIP install -e features/
+    PIP install -e core/[all]
     PIP install -e tabular/[all]
-    PIP install -e mxnet/
-    PIP install -e extra/
     PIP install -e text/
     PIP install -e vision/
     PIP install -e autogluon/
--- a/frameworks/AutoGluon/setup.sh
+++ b/frameworks/AutoGluon/setup.sh
@@ -12,36 +12,28 @@ fi
 
 # Below fixes seg fault on MacOS due to bug in libomp: https://github.com/awslabs/autogluon/issues/1442
 if [[ -x "$(command -v brew)" ]]; then
-    wget https://raw.githubusercontent.com/Homebrew/homebrew-core/fb8323f2b170bd4ae97e1bac9bf3e2983af3fdb0/Formula/libomp.rb
-    brew install libomp.rb
-    rm libomp.rb
+    wget https://raw.githubusercontent.com/Homebrew/homebrew-core/fb8323f2b170bd4ae97e1bac9bf3e2983af3fdb0/Formula/libomp.rb -P "${HERE}/lib"
+    brew install "${HERE}/lib/libomp.rb"
 fi
 
 PIP install --upgrade pip
 PIP install --upgrade setuptools wheel
-PIP install "scikit-learn-intelex<2021.6"
 
 if [[ "$VERSION" == "stable" ]]; then
-    PIP install --no-cache-dir -U ${PKG}
+    PIP install --no-cache-dir -U "${PKG}"
+    PIP install --no-cache-dir -U "${PKG}.tabular[skex]"
 elif [[ "$VERSION" =~ ^[0-9] ]]; then
-    PIP install --no-cache-dir -U ${PKG}==${VERSION}
+    PIP install --no-cache-dir -U "${PKG}==${VERSION}"
+    PIP install --no-cache-dir -U "${PKG}.tabular[skex]==${VERSION}"
 else
     TARGET_DIR="${HERE}/lib/${PKG}"
     rm -Rf ${TARGET_DIR}
     git clone --depth 1 --single-branch --branch ${VERSION} --recurse-submodules ${REPO} ${TARGET_DIR}
     cd ${TARGET_DIR}
-    # Note: Normally we would just call `./full_install.sh` but because `pip` and `PIP` are not the same,
-    # the script does not work here. Therefore, we have to explicitly install each submodule below.
-    # This has the downside that source installs of old versions of the package might not follow the below steps.
-    # It is recommended to instead use pip install of older AG versions to ensure it works correctly.
-    # https://github.com/awslabs/autogluon/blob/master/full_install.sh
-    PIP install -e common/
-    PIP install -e features/
-    PIP install -e core/[all]
-    PIP install -e tabular/[all]
-    PIP install -e text/
-    PIP install -e vision/
-    PIP install -e autogluon/
+    PY_EXEC_NO_ARGS="$(cut -d' ' -f1 <<<"$py_exec")"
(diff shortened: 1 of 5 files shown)


frameworks/AutoGluon/setup.sh
```diff
@@ -9,16 +9,17 @@ fi
...
+    wget https://raw.githubusercontent.com/Homebrew/homebrew-core/fb8323f2b170bd4ae97e1bac9bf3e2983af3fdb0/Formula/libomp.rb
+    brew install libomp.rb
+    rm libomp.rb
 fi
 
 PIP install --upgrade pip
 PIP install --upgrade setuptools wheel
-PIP install "mxnet<2.0.0"
-PIP install "scikit-learn-intelex<2021.3"
+PIP install "scikit-learn-intelex<2021.6"
```
PGijsbers, line 22: Why are these dependencies installed separately?

resources/frameworks.yaml
```diff
@@ -32,9 +32,26 @@ AutoGluon_bestquality:
...
+  params:
+    presets: high_quality
+
+AutoGluon_gq:
+  extends: AutoGluon
+  description: |
+    AutoGluon with 'good_quality' preset provides a strong predictor with 16x+ faster inference speed than 'best_quality'.
+    Refer to https://auto.gluon.ai/stable/tutorials/tabular_prediction/tabular-quickstart.html#presets for a description of the presets.
+  params:
+    presets: good_quality
```
PGijsbers, line 53: @sebhrusen it feels like this is getting unwieldy. Should we extend the configuration format to more easily allow for different presets?

PGijsbers (commented):
Thanks for the update and sorry for the inactivity. I will be defending my dissertation in two weeks, and I've been very busy with all the necessary preparations. Things should return to normal (for me) starting in June.

frameworks/AutoGluon/setup.sh
Innixma: Currently it isn't part of our build-from-source install by default, and I haven't tested across windows, mac, and ARM to ensure this isn't unstable in those configs. It is stable on Linux though. If it is important, I could switch to having it in the same line as the general pip install (`pip install autogluon.tabular[all,skex]`), but would prefer it to be as it is now for stability reasons outside of the benchmark.

Innixma:
No worries @PGijsbers , and thanks for your review :)

frameworks/AutoGluon/setup.sh
Innixma: Basically the alternative is to shift it to be specified in two locations:

for pypi install:

instead of:
`PIP install autogluon`

it would be:

`PIP install autogluon.tabular[all,skex]`

and for source install:

instead of

`PIP install -e tabular/[all]`

it would be:

`PIP install -e tabular/[all,skex]`

PGijsbers: I was just trying to consider ways to make the relationship of the installation script to current versions less hard-coded. In principle, we would like to be able to provide backward compatibility to older versions of the frameworks when reasonable.

Innixma: This is hard currently because of the existence of `PIP` being different from `pip`. I'm unsure the reasoning for this in AutoMLBenchmark, but it is the reason I can't use AutoGluon's default build-from-source logic mentioned in our install instructions: https://auto.gluon.ai/stable/index.html

This is the file I'd like to call to build AG from source: https://github.com/awslabs/autogluon/blob/master/full_install.sh

However, I can't because this file uses `pip` while I need to use `PIP` in AutoMLBenchmark.

If this can be resolved then it should be do-able to have backwards compatibility

resources/frameworks.yaml
sebhrusen: well, I don't have the feeling that the configuration format is the problem here (or is it? do you have sth in mind like a specific CLI syntax only for presets? sth that would be simple enough to work for various frameworks?)

I mean users can create as many framework definitions in their user `frameworks.yaml`, we don't have to accept all of those in the default `frameworks.yaml`.

sebhrusen: currently, CLI syntax allows passing framework params like
`-Xf.presets=good_quality` but this works only for local runs, this is not forwarded to ec2 instances or to docker images, we need custom `frameworks.yaml` for that, but maybe this could be improved (whitelisting?) although I'm not sure I like the idea of having params coming from different sources when executed remotely.

frameworks/AutoGluon/setup.sh
```diff
@@ -29,11 +30,15 @@ else
     rm -Rf ${TARGET_DIR}
     git clone --depth 1 --single-branch --branch ${VERSION} --recurse-submodules ${REPO} ${TARGET_DIR}
     cd ${TARGET_DIR}
-    PIP install -e core/
+    # Note: Normally we would just call `./full_install.sh` but because `pip` and `PIP` are not the same,
+    # the script does not work here. Therefore, we have to explicitly install each submodule below.
+    # This has the downside that source installs of old versions of the package might not follow the below steps.
+    # It is recommended to instead use pip install of older AG versions to ensure it works correctly.
+    # https://github.com/awslabs/autogluon/blob/master/full_install.sh
```
sebhrusen, line 37: hack alert! what about sth like:
```bash
env PATH=`dirname $py_exec`:$PATH  bash -c ./full_install.sh
```
haven't tried, but shouldn't it work?

frameworks/AutoGluon/setup.sh
sebhrusen: @Innixma see my suggestion below.

> I'm unsure the reasoning for this in AutoMLBenchmark

it was mainly to hide the automatic creation of the venv, but still be able to access both default python exec and the venv python exec if needed.

the exec is stored in `$py_exec` variable if you need it, for example to set a custom PATH as I suggest below. I think this should be enough for your needs.

Innixma: Thanks for this! I wasn't so savvy with bash so this was very helpful.

frameworks/AutoGluon/setup.sh
```diff
@@ -9,16 +9,17 @@ fi
...
 . ${HERE}/../shared/setup.sh ${HERE} true
-#if [[ -x "$(command -v apt-get)" ]]; then
-#    SUDO apt-get install -y libomp-dev
+
+# Below fixes seg fault on MacOS due to bug in libomp: https://github.com/awslabs/autogluon/issues/1442
 if [[ -x "$(command -v brew)" ]]; then
-    brew install libomp
+    wget https://raw.githubusercontent.com/Homebrew/homebrew-core/fb8323f2b170bd4ae97e1bac9bf3e2983af3fdb0/Formula/libomp.rb
+    brew install libomp.rb
+    rm libomp.rb
```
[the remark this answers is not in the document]
Innixma: Thanks! I updated the PR to use a git ignored folder for this.

frameworks/AutoGluon/setup.sh
Innixma: Thanks! This makes things a lot nicer and allows us to support AutoGluon v0.1.0+ all at once. Refer to the updated PR.

frameworks/AutoGluon/setup.sh
Innixma: Re scikit-intelex, I've resolved this by not hardcoding the version but instead using the version defined via the `skex` extra_dependencies key.

Innixma:
@PGijsbers @sebhrusen thanks for the comments! I have updated the PR to reflect them and this version should now be compatible with AutoGluon versions `>=0.1.0` (including v0.4, v0.5, and soon to be released v0.5.1).

resources/frameworks.yaml
PGijsbers: I don't really know what I had in mind at the time. But if every framework does this, it will blow up the configuration file many times over. MLJar has 4 presets, GAMA has more presets, and I suspect (and hope) more frameworks might follow.

A custom parameter which is specifically for the mode/preset might be reasonable, considering it's a "special" hyperparameter in the sense that it's the only one we allow for benchmarking. Consider the last three additional entries in the framework definition:

```yaml
AutoGluon:
  version: "stable"
  description: |
    AutoGluon-Tabular: Unlike existing AutoML frameworks that primarily focus on model/hyperparameter selection,
    AutoGluon-Tabular succeeds by ensembling multiple models and stacking them in multiple layers.
  project: https://auto.gluon.ai
  refs: [https://arxiv.org/abs/2003.06505]
  presets: ['best_quality', 'high_quality', 'good_quality']
  preset: best_quality    # could also be the first of `modes`  by convention
  param_name: presets  # the framework name for the hyperparameter, so that we can easily convert internally to add it to `params` to avoid changing integration scripts
```
This could be complimented by a CLI argument which overrides the `mode` argument. It wouldn't solve the issue of forwarding.

On the other hand, it requires additional logic for something that works well already (even if it's minimal).

PGijsbers (approved):
I can't test it currently (I can't install fairscale locally and right now have no access to aws), but in principle it looks good to me.

resources/frameworks.yaml
sebhrusen: maybe the first mistake was to allow non-default presets in the `resources/frameworks.yaml`.
The tool offers the possibility for anyone to add custom configurations, we're not going to limit that, and for the paper's benchmarks, we use the `franeworks_yyyyQn.yaml` files where we can decide to use non-default presets after agreeing with each team ("was it a good idea to allow multiple configs for those?" is a separate question).
The problem with making `preset` special is that we're officially encouraging this approach (which I like btw), and, as we already see, it creates issues on our side (multiplication, way to fine tune the framework only for the  benchmark, in contradiction with the original approach of benchmarking all tools with their default setup...).

That's why personally, I'd go backwards and accept only default setup in `frameworks.yaml`, rather than creating a new convoluted syntax. And we would decide "customizations" only for the `franeworks_yyyyQn.yaml` ones.
My second choice would be to just add a param/section to the framework description whitelisting framework params that are always forwarded, it's relatively simple to do and will just allow the usual CLI syntax also for docker and aws.

Innixma:
Any other action on my end to take for this PR?

PGijsbers:
I was waiting to see if @sebhrusen agrees. We don't (yet) have a policy in place that determines what to do if someone does not check in.

sebhrusen (approved):
@PGijsbers do we agree to create a follow-up ticket to clean up the `frameworks.yaml`?

PR #464, opened 2022-06-02
For docker to work and work better

alanwilter:
This line `**/.setup` in `.dockerignore` denies 

```bash 
python3 runbenchmark.py ... -m docker -s force
```
execution, e.g.: `automlbenchmark/frameworks/TPOT/.setup/Dockerfile`.

And to avoid docker created files to be owned by `root` on host filesystem.

### DIFF
--- a/amlb/runners/docker.py
+++ b/amlb/runners/docker.py
@@ -54,6 +54,7 @@ class DockerBenchmark(ContainerBenchmark):
         inst_name = f"{self.sid}.{str_sanitize(str_digest(script_params))}"
         cmd = (
             "docker run --name {name} {options} "
+            '-u "$(id -u $USER):$(id -g $USER)" '
             "-v {input}:/input -v {output}:/output -v {custom}:/custom "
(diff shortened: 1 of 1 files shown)


PGijsbers:
Thanks for the contribution! The setup is fixed in the `stable-v2` branch (https://github.com/openml/automlbenchmark/pull/437), but unfortunately, we have not been able to get around to merging this with the main branch. After we merge `stable-v2` back to `master` we will see how to incorporate the ownership patch (at first glance it looks like it might be Unix only, and we would like to make sure running through docker also works on Windows). Bear with us while sort this out 😓

sebhrusen:
@PGijsbers I'll have time to work on `automlbenchmark` this week. I have some work done to fix the workflow, so we should be able to merge `stable-v2` to `master` in a few days.

PGijsbers:
That's great news! 🎉

sebhrusen (approved):
rebased your PR as I just merged docker fix in `stable-v2` back to `master`, but still keeping your user part and now merging it on top of `master`.
Thanks for this fix.

Issue #466, opened 2022-06-18
Encoding error causing ResultError


Robinnibor:
I am getting the following error:
```
     id    task fold framework constraint      metric  duration      seed                                                    info
disease disease    0      TPOT       1h4c neg_logloss      68.5 970204891 ResultError: y contains previously unseen labels: '1.0'
```
At first glance I thought that I have label in test data which is unseen in my train data.
However, I still get this error when using the same data for both training and testing

My .yaml looks like this
```
- name: disease
  dataset:
    train: ./test_data/automl_disease.csv
    test:  ./test_data/automl_disease.csv
    target: output
    type: multiclass
    #type: binary
  folds: 1
```

And I find that if I modify the label_encoder in `amlb/data.py` to use int instead of float, the error is gone.
ex:
```
    @lazy_property
    def label_encoder(self):
        return Encoder('label' if self.values is not None else 'no-op',
                       target=self.is_target,
                       encoded_type=int if self.is_target and not self.is_numerical() else int, # <= this "int" was originally "float"
                       missing_values=[None, np.nan, pd.NA],
                       missing_policy='mask' if self.has_missing_values else 'ignore',
                       normalize_fn=self.normalize
                       ).fit(self.values)
```
I am wondering what is the proper way to resolve it.

Robinnibor:
[automl_disease.csv](https://github.com/openml/automlbenchmark/files/8932830/automl_disease.csv)
This is the data I am using.

sebhrusen:
Hi @Robinnibor, thanks for reporting this.
I'll try to reproduce it with TPOT and compare with `constantpredictor` or `randomforest`?

My guess is that it's either a parsing issue of test data or of TPOT predict results, rather than an encoding one: we had to handle this for some frameworks.

sebhrusen:
@Robinnibor I seem to have a fix, will properly test and create a PR tmrw.

Robinnibor:
Same setting also do not work on H2Oautoml, AutoGluon, GAMA, MLNet, autosklearn etc. but might be the other reason?
The tree-based framework works, such as DecisionTree RandomForest TunedRandomForest. Other frameworks don't.

PR #467, opened 2022-06-22
issue 466: fixed parsing of int targets when loading file in CSV format

sebhrusen:
Fix for https://github.com/openml/automlbenchmark/issues/466

When user uses the file loading system and specifies in his benchmark definition that the dataset corresponds to a classification task, then `int` targets should be converted into categoricals. Otherwise, it can be interpreted variously by frameworks. Also, with the sklearn-based frameworks using encoded data (numpy arrays), `amlb` was failing to produce a correct predictions file as it could not interprete framework predictions (e.g. as float 1.0) as a target value (1).


Robinnibor:
Some tests are failing

PGijsbers:
I think the GAMA failures are a GAMA issue ([fixed here](https://github.com/openml-labs/gama/pull/148)). We set up the tests so that one failure cancels all remaining tests. Otherwise, it requires a lot of additional compute and if the error is caused by the benchmark tool as opposed to the automl framework, it would be very wasteful. We should probably redesign this somehow so it's less fragile to any individual framework failing however, because this happens too regularly.

PGijsbers (approved):
Looks good, left a few comments. @Robinnibor can you confirm this fixed your issue?

I got one failure the first time I ran `test_load_multiclass_task_with_num_target_csv`:
```
================================== FAILURES ===================================
________________ test_load_multiclass_task_with_num_target_csv ________________

file_loader = FileLoader({})

    @pytest.mark.use_disk
    def test_load_multiclass_task_with_num_target_csv(file_loader):
        ds_def = ns(
            train=os.path.join(res, "iris_num_train.csv"),
            test=os.path.join(res, "iris_num_test.csv"),
            target="class",
            type="multiclass"
        )
        ds = file_loader.load(ds_def)
        assert ds.type is DatasetType.multiclass
        _assert_X_y_types(ds.train)
        _assert_data_consistency(ds)
>       _assert_data_paths(ds, ds_def)

test_file_dataloader.py:125: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
test_file_dataloader.py:241: in _assert_data_paths
    assert dataset.train.data_path(f) == path_from_split(s)
..\..\..\..\..\amlb\datasets\file.py:188: in data_path
    return self._get_data(format)
..\..\..\..\..\amlb\datasets\file.py:207: in _get_data
    self._data[fmt] = converter.convert(self)
..\..\..\..\..\amlb\datasets\file.py:399: in convert
    self._write_file(split.data, target_path)
..\..\..\..\..\amlb\datasets\file.py:422: in _write_file
    arff.dump(dict(
..\..\..\..\..\venv\lib\site-packages\arff.py:1091: in dump
    for row in generator:
..\..\..\..\..\venv\lib\site-packages\arff.py:1028: in iter_encode
    yield self._encode_attribute(attr[0], attr[1])
..\..\..\..\..\venv\lib\site-packages\arff.py:964: in _encode_attribute
    type_tmp = [u'%s' % encode_string(type_k) for type_k in type_]
..\..\..\..\..\venv\lib\site-packages\arff.py:964: in <listcomp>
    type_tmp = [u'%s' % encode_string(type_k) for type_k in type_]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

s = 1

    def encode_string(s):
>       if _RE_QUOTE_CHARS.search(s):
E       TypeError: cannot use a string pattern on a bytes-like object

..\..\..\..\..\venv\lib\site-packages\arff.py:420: TypeError
=========================== short test summary info ===========================
FAILED test_file_dataloader.py::test_load_multiclass_task_with_num_target_csv
============================== 1 failed in 0.59s ==============================
```
I assume it's a Windows issue, but I can't say I understand why it only was an issue the first run.
Since it does not seem reproducible and we don't officially support Windows, I don't mind the merge regardless (it can always be sorted out later).

As an aside, I noticed some logging changes mixed in. It would be much appreciated if PRs can be contained to single issues as much as possible (e.g., one for data loading, one for minor logging updates). Smaller PRs are easier to deal with for a number of reasons (though I'll admit I too violate this sometimes).

Robinnibor:
Yes thanks it fixes the issue. TPOT and H2Oautoml are working now.
Others are still failing but might due to other issue.

Issue #416, opened 2021-11-09
Make `versioning` workflow work for subsequent releases


PGijsbers:
The [`versioning` workflow](https://github.com/openml/automlbenchmark/blob/master/.github/workflows/versioning.yml) fails on any release which is not immediately based on `master` because the `__version__` file would have previously been updated. E.g., the [v2.0.2](https://github.com/openml/automlbenchmark/runs/4149721150?check_suite_focus=true) fails to update the version to `2.0.2` because it should replace `2.0.1` instead of `_dev_version`.

PGijsbers:
Additionally, `stable` tag now lags behind:
![image](https://user-images.githubusercontent.com/15890747/147818931-ddaaadae-f193-4ecc-9c2c-2824726ff0a3.png)
![image](https://user-images.githubusercontent.com/15890747/147818952-93d64fe5-b31b-4a8b-81ab-5c2bb664b416.png)

PR #468, opened 2022-06-28
Issue 416: fixing versioning workflow for releases and merges to master

sebhrusen:
Addressing https://github.com/openml/automlbenchmark/issues/416

This workflow:
- automatically updates `__version__.py` during releases to set the app version to the release tag.
- if the new release is the highest version, it will also automatically tag it as `stable`. Note that this is for the highest (not latest) version, so that we can for example possibly create a hotfix release for `stable-v(N-1)` without making it the stable one.
- as the release should be done from a `stable-vN` branch, it also automatically adds a new commit after the release, resetting the version to `dev`. This also allows smooth merge of `stable-vN` branches back to `master`.
- automatically reset the version to `dev` in case of "forced" merge of released tag back to `master`: this should not be a common scenario, but better safe than worry.


.github/workflows/stable-tag.yml
```diff
@@ -1,24 +1,16 @@
-name:  Tag with Stable
+name: Tag with Stable
```
PGijsbers, line 1: Does the revised versioning workflow not already apply a stable tag in its tagging step?

.github/workflows/versioning.yml
```diff
@@ -3,7 +3,7 @@ name: Version Bump
 on:
```
PGijsbers, line 3: Do we ever want to release from `master`? I don't believe it's something we currently want in our workflow? If not, we should probably add a check for that here. I have in the past accidentally released from `master` instead of `stable-vX`, because `master` is our default branch and I forgot to change the option in the release UI.

PGijsbers (approved):
Overall I think this is great work, thanks a lot :) I left some comments with questions.

.github/workflows/stable-tag.yml
PGijsbers: i.e. is there a reason to have this workflow at all?

sebhrusen: outside if we need to run this one manually on some occaion, we don't need it.
In the `git-playground` repo, I actually deactivated it...

sebhrusen: actually, I'm ready to delete it: I can't see any scenario where we would have to run this: I expect the release workflow to be more reliable for this now.

.github/workflows/versioning.yml
sebhrusen: No, I don't think we should ever release from `master`.

Let me find what's the best way to constraint this here.

.github/scripts/version-set.sh
```diff
@@ -0,0 +1,40 @@
...
+  echo $(get_version "$1" "$2")
+}
+
+OLD_VERSION=$(get_version "$VERSION_FILE" "$VERSION_PLACEHOLDER")
+DEV_VERSION=$(get_version "$VERSION_FILE" "$DEV_PLACEHOLDER")
+NEW_VERSION=$(echo ${VERSION:-"$DEV_VERSION"} | sed -E 's/^v([0-9].*)/\1/')
+echo "setting version from \"$OLD_VERSION\" to \"$NEW_VERSION\""
+FINAL_VERSION=$(set_version "$VERSION_FILE" "$VERSION_PLACEHOLDER" "$NEW_VERSION")
+echo "version changed to \"$FINAL_VERSION\""
+echo "VERSION=$FINAL_VERSION" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
sebhrusen: Originally I think still wanted to use:
```__version__ =  _dev_version```
But looking at the new script, I don't see a major difference. The only important thing is to ensure that `_dev_version` value and `__version__ = ` (on a dev branch) are in sync: the script guarantees this after releases, but not if for some reason, we manually change the value of `_dev_version`. However, I don't see any reason why we would do that... so I feel like leaving this as is to avoid adding twisted logic to this script.

File: .github/workflows/versioning-reset.yml [hunk 4]
```diff
@@ -0,0 +1,27 @@
+name: Version reset to dev
+# reset the version if merging a release branch into main
+
+on:
+  push:
+    branches:
+      - main
+      - master
+    paths:
+      - 'app/__version__.py'
```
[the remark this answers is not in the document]
sebhrusen: > In principle, this should no longer trigger after this patch is merged with master, right? Since the new release workflow resets __version__ to dev after a tagged commit with versioning is created, and then the __version__.py file is reset which should be identical to what lives in the master branch?

correct, this should not be triggered if merging the head of `stable-vN` onto `master`.
But, this could be triggered for example in the following scenario:
- working on `stable-v2`.
- make release `v2.15` (head of `stable-v2` has `__version__ = "dev"`).
- add some unreleased patches to `stable-v2`.
- decide to merge released v2 patches to `master` (and only those): simplest way is to create a PR from tag `v2.15` into `master`, review it quickly and merge. In this case the `versioning-reset` action will apply immediately after the merge.

.github/workflows/versioning.yml
sebhrusen: should be addressed by the new changes: learned much about GH workflows thanks to this one, especially to be able to bypass the default parallelism of actions and jobs

sebhrusen:
@PGijsbers you may want to look at the new changes to protect the main branch. Otherwise, I'm ready to merge this, then I'll merge this to master as well as some other pending PRs.

.github/workflows/versioning.yml
PGijsbers: Nice, we also use that to first test on baselines and only then frameworks too 👍

PGijsbers:
LGTM

PR #469, opened 2022-06-28
Allow custom tag for all AWS entities

sebhrusen:
Having properly tagged AWS objects can be a requirement in some companies to be able to launch and administrate those objects (used internally at H2O.ai), so I added a non-invasive way of specifying custom tags for all AWS created created by the `amlb` app.


PGijsbers (approved):
Looks fine and did a manual test to confirm.

PR #479, opened 2022-07-25
New site

PGijsbers:
This PR:
 - removes the old site
 - adds the static files for the new site
 - moves the old notebooks in the reports folder to `reports/old`
 - adds a readme placeholder in the `reports` folder to indicate the new notebooks will be uploaded soon.
 
Not in this PR:
 - a site responsive (it is not yet mobile friendly)
 - links and citation information for the arxiv paper to be published tonight
 - scripts to generate the site from files (e.g., yaml files with entries for each framework)
 - the notebooks used for generating the figures and tables in the updated paper
 - a solution to world hunger

PR #480, opened 2022-07-25
Update minimal responsiveness

PGijsbers:
Should make the of #479  site minimally navigable on mobile. It's not great or that well tested... but better than nothing, I suppose. Will revisit later (in a separate PR). I know the icons aren't the best and that there should be an option for a text menu but this is all I could manage before the next part of the AutoML Conf :)


sebhrusen (approved):
LGTM, thanks @PGijsbers

PR #481, opened 2022-07-27
Link to arxiv paper

PGijsbers:
🎉  our preprint is finally live, so I updated the site with the relevant information


sebhrusen (approved):
nice!

PR #488, opened 2022-08-04
Fix Namespace error

Innixma:
#469 appears to have introduced a bug where running in AWS mode will crash immediately due to a key error if the user doesn't specify `instance_tags` in a custom config. This code fixes the error.

Command:

```
python runbenchmark.py constantpredictor test -f 0 -m aws -p 10
```

Error:

```
---------------------------------------------------
Starting job aws.test.test.kc2.0.constantpredictor.
Starting new EC2 instance with params: constantpredictor test test -t kc2 -f 0 -Xseed=380416698
Job aws.test.test.kc2.0.constantpredictor failed with: 'Namespace' object has no attribute 'instance_tags'
fatal: not a git repository (or any of the parent directories): .git

Metric scores: { 'app_version': 'dev [NA, NA, NA]',
  'constraint': 'test',
  'duration': nan,
  'fold': 0,
  'framework': 'constantpredictor',
  'id': 'openml.org/t/3913',
  'info': "AttributeError: 'Namespace' object has no attribute 'instance_tags'",
  'metric': None,
  'mode': 'aws',
  'models_count': nan,
  'params': '',
  'predict_duration': nan,
  'result': nan,
  'seed': None,
  'task': 'kc2',
  'training_duration': nan,
  'type': None,
  'utc': '2022-08-04T00:46:26',
  'version': None}
Job `aws.test.test.kc2.0.constantpredictor` executed in 0.021 seconds.
Error when handling state change to State.completing for job `aws.test.test.kc2.0.constantpredictor`: cannot unpack non-iterable bool object
  File "/home/ubuntu/benchmark/ag-2022_08_03_hstree/tmp6/automlbenchmark/amlb/job.py", line 158, in set_state
    skip_default = bool(self._on_state(state))
  File "/home/ubuntu/benchmark/ag-2022_08_03_hstree/tmp6/automlbenchmark/amlb/runners/aws.py", line 388, in _on_state
    terminate, failure = self._download_results(_self.ext.instance_id)
TypeError: cannot unpack non-iterable bool object

```

### DIFF
--- a/amlb/runners/aws.py
+++ b/amlb/runners/aws.py
@@ -574,8 +574,14 @@ class AWSBenchmark(Benchmark):
             if instance_def.volume_size:
                 ebs['VolumeSize'] = instance_def.volume_size
 
-            instance_tags = ec2_config.instance_tags | ns(Name=f"amlb_{inst_key}")
-            volume_tags = (ec2_config.volume_tags or instance_tags) | ns(Name=f"amlb_{inst_key}")
+            if hasattr(ec2_config, 'instance_tags'):
+                instance_tags = ec2_config.instance_tags | ns(Name=f"amlb_{inst_key}")
+            else:
+                instance_tags = ns(Name=f"amlb_{inst_key}")
+            if hasattr(ec2_config, 'volume_tags'):
+                volume_tags = (ec2_config.volume_tags or instance_tags) | ns(Name=f"amlb_{inst_key}")
+            else:
+                volume_tags = instance_tags | ns(Name=f"amlb_{inst_key}")
             instance_params = dict(
                 BlockDeviceMappings=[dict(
                     DeviceName=ec2_config.root_device_name,


sebhrusen:
@Innixma I can't reproduce your error in a default environment: with unmodified `resources/config.yaml` and with custom config without `instance_tags` entry.

The only way for me to reproduce the error is to add the following to the custom `config.yaml`:
```yaml
aws:
  ec2:
    instance_tags:
```
which will override the default `instance_tags` entry and set it to `None` instead of `{}`, leading to the error above.
But otherwise, due to hierarchical structure of the configuration files, this entry is an empty map, not `None`.

Did you by any chance edit the `resources/config.yaml` manually instead?

amlb/runners/aws.py
```diff
@@ -574,8 +574,14 @@ def _start_instance(self, instance_def, script_params="", instance_key=None, tim
             if instance_def.volume_size:
                 ebs['VolumeSize'] = instance_def.volume_size
 
-            instance_tags = ec2_config.instance_tags | ns(Name=f"amlb_{inst_key}")
-            volume_tags = (ec2_config.volume_tags or instance_tags) | ns(Name=f"amlb_{inst_key}")
+            if hasattr(ec2_config, 'instance_tags'):
```
sebhrusen, line 577: there should ALWAYS be an `instance_tags` config entry, this is by design: all config points are made available in `resources/config.yaml` with a default value, this for several reasons like single point of reference, documentation, code simplification (no need to check existence of the entry like you do here).

Innixma: This is probably on my end, I have used an EC2 instance as the main node to call runbenchmark.py for a long time, and just recently updated to latest mainline, however I am going to provide more info as I find the behavior extremely strange and can't pinpoint what I've done wrong.

I use a different directory for each run, in the format `benchmark/{run_name}/automlbenchmark`

For example:

```
ubuntu@ip-172-31-16-62:~$ ls
LICENSE  Nvidia_Cloud_EULA.pdf  README  anaconda3  benchmark  examples  src  tmp  tmp2  tools  tutorials  virtual

ubuntu@ip-172-31-16-62:~$ cd benchmark
ubuntu@ip-172-31-16-62:~/benchmark$ ls
ag-2022_05_09_rf  ag-2022_06_19  ag-2022_06_19_nn  ag-2022_06_21  ag-2022_06_21_nn  ag-2022_06_24_nn  ag-2022_06_25_nn  ag-2022_06_26_binary  ag-2022_07_12_torch  ag-2022_08_03_hstree
ubuntu@ip-172-31-16-62:~/benchmark$ cd ag-2022_08_03_hstree
ubuntu@ip-172-31-16-62:~/benchmark/ag-2022_08_03_hstree$ ls
automlbenchmark  venv
ubuntu@ip-172-31-16-62:~/benchmark/ag-2022_08_03_hstree$ source venv/bin/activate
(venv) ubuntu@ip-172-31-16-62:~/benchmark/ag-2022_08_03_hstree$ python automlbenchmark/runbenchmark.py constantpredictor test -f 0 -m aws -p 10

```

Basically I have a script which git clones the automlbenchmark branch/fork I want to run, creates a new venv, installs the AMLB dependencies, and sends the runbenchmark.py command.

The fact that this is failing with that error is pretty strange, as I know that it can detect the `resources/config.yaml` file (I've tested this, for example I need to put region in or it errors.) But as you said it somehow isn't detecting the `instance_tags={}` line (which I see in my file). Maybe it is somehow finding the config.yaml files from the other automlbenchmark source code on disk, but I can't think of the code logic that would cause that to happen (and if it did, it is strange because all the other config options work correctly).


I'll probably create a fresh EC2 instance and see if it still happens (I'm assuming it will work at that point, but no idea why).

PGijsbers: Did you manage to confirm the problem was on your end?

Innixma: Sorry for late reply, yes I was unable to reproduce the error when I created a fresh EC2 instance to run the experiments. It was a very odd bug and I'm unsure why it happened, but it seems to be gone now.

PR #492, opened 2022-08-22
Paper notebooks

PGijsbers:
Added notebooks that:
 - helped identify/classify errors
 - cleaned the data, performed imputation for baselines
 - produced some of the figures and tables for our new paper
 
 and updated the readme to refer to our new preprint.


docs/README.md
```diff
@@ -28,7 +28,7 @@ This toolkit aims to address these problems by setting up standardized environme
 Documentation: <https://openml.github.io/automlbenchmark/>
 
 ### Features:
-* Curated suites of [benchmarking datasets](https://openml.github.io/automlbenchmark/benchmark_datasets.html) from [OpenML](https://www.openml.org/s/218/data).
+* Curated suites of [benchmarking datasets]([https://openml.github.io/automlbenchmark/benchmark_datasets.html](https://github.com/openml/automlbenchmark/blob/master/docs/benchmark_datasets.md) from [OpenML](https://www.openml.org/s/218/data).
```
qingyun-wu, line 31: Should remove the html link here? 
```suggestion
* Curated suites of [benchmarking datasets](https://github.com/openml/automlbenchmark/blob/master/docs/benchmark_datasets.md) from [OpenML](https://www.openml.org/s/218/data).
```

qingyun-wu, line 31: This link: https://www.openml.org/s/218/data is not leading me to the dataset list on openml.

qingyun-wu (changes requested):

docs/README.md
PGijsbers: I contacted OpenML about this, thanks for letting me know (and for the other fix)!

PGijsbers:
We'll release updated notebooks after we finish our revision that will be better documented, but I figured I'd update the PR since I noticed it is incomplete: the `box_plot` function was missing. Now the notebook successfully runs top-to-bottom and reproduces the plots. I also included some installation instructions that may help people install the right version of dependencies (as newer versions break parts of the notebook).

PR #494, opened 2022-09-13
[PoC] AutoGluon TimeSeries Prototype

Innixma:
[Don't merge this PR]

This PR is a proof of concept of time series data and framework support in AutoMLBenchmark.

To run, follow the instructions in the newly added `frameworks/AutoGluonTS/README.md`.


Innixma:
@sebhrusen @PGijsbers 

Some questions I have:

1. Is there a way to specify information such as `prediction_length=5` on a per dataset basis? prediction_length is the look-ahead requirement for prediction and dictates the difficulty of the task. I'm wondering if I can specify it as part of the yaml file definition of the dataset in `ts.yaml`. Ditto for a couple other things like `timestamp_column="Date"` and `item_id="name"`.

2. How can I update and specify the logic that does the final scoring based on predictions and truth? It needs to be altered to work with TimeSeries. Additionally, it may take a different form, such as if the metric requires quantile predictions to calculate.

PGijsbers:
@sebhrusen I'd appreciate it if you can have a look, I have very limited availability due to a paper deadline.

sebhrusen:
@PGijsbers sunny holidays right now: will look at it when I'm back next week.

Innixma:
Update: Several of the TODO / FIXME comments have been addressed by @limpbot in https://github.com/Innixma/automlbenchmark/pull/6

Innixma:
@sebhrusen Sorry to ping but would you be interested in reviewing this PR? A large chunk of the logic was written by @limpbot who is interning with us currently, and it would be great if he received feedback so as not to block his time-series benchmarking efforts.

sebhrusen:
@Innixma I'm looking at it now and will make a full review before Monday.
Outside implementation details/modularity, I mainly want to be sure that it is not designed to first satisfy AG's timeseries implementation and can be generalized to other implementations (would be nice to have an alternative implementation): for now to satisfy your needs, I'll mainly ensure that the changes are limited to data loading + AG implementation when possible.

Innixma:
Sounds good! I agree that we should make sure the input/ouput/scoring definitions are generic and not AG specific. perhaps the AutoPyTorch-TimeSeries folks (@dengdifan) would be interesting in reviewing / trying to add on their AutoML system as a framework extension to this logic?

amlb/results.py
```diff
@@ -308,6 +314,24 @@ def save_predictions(dataset: Dataset, output_file: str,
...
+            # we aim to calculate the mean period error from the past for each sequence: 1/N sum_{i=1}^N |x(t_i) - x(t_i - T)|
+            # 1. retrieve item_ids for each sequence/item
+            item_ids, inverse_item_ids = np.unique(dataset.test.X[dataset.id_column].squeeze().to_numpy(), return_index=False, return_inverse=True)
+            # 2. capture sequences in a list
+            y_past = [dataset.test.y.squeeze().to_numpy()[inverse_item_ids == i][:-dataset.prediction_length] for i in range(len(item_ids))]
+            # 3. calculate period error per sequence
+            y_past_period_error = [np.abs(y_past_item[period_length:] - y_past_item[:-period_length]).mean() for y_past_item in y_past]
+            # 4. repeat period error for each sequence, to save one for each element
+            y_past_period_error_rep = np.repeat(y_past_period_error, dataset.prediction_length)
+            df = df.assign(y_past_period_error=y_past_period_error_rep)
```
sebhrusen, line 334: I'd rather not have this here, this looks like a lot of calculations + assumptions (apparently you can't have time series without an `id_column`) for a method that just supposed to be save predictions into a standard format. Even more as this `y_past_period_error` seems to be useful only for the `mase` metric, therefore, either you can compute it with the metric or you compute it before (in AG framework integration).

For now, I'd move your computations to the `__init__.py` or `exec.py` file, and simply ensure that we can customize the result by adding `optional` columns (in this case, this includes both `quantiles` and your additional results).

Suggestion:
change signature to
```python
 def save_predictions(dataset: Dataset, output_file: str,
                         predictions: Union[A, DF, S] = None, truth: Union[A, DF, S] = None,
                         probabilities: Union[A, DF] = None, probabilities_labels: Union[list, A] = None,
                         optional_columns: Union[A, DF] = None,
                         target_is_encoded: bool = False,
                         preview: bool = True):
```
and automatically concatenate the `optional_columns` to the predictions if provided. For now, you should be able to generate those in `exec.py`

frameworks/AutoGluonTS/setup.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
sebhrusen, line 1: it is the same setup as for default `AutoGluon`, right?
Why create another framework then? It adds a lot of complexity regarding testing, distribution, docker images and so on...
In the `AutoGluon.__init__.py` you could just fork the logic like this:
```python
exec_file = "exec_ts.py" if dataset.type is not DatasetType.timeseries else "exec.py"
return run_in_venv(__file__, exec_file,
```

resources/frameworks.yaml
```diff
@@ -191,7 +191,15 @@ TPOT:
...
 
+####################################
+### TimeSeries AutoML frameworks ###
+####################################
 
+AutoGluonTS:
+  version: "stable"
+  description: |
+    AutoGluon-TimeSeries
+  project: https://auto.gluon.ai
```
sebhrusen, line 202: Ideally, we don't want to have a new framework if it supports only a specific kind of dataset. See my comment above.
If we start to have one framework for regression, one for classification, one for time series, one for anomaly detection and so on... then if becomes hard to compare "AutoML" frameworks.

amlb/benchmark.py
```diff
@@ -489,7 +489,20 @@ def load_data(self):
...
+            if self._dataset.type == DatasetType.timeseries:
+                if self._task_def.dataset['id_column'] is None:
+                    log.warning("Warning: For timeseries task setting undefined itemid column to `item_id`.")
+                    self._task_def.dataset['id_column'] = "item_id"
+                if self._task_def.dataset['prediction_length'] is None:
+                    log.warning("Warning: For timeseries task setting undefined prediction length to `1`.")
+                    self._task_def.dataset['prediction_length'] = "1"
+                self._dataset.timestamp_column=self._task_def.dataset['timestamp_column']
+                self._dataset.id_column=self._task_def.dataset['id_column']
+                self._dataset.prediction_length=self._task_def.dataset['prediction_length']
```
sebhrusen, line 505: looks like most of this logic could reside in the loading logic itself as this is dealing with information available in `self._task_def.dataset` which is directly available to the file loader.
I'll move the logic to `dataset/file.py` for now to minimize scope of changes.

amlb/datasets/file.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, cache_dir=None):
         self._cache_dir = cache_dir if cache_dir else tempfile.mkdtemp(prefix='amlb_cache')
 
     @profile(logger=log)
-    def load(self, dataset, fold=0):
+    def load(self, dataset, fold=0, timestamp_column=None):
```
sebhrusen, line 33: you obtained this new column using
```
timestamp_column=self._task_def.dataset['timestamp_column']
```
so you already have the information in the `dataset` object

sebhrusen (changes requested):
thanks for this contribution, this is a first feedback.
I'm good with most of the file loading logic, and the added metrics.
For the "middle" layers like `benchmark` and `results` transformations, I'd like to avoid changes as much as possible there as they look more ad-hoc.

Innixma:
Thanks @sebhrusen for the detailed review!

@limpbot would you like to have a go at addressing some of the comments? Feel free to send a PR to my branch as you did in your prior update.

amlb/datasets/file.py
limpbot: true

amlb/results.py
```diff
@@ -255,7 +259,8 @@ def save_predictions(dataset: Dataset, output_file: str,
                          predictions: Union[A, DF, S] = None, truth: Union[A, DF, S] = None,
                          probabilities: Union[A, DF] = None, probabilities_labels: Union[list, A] = None,
                          target_is_encoded: bool = False,
-                         preview: bool = True):
+                         preview: bool = True,
+                         quantiles: Union[A, DF] = None):
```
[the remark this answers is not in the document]
limpbot: got it

frameworks/AutoGluonTS/setup.sh
limpbot: true

resources/frameworks.yaml
limpbot: got it

resources/benchmarks/ts.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+- name: covid
+  dataset:
+    train: https://autogluon.s3-us-west-2.amazonaws.com/datasets/CovidTimeSeries/train.csv
+    # s3://autogluon-ts-bench/data/covid_deaths/csv/train.csv | https://autogluon.s3-us-west-2.amazonaws.com/datasets/CovidTimeSeries/train.csv
+    test: https://autogluon.s3-us-west-2.amazonaws.com/datasets/CovidTimeSeries/test.csv
+    # s3://autogluon-ts-bench/data/covid_deaths/csv/test.csv  |  https://autogluon.s3-us-west-2.amazonaws.com/datasets/CovidTimeSeries/test.csv
+    target: ConfirmedCases              # target    | ConfirmedCases
+    type: timeseries
+    prediction_length: 30
```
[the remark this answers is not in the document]
limpbot: it is the number of predictions per sequence. So 'num_predictions_per_id' sounds good?

amlb/benchmark.py
sebhrusen: I think you can extract this logic in a dedicated method in `file.py` for clarity (it's just mutating `dataset` after all), and if you just support CVS right now, then please apply it only there.

frameworks/AutoGluonTS/setup.sh
Innixma: One difference @limpbot, I install MXNet in addition to the other normal dependencies for TimeSeries, since it isn't a default install. We can simply install MXNet by default for now, hopefully it won't cause issues.

@sebhrusen The one concern is if AutoGluon becomes too monolithic of an install, we may want to consider having separate install logic for the different submodules that are unrelated to eachother (for example, timeseries doesn't need vision and text modules, tabular doesn't need timeseries module). Probably not needed now, but something to keep in mind since AutoGluon covers more data types/domains than most AutoML systems and that comes with many dependencies.

sebhrusen: @Innixma I understand the concern.
for better encapsulation and to allow you to reuse code easily in benchmarks when using different submodules, I'd still advise to keep a single `AutoGluon` folder.
You can then provide different flavors for the setup just in the framework definition using the `setup_env` syntax:
```yaml
AutoGluon_TimeSeries
  extends: AutoGluon
  setup_env:
    MODULE: timeseries
    VAR: string
```
this makes the 2 variables `MODULE` and `VAR` directly available in `setup.sh` (right after the call to `. ${HERE}/../shared/setup.sh ${HERE} true`) and allows you to customize the setup: you may already be using this for dev environment.

Also, thanks to the definitions hierarchy (`extends: AutoGluon`), maybe we can later tweak the results to make it appear as just `AutoGluon`, or we can add a notion of group, whatever...


This may not be perfect when you switch frequently between definitions, but for now, I'd like to keep the framework folders to a minimum.
I agree with you that we probably need to start thinking about distinguishing the setup/exec for different kind of tasks, ideally it should be smooth and not even require the additional definition above: for a given type of task, the framework should be able to tell early if it can handle it, if not it could try to apply some additional setup before replying, and if it's ready handle it, then it continues as before: but all of this is much easier to change if there's already one single folder in the first place.
I can create an issue for this, although I don't have huge time to dedicate to AMLB lately, but this may change in a couple of months.

limpbot: sounds good to me, how can I access the variables MODULE and VAR?

limpbot: got it :)

resources/benchmarks/ts.yaml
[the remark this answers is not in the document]
limpbot: so as predictions is a more general term, I suppose ```forecast_range_in_steps``` is sufficiently explicit?

amlb/benchmark.py
limpbot: so I added it to a dedicated method in `file.py` inside the ```FileLoader``` class.

Innixma:
I merged @limpbot's changes into this branch via his PR: https://github.com/Innixma/automlbenchmark/pull/7

@sebhrusen The branch should be ready for 2nd round of review.

sebhrusen (approved):
The changes to the core logic are much smaller now, which is what I mostly care about for this feature in its current state and scope as I don't want to prevent you from moving forward.

I think it will be interesting for us (cc: @PGijsbers) to start thinking about supporting new kind of tasks, and see how we can integrate this smoothly (mixins after restructuring code?). Maybe even some kind of plugin logic (I have a PoC PR allowing the user to plug custom code in various places, mainly thinking about data loading, result metrics, and whatever the framework may need.
If you have any idea on your side, feel free to make suggestions in the https://github.com/openml/automlbenchmark/discussions or contribute directly.

Thanks a lot for this @limpbot and @Innixma

frameworks/AutoGluon/setup.sh
```diff
@@ -36,4 +37,10 @@ else
...
 fi
 
-PY -c "from autogluon.tabular.version import __version__; print(__version__)" >> "${HERE}/.setup/installed"
+if [[ ${MODULE} == "timeseries" ]]; then
+    PY -c "from autogluon.tabular.version import __version__; print(__version__)" >> "${HERE}/.setup/installed"
+    # TODO: GPU version install
+    PIP install "mxnet<2.0"
+else
+    PY -c "from autogluon.timeseries.version import __version__; print(__version__)" >> "${HERE}/.setup/installed"
+fi
```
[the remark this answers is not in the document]
limpbot: Oh yes, that got mixed-up. I don't think the versions should ever differ, but to be safe I will correct it and add it in a future pull request. Thank you for the reviews and merge @sebhrusen!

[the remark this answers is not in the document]
sebhrusen: default autogluon setup looks broken: see https://github.com/openml/automlbenchmark/actions/runs/3199648554/jobs/5225661120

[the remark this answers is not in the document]
sebhrusen: I can merge only once the default setup works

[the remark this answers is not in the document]
limpbot: okay thanks for the pointer, most likely because of the mixed-up version call. I am taking a look at it.

Innixma:
Thanks @sebhrusen for the detailed review! @limpbot has addressed some final comments in the latest update, which should also fix the autogluon.tabular error you mentioned.

PGijsbers:
> I think it will be interesting for us (cc: @PGijsbers) to start thinking about supporting new kind of tasks

I missed the "mention" ping (just thought it said a "subscribed"), sorry I didn't check earlier. Definitely, I want to first wait for the JMLR reviews and finish "that part of the project", but creating a more flexible environment for people to add new types of tasks would be a great next thing that invites more people to use (and extend) the benchmark tool.

Thanks Innixma and Limpbot for your contribution 🎉

PR #495, opened 2022-09-17
Fixed user and sudo under docker

alanwilter:
This fixed several issues I created with my previous RP re `docker.py`.

Now it runs under $USER, grant permissions and allow `sudo`.

### DIFF
--- a/amlb/runners/docker.py
+++ b/amlb/runners/docker.py
@@ -9,10 +9,9 @@ import os
 import re
 
 from ..benchmark import _setup_dir_
-from ..resources import config as rconfig, get as rget
+from ..resources import config as rconfig
 from ..utils import dir_of, run_cmd, str_digest, str_sanitize, touch
 from .container import ContainerBenchmark
-
 
 log = logging.getLogger(__name__)
 
@@ -31,17 +30,23 @@ class DockerBenchmark(ContainerBenchmark):
         """
         super().__init__(framework_name, benchmark_name, constraint_name)
         self._custom_image_name = rconfig().docker.image
+        self.user = os.getlogin()
+        self.userid = os.getuid()
+        self.usergid = os.getgid()
         self.minimize_instances = rconfig().docker.minimize_instances
-        self.container_name = 'docker'
+        self.container_name = "docker"
         self.force_branch = rconfig().docker.force_branch
-        self.custom_commands = self.framework_module.docker_commands(
-            self.framework_def.setup_args,
-            setup_cmd=self.framework_def._setup_cmd
-        ) if hasattr(self.framework_module, 'docker_commands') else ""
+        self.custom_commands = (
+            self.framework_module.docker_commands(
+                self.framework_def.setup_args, setup_cmd=self.framework_def._setup_cmd
+            )
+            if hasattr(self.framework_module, "docker_commands")
+            else ""
+        )
 
     @property
     def _script(self):
-        return os.path.join(self._framework_dir, _setup_dir_, 'Dockerfile')
+        return os.path.join(self._framework_dir, _setup_dir_, "Dockerfile")
 
     def _start_container(self, script_params=""):
         """Implementes the container run method"""
@@ -50,11 +55,12 @@ class DockerBenchmark(ContainerBenchmark):
         custom_dir = rconfig().user_dir
         for d in [in_dir, out_dir, custom_dir]:
             touch(d, as_dir=True)
-        script_extra_params = "--session="  # in combination with `self.output_dirs.session` usage below to prevent creation of 2 sessions locally
+        # in combination with `self.output_dirs.session` usage below to prevent creation of 2 sessions locally
+        script_extra_params = "--session="
         inst_name = f"{self.sid}.{str_sanitize(str_digest(script_params))}"
         cmd = (
             "docker run --name {name} {options} "
-            '-u "$(id -u $USER):$(id -g $USER)" '
+            f'-u "{self.userid}:{self.usergid}" '
             "-v {input}:/input -v {output}:/output -v {custom}:/custom "
             "--rm {image} {params} -i /input -o /output -u /custom -s skip -Xrun_mode=docker {extra_params}"
         ).format(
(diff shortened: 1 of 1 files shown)


amlb/runners/docker.py
```diff
@@ -111,17 +119,20 @@ def _upload_image(self, image):
 
     def _generate_script(self, custom_commands):
         docker_content = """FROM ubuntu:18.04
-
 ENV DEBIAN_FRONTEND noninteractive
-RUN apt-get update
-RUN apt-get -y install apt-utils dialog locales
+RUN apt-get update && apt-get install -y tzdata
+RUN apt-get -y install apt-utils dialog locales sudo
```
sebhrusen, line 124: please add `tzdata` here, looks more natural than on the previous line

sebhrusen:
@alanwilter, to be clear, this fixes issues introduced with https://github.com/openml/automlbenchmark/pull/464, correct?

Could you please detail a bit more the kind of issues that were introduced by this? For us to understand the changes, for other users that may face the same problems… thanks.

amlb/runners/docker.py
```diff
@@ -9,7 +9,7 @@
 import re
 
 from ..benchmark import _setup_dir_
-from ..resources import config as rconfig, get as rget
+from ..resources import config as rconfig
```
alanwilter, line 12: rget is not used anywhere here and it's not imported from here.

amlb/runners/docker.py
```diff
@@ -31,17 +30,23 @@ def __init__(self, framework_name, benchmark_name, constraint_name):
         """
         super().__init__(framework_name, benchmark_name, constraint_name)
         self._custom_image_name = rconfig().docker.image
+        self.user = os.getlogin()
+        self.userid = os.getuid()
+        self.usergid = os.getgid()
         self.minimize_instances = rconfig().docker.minimize_instances
-        self.container_name = 'docker'
+        self.container_name = "docker"
```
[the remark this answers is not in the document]
alanwilter: > @alanwilter, to be clear, this fixes issues introduced with #464, correct?
> 
> Could you please detail a bit more the kind of issues that were introduced by this? For us to understand the changes, for other users that may face the same problems… thanks.

Yes, I found out that when running docker and docker line `RUN $PY {script} {framework} -s only` was called, some frameworks need to install ubuntu packages etc. and this was failing.

Now all runs as the local `user` issuing the docker cmd and will save all for that `user`, and this `user` is enabled to run `sudo` inside the docker, without password.

The other issue addressed was that, again, when docker running `RUN $PY {script} {framework} -s only`, for some frameworks, directive `ENV DEBIAN_FRONTEND noninteractive` was not working to avoid `tzdata` to enter in the interactive mode. So, I found better to install it at the first opportunity. You can rearrange it as you wish.

I found this issue when debugging `autoxgboost`, which installs 

```bash
  SUDO apt-get install -y r-base r-base-dev
  SUDO apt-get install -y libgdal-dev libproj-dev
  SUDO apt-get install -y libssl-dev libcurl4-openssl-dev
  SUDO apt-get install -y libcairo2-dev libudunits2-dev
```
and one of these pkgs has `tzdata` dependency.

amlb/runners/docker.py
alanwilter: Check if it's better now.

amlb/runners/docker.py
```diff
@@ -54,7 +57,7 @@ def _start_container(self, script_params=""):
         inst_name = f"{self.sid}.{str_sanitize(str_digest(script_params))}"
         cmd = (
             "docker run --name {name} {options} "
-            '-u "$(id -u $USER):$(id -g $USER)" '
+            f'-u "{self.userid}:{self.usergid}" '
```
sebhrusen, line 60: for Windows users, docker mode is often the only mode available to run most of the frameworks, so we need to ensure that this still works on Win.

@PGijsbers I now wonder how the original `$(id -u $USER)` calls worked on Win, are we requiring Bash on Windows? In this case, how does the Py commands `os.getuid()` and `os.getgid()` behave on Win?

amlb/runners/docker.py
```diff
@@ -152,6 +159,9 @@ def _generate_script(self, custom_commands):
 ADD . /bench/
 
 RUN (grep -v '^\\s*#' | xargs -L 1 $PIP install --no-cache-dir) < requirements.txt
+RUN chown -R {username}.{username} /bench
+RUN chown -R {username}.{username} /home/{username}
```
sebhrusen, line 163:
```suggestion
RUN chown -R {username}:{username} /bench
RUN chown -R {username}:{username} /home/{username}
```

`.` is a deprecated syntax

sebhrusen (approved):
looks good to me, mainly concerned about the behaviour of `os.getXXid()` on Windows.
Waiting for feedback from @PGijsbers about this, then I think we can merge.

amlb/runners/docker.py
alanwilter: I can't test that, but Windows (and Mac), don't mess up root with user, i.e., you run a docker a root and save files, those files will belong to original windows/mac user and not to the root, like it happens in Linux.
Perhaps, just wrap these lines with a `if linux` ...

PGijsbers: Seems reasonable to me

PGijsbers:
@alanwilter 
I picked up this PR since it went stale. I made sure Windows support was not broken.
Additionally, I added HDF5 libraries required to install `tables`. After the patches building and running in docker mode works for me on both MacOS and Windows. Could you confirm that after these patches the original program you fixed is still resolved with your setup?

alanwilter:
You're welcome. I still need to find time to check the out https://github.com/openml/automlbenchmark/pull/495#issuecomment-1453378049
But I have no idea when I will do that.

PGijsbers:
Since there was no response for 3 weeks, I wasn't sure if you were still interested in trying so I went ahead and merged. If you do ever find the time to check, we would appreciate a quick confirmation :) But there's no rush, we'll operate under the assumption the problem is fixed (because, as far as I can tell, it is).

PGijsbers:
This introduced an issue when building public images: the image is set up for the builder, but the user will start the container under a different uid, which leads to permission errors. I think we should maybe remove any user information from the docker container, and instead create/assign permissions on startup with something like:

```bash
set -e

if [ $UID = 0 ]; then
  echo "Docker started as root, not changing file permissions."
  exit 0
fi

user_id=$UID
echo "root" | su -c "adduser --disabled-password --gecos '' -uid 502 amlb"
echo "root" | su -c "echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers"
chown -R amlb:amlb /home/amlb
chown -R amlb:amlb /bench
echo "root" | su -c "passwd -d amlb"
su "amlb"
```

PGijsbers:
There's probably cleanier ways to set this up (though whatever UID is inserted is unlikely to be in de sudo file), but my experience lacks here.

PGijsbers:
@alanwilter something like that should still solve your issue, right?

alanwilter:
Unfortunately I'm not working on related projects anymore so I'd say do what is better for you guys. If I'll come back one day then I will check and test all that.

PGijsbers:
Thanks for letting us know!

PR #507, opened 2022-12-04
Metric fix, added GluonTS for Baselines, added AutoPytorchTS, added Local Code Option

limpbot:
Hi @sebhrusen, @PGijsbers,
Strongly hoping to get this PR accepted by Christmas, if you could review it in time, so I can add the required changes, I would highly appreciate it.

The following changes exist.

Renaming forecast_range_in_steps to forecast_horizon_in_steps, because it is much more popular in the research field.

Fixed metrics calculation. A) Changed y_past_period_error to seasonal error, because this is the normally used for MASE. Specifying seasonality means fixed metric calculation. B) Renamed ncrps (Normalized Continuous Ranked Probability Score) to mwql (Mean Weighted Quantile Loss), because that is what is actually calculated. This is an approximation of ncrps. C) Added item_id column to results, to be able to calculate itemwise_mean. 

Added GluonTS framework, because it allows to calculate multiple timeseries baselines. GluonTS_* 1) Prophet, 2) DeepAR, 3) NBEATS, 4) NPTS, 5) SeasonalNaive, 6) SimpleFeedForward, 7) MQCNN, 8) MQRNN, 9) TFT, 10) ARIMA, 11) ETS, 12) STL-AR, 13) Theta.

Added AutoPyTorchTS for comparison. Only timeseries for now, tabular implementation missing.

Added improvements for AutoGluonTS. A) Do not deepcopy the dataset to save memory. B) Use correct module to retrieve the AutoGluon version. 

Added "local" framework configuration file for testing local code. This is achieved by forwarding the USER_DIR to the setup.sh script leveraging the environment variables.

Kind regards,
Leo

FYI: @Innixma @canerturkmen @gidler


frameworks/AutoPyTorch/exec.py
```diff
@@ -0,0 +1,192 @@
...
+    id_column = dataset.id_column
+    prediction_length = dataset.forecast_horizon_in_steps
+    target_column = dataset.target.name
+    seasonality = dataset.seasonality
+
+    # data and metric imports
+    # from sktime.datasets import load_longley
+    # targets, features = load_longley()
+    # targets: pandas.core.series.Series (N, ), index = timestamp
+    # features: pandas.core.frame.DataFrame (N, D) , index = timestamp
```
PGijsbers, line 35: Please remove the commented out code (also in other files like `setup.sh`).

frameworks/AutoPyTorch/setup.sh
```diff
@@ -0,0 +1,101 @@
...
+#     if [[ "${OS}" == "Ubuntu 18.04" ]]; then
+#         SUDO apt install -y software-properties-common
+#         SUDO apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys E298A3A825C0D65DFD57CBB651716619E084DAB9
+#         SUDO add-apt-repository 'deb https://cloud.r-project.org/bin/linux/ubuntu bionic-cran35/'
+#         SUDO apt-get -y update
+#     fi
+#     apt install -y r-base
+#     SUDO R -e 'install.packages(c("forecast", "nnfor"), repos="https://cloud.r-project.org")'
+#     # ---------------------
+    SUDO apt-get install -y python3-opencv
```
PGijsbers, line 76: Just to double-check, is `python3-opencv` needed (for time series)? Why isn't it installed on the other OS?

resources/frameworks.yaml
```diff
@@ -195,6 +203,17 @@ TPOT:
 ### TimeSeries AutoML frameworks ###
 ####################################
 
+AutoPyTorchTS:
```
PGijsbers, line 206: @sebhrusen should we work with a separate framework file for timeseries frameworks/baselines (even if they might reference the same `frameworks/` implementation with a different setting)? I think I would prefer that.

resources/frameworks_local.yaml
```diff
@@ -0,0 +1,43 @@
+---
```
PGijsbers, line 1: Please remove this file.

resources/config.yaml
```diff
@@ -54,7 +54,7 @@ benchmarks:                     # configuration namespace for the benchmarks def
     binary: ['auc', 'logloss', 'acc', 'balacc']     # available metrics: auc (AUC), acc (Accuracy), balacc (Balanced Accuracy), pr_auc (Precision Recall AUC), logloss (Log Loss), f1, f2, f05 (F-beta scores with beta=1, 2, or 0.5), max_pce, mean_pce (Max/Mean Per-Class Error).
     multiclass: ['logloss', 'acc', 'balacc']        # available metrics: same as for binary, except auc, replaced by auc_ovo (AUC One-vs-One), auc_ovr (AUC One-vs-Rest). AUC metrics and F-beta metrics are computed with weighted average.
     regression: ['rmse', 'r2', 'mae']               # available metrics: mae (Mean Absolute Error), mse (Mean Squared Error), msle (Mean Squared Logarithmic Error), rmse (Root Mean Square Error), rmsle (Root Mean Square Logarithmic Error), r2 (R^2).
-    timeseries: ['mase', 'mape', 'smape', 'rmse', 'mse', 'nrmse', 'wape', 'ncrps']
+    timeseries: ['mase', 'mape', 'smape', 'rmse', 'mse', 'nrmse', 'wape', 'mwql', 'mean_seasonal_error']
```
PGijsbers, line 57: Please add information about the metric names similar to that for other task types.

amlb/benchmark.py
```diff
@@ -116,7 +116,9 @@ def setup(self, mode: SetupMode):
 
         log.info("Setting up framework {}.".format(self.framework_name))
 
-        self._write_setup_env(self.framework_module.__path__[0], **dict(self.framework_def.setup_env))
+        framework_def_plus_user_dir_dict = self.framework_def.setup_env
+        framework_def_plus_user_dir_dict['USER_DIR'] = rconfig().user_dir
+        self._write_setup_env(self.framework_module.__path__[0], **dict(framework_def_plus_user_dir_dict))
```
sebhrusen, line 121: if you just need to add some constants, please add them to `_write_setup_env` definition: it already sets some hardcoded defaults

frameworks/AutoGluon/__init__.py
```diff
@@ -53,8 +65,23 @@ def run_autogluon_timeseries(dataset: Dataset, config: TaskConfig):
...
+    try:
+        with open(fpath, 'r') as f:
+            for line in f:
+                if line.startswith('#') or not line.strip():
+                    continue
+                key, value = line.strip().split('=', 1)
+                env_vars[key] = value
+    except OSError:
+        print(f'Could not open/read file {fpath}')
+    return env_vars
```
sebhrusen, line 87: I'd like to avoid starting to see multiply implementations of this.
Please move to `benchmark.py` as `_read_setup_env()` (after `_write_setup_env`)
and then add to `TaskConfig`:
```python
class BenchmarkTask:

    def __init__(self, benchmark: Benchmark, task_def, fold):
        ....
        self.task_config = TaskConfig(
            ...
            output_dir=benchmark.output_dirs.session,
            setup_env=benchmark._read_setup_env(),
        )
```
then you can access those setup vars directly in `__init__` or even in `exec.py`

resources/frameworks_local.yaml
sebhrusen: @limpbot those custom definitions files should ideally reside in the `user_dir`, or you can rename it locally to `my_frameworks.yaml` to have them ignored by git in this project (or create a `.gitignore` in your user home folder)

resources/frameworks.yaml
sebhrusen: yes, makes sense until we have plans to officially include TS in the benchmark paper

sebhrusen (changes requested):
fyi, only reviewed common files for now

resources/config.yaml
```diff
@@ -33,7 +33,7 @@ frameworks:              # configuration namespace for the frameworks definition
     - '{root}/resources/frameworks.yaml'
   root_module: frameworks     # the default python module under which the frameworks modules are defined.
   allow_duplicates: false     # if true, the last definition is used.
-  tags: ['stable', 'latest', '2020Q2', '2021Q3']  # the list of supported tags when looking up frameworks:
+  tags: ['stable', 'latest', 'local', '2020Q2', '2021Q3']  # the list of supported tags when looking up frameworks:
```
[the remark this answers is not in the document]
limpbot: To enable the user to evaluate a local implementation of an AutoML framework. Basically, to evaluate a change without pushing it to github. Especially time saving if the framework is installed in editable mode (`pip install -e`). In case you want us to use this only internally, I can of course reverse the changes to the affected config files.

resources/config.yaml
limpbot: got it

amlb/benchmark.py
limpbot: got it

frameworks/AutoPyTorch/exec.py
limpbot: got it

frameworks/AutoPyTorch/setup.sh
limpbot: This dependency chain is autoPyTorch[forecasting] ->imgaug>=0.4.0->opencv-python.

limpbot: At least one dynamic library is not automatically installed in Ubuntu. Otherwise, we get the following error `ImportError: libGL.so.1: cannot open shared object file: No such file or directory`
In AL2 this error does not appear.

frameworks/AutoGluon/__init__.py
limpbot: got it

resources/frameworks_local.yaml
limpbot: Thank you for the pointer. Please read my arguments in the next thread and let me know, if you still want me to keep these configurations locally.

resources/frameworks.yaml
limpbot: One more suggestion, we have a separate `framework_timeseries.yaml`, and leave the AutoML timeseries frameworks in the `framework_latest.yaml` file?

limpbot: Okay, this sounds good to me. Besides that, are you fine with the configurations of `AutoGluonTS` and `AutoPyTorchTS` in the `framework_latest.yaml`?

amlb/results.py
```diff
@@ -671,67 +671,104 @@ class TimeSeriesResult(RegressionResult):
...
+        if np.sum(np.abs(np.mean(unique_item_ids_counts) - unique_item_ids_counts)) != 0.:
+            msg = f'Error: Predicted sequences have different lengths {unique_item_ids_counts}.'
+            raise ValueError(msg)
+        self.prediction_length = unique_item_ids_counts[0]
+        self.unique_item_ids = self.item_ids.copy()[::self.prediction_length]
+
+        self.dtype = float
+        self.truth = self.truth.astype(self.dtype, copy=False)
+        self.pred_mean = self.pred_mean.astype(self.dtype, copy=False)
+        self.quantiles = self.quantiles.astype(self.dtype, copy=False)
```
PGijsbers, line 696: Why are these separate from the assignments above? Is this more performant than keeping them in a dataframe? To me it seems like they are mostly just used as aliases. Doesn't the dataframe typing generally provide more readable code? E.g.,
```python
     def itemwise_mean(self, values):
         return np.array([np.mean(values[self.item_ids == unique_item_id]) for unique_item_id in self.unique_item_ids], dtype=self.dtype)

     def itemwise_select_first(self, values):
         return values.copy()[::self.prediction_length]
```
to
```python
     def itemwise_mean(self, column):
         return self.df.groupby("item_ids")[column].mean()

     def itemwise_select_first(self, column):
         return self.df.groupby("item_ids")[column].first()
```

frameworks/AutoPyTorch/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+        raise ValueError("Framework `AutoPyTorch` does expect timeseries tasks for now.")
+
+    from frameworks.shared.caller import run_in_venv
+    dataset = deepcopy(dataset)
+    if not hasattr(dataset, 'timestamp_column'):
+        dataset.timestamp_column = None
+    if not hasattr(dataset, 'id_column'):
+        dataset.id_column = None
+    if not hasattr(dataset, 'forecast_horizon_in_steps'):
+        raise AttributeError("Unspecified `forecast_horizon_in_steps`.")
```
PGijsbers, line 22: Maybe move this before the `deepcopy` call.

frameworks/AutoPyTorch/exec.py
```diff
@@ -0,0 +1,178 @@
...
+
+    log.info(f'Target name {dataset.target.name}. Forecast horizon {forecast_horizon}.')
+    log.info(f'There are {len(list(dataset_test.groupby(dataset.id_column)))} sequences in the test dataset.')
+    log.info(f'There are {len(list(dataset_train.groupby(dataset.id_column)))} sequences in the train dataset.')
+
+    item_ids =  dataset.test.X[dataset.id_column].unique()
+    items_indices_timestamp = [dataset.test.X[dataset.test.X[dataset.id_column] == item_id].set_index(dataset.timestamp_column).index for item_id in item_ids[:100]]
+    items_freqs = [item_id_indices_timestamp.freq or item_id_indices_timestamp.inferred_freq for item_id_indices_timestamp in items_indices_timestamp]
+    items_freqs_unique = set(items_freqs)
+    if not len(items_freqs_unique) == 1:
```
PGijsbers, line 55:
```suggestion
    if len(items_freqs_unique) != 1:
```

frameworks/AutoPyTorch/exec.py
```diff
@@ -0,0 +1,178 @@
...
+        api = TimeSeriesForecastingTask()
+
+        # Search for an ensemble of machine learning algorithms
+        api.search(
+            X_train=X_train,
+            y_train=y_train,
+            X_test=X_test,
+            optimize_metric=eval_metric,
+            n_prediction_steps=forecast_horizon,
+            memory_limit=30 * 1024,  # Currently, forecasting models use much more memories
```
PGijsbers, line 71: I have a note on this in `GluonTS/exec.py`, please explicitly log warnings where you deviate constraints from those defined in the configuration (time, memory, cpu, metric).

frameworks/AutoPyTorch/exec.py
```diff
@@ -0,0 +1,178 @@
...
+        median_mae='median_MAE_forecasting',
+        mape='mean_MAPE_forecasting',
+        median_mape='median_MAPE_forecasting',
+        mse='mean_MSE_forecasting',
+        median_mse='median_MSE_forecasting'
+        #smape=None,
+        #rmse=None,
+    )
+
+    eval_metric = metrics_mapping[config.metric] if config.metric in metrics_mapping else None
```
PGijsbers, line 143:
```suggestion
    eval_metric = metrics_mapping.get(config.metric)
```

PGijsbers (changes requested):
I did a pass, but unfortunately I couldn't run it (yet). I am running into some installation issues, but that's due to my setup. 

In addition to the other comments, please also add some documentation about the expected dataset and prediction formatting.

frameworks/AutoPyTorch/setup.sh
```diff
@@ -0,0 +1,101 @@
...
+
+
+if [[ "${OS}" == "Amazon Linux 2" ]]; then
+    # TOPIC: update system
+    echo "updating system..."
+    SUDO yum clean all
+    SUDO rm -rf /var/cache/yum
+    SUDO yum -y update
+
+#     # TOPIC: install readline-devel (for rpy2)
```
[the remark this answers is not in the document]
limpbot: done

frameworks/GluonTS/exec.py
```diff
@@ -0,0 +1,233 @@
...
+                  truth=truth_only,
+                  probabilities=None,
+                  probabilities_labels=None,
+                  target_is_encoded=False,
+                  models_count=1, #num_models_trained,
+                  training_duration=training.duration,
+                  predict_duration=predict.duration,
+                  optional_columns=optional_columns)
+
+def estimate_predictor(model, prediction_length, freq, dataset_train_gluonts, time_limit):
```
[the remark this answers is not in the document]
limpbot: You are mostly right, I added the warnings for ignored performance metric, memory and core constraints. The training time constraint is not ignored. The models, that do not use the training time constraint as parameters, are actually never trained, but tune their parameters only depending on the observations in the test dataset.

frameworks/AutoPyTorch/exec.py
limpbot: Done

amlb/results.py
```diff
@@ -671,67 +671,104 @@ class TimeSeriesResult(RegressionResult):
...
+        if len(np.where(self.quantiles_probs == 0.5)[0]) > 0:
+            self.pred_median = self.quantiles[:, np.where(self.quantiles_probs == 0.5)[0][0]]
```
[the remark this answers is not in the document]
limpbot: Can you give more information, about what command exactly does not run? The use of this line is to select the column, that represents the 50% quantiles (=median).

frameworks/GluonTS/exec.py
```diff
@@ -0,0 +1,233 @@
...
+                w = csv.writer(f)
+                w.writerow(agg_metrics.keys())
+                w.writerow(agg_metrics.values())
+        if 'item_metrics' in artifacts:
+            item_metrics.to_csv(os.path.join(metrics_dir, 'item_metrics.csv'))
+    except Exception:
+        log.warning("Error when saving artifacts.", exc_info=True)
+
+def calc_seasonal_error(dataset_test, id_column, target_column, prediction_length, seasonality):
+    """Calculates the sesonal error for the test dataset and repeates it for each element in the forecast sequence.
```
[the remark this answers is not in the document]
limpbot: Done

amlb/datasets/file.py
```diff
@@ -143,31 +143,35 @@ def extend_dataset_with_timeseries_config(self, dataset, dataset_config):
...
+        dataset.seasonality=int(dataset_config['seasonality'])
 
         train_seqs_lengths = dataset.train.X.groupby(dataset.id_column).count()
         test_seqs_lengths = dataset.test.X.groupby(dataset.id_column).count()
-        forecast_range_in_steps_mean_diff_train_test = int((test_seqs_lengths - train_seqs_lengths).mean())
-        forecast_range_in_steps_max_min_train_test = int(min(int(test_seqs_lengths.min()), int(train_seqs_lengths.min()))) - 1
-        if not dataset.forecast_range_in_steps == forecast_range_in_steps_mean_diff_train_test:
-            msg = f"Warning: Forecast range {dataset.forecast_range_in_steps}, does not equal mean difference between test and train sequence lengths {forecast_range_in_steps_mean_diff_train_test}."
+        forecast_horizon_in_steps_mean_diff_train_test = int((test_seqs_lengths - train_seqs_lengths).mean())
+        forecast_horizon_in_steps_max_min_train_test = int(min(int(test_seqs_lengths.min()), int(train_seqs_lengths.min()))) - 1
```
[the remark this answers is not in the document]
limpbot: This pandas `.count()` method returns either a `pandas.Series` or a `pandas.Dataframe`. But I have cleaned it now by using `.count().values` to retrieve a `numpy.array` object and calling its method `.min()`.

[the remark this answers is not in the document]
limpbot: Further, you were right the `max/min` was confusing. Removed some of the warnings and changed some errors to warnings, when they are more framework specific.

amlb/datasets/file.py
```diff
@@ -143,31 +143,35 @@ def extend_dataset_with_timeseries_config(self, dataset, dataset_config):
...
+        dataset.forecast_horizon_in_steps=int(dataset_config['forecast_horizon_in_steps'])
+        dataset.seasonality=int(dataset_config['seasonality'])
 
         train_seqs_lengths = dataset.train.X.groupby(dataset.id_column).count()
         test_seqs_lengths = dataset.test.X.groupby(dataset.id_column).count()
-        forecast_range_in_steps_mean_diff_train_test = int((test_seqs_lengths - train_seqs_lengths).mean())
-        forecast_range_in_steps_max_min_train_test = int(min(int(test_seqs_lengths.min()), int(train_seqs_lengths.min()))) - 1
-        if not dataset.forecast_range_in_steps == forecast_range_in_steps_mean_diff_train_test:
-            msg = f"Warning: Forecast range {dataset.forecast_range_in_steps}, does not equal mean difference between test and train sequence lengths {forecast_range_in_steps_mean_diff_train_test}."
+        forecast_horizon_in_steps_mean_diff_train_test = int((test_seqs_lengths - train_seqs_lengths).mean())
```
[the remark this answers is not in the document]
limpbot: Done

amlb/results.py
limpbot: These are additional functions, which we call on the difference between the `numpy.array`'s of e.g. `self.pred_mean` and `self.truth`. Improved the solution for readability
`    def itemwise_mean(self, values):
        return pd.DataFrame(np.stack([self.item_ids, values], axis=1), columns=['item_id', 'value']).groupby('item_id').mean()['value'].values

    def itemwise_select_first(self, values):
        return pd.DataFrame(np.stack([self.item_ids, values], axis=1), columns=['item_id', 'value']).groupby('item_id').first()['value'].values`

frameworks/AutoPyTorch/__init__.py
limpbot: Done

frameworks/AutoPyTorch/exec.py
limpbot: Thanks, for the pointer. Done

frameworks/AutoPyTorch/exec.py
limpbot: Done

amlb/results.py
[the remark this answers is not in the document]
limpbot:
```suggestion
        if 0.5 in self.quantiles_probs:
            self.pred_median = self.quantiles[:, self.quantiles_probs == 0.5].squeeze()
```

amlb/results.py
```diff
@@ -671,67 +671,104 @@ class TimeSeriesResult(RegressionResult):
...
+            msg = f'Error: Predicted sequences have different lengths {unique_item_ids_counts}.'
+            raise ValueError(msg)
+        self.prediction_length = unique_item_ids_counts[0]
+        self.unique_item_ids = self.item_ids.copy()[::self.prediction_length]
+
+        self.dtype = float
+        self.truth = self.truth.astype(self.dtype, copy=False)
+        self.pred_mean = self.pred_mean.astype(self.dtype, copy=False)
+        self.quantiles = self.quantiles.astype(self.dtype, copy=False)
+        if len(np.where(self.quantiles_probs == 0.5)[0]) > 0:
```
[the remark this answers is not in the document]
limpbot: Thx, done.

amlb/results.py
```diff
@@ -671,67 +671,104 @@ class TimeSeriesResult(RegressionResult):
...
+
+        self.truth = self.df['truth'].values if self.df is not None else None
+        self.pred_mean = self.df['predictions'].values if self.df is not None else None
+
+        self.seasonal_error = self.df['seasonal_error'].values
+        self.quantiles_probs = np.array(list(float(quantile_prob_str) for quantile_prob_str in self.quantiles_probs_str))
+        self.quantiles = self.df.filter(self.quantiles_probs_str).values
+        self.item_ids = self.df['item_id'].values
+        _, unique_item_ids_counts = np.unique(self.item_ids, return_counts=True)
+        if np.sum(np.abs(np.mean(unique_item_ids_counts) - unique_item_ids_counts)) != 0.:
```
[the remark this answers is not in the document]
limpbot: True, thanks bringing light into this dark line. Done.

amlb/results.py
```diff
@@ -671,67 +671,104 @@ class TimeSeriesResult(RegressionResult):
...
+        if not all(x in self.df.columns for x in ['truth', 'predictions', 'item_id', 'seasonal_error']) or len(self.quantiles_probs_str) == 0:
+            msg=f'Missing columns for calculating time series metrics. Given columns are {list(self.df.columns)}.'
+            raise ValueError(msg)
+
+        self.truth = self.df['truth'].values if self.df is not None else None
+        self.pred_mean = self.df['predictions'].values if self.df is not None else None
+
+        self.seasonal_error = self.df['seasonal_error'].values
+        self.quantiles_probs = np.array(list(float(quantile_prob_str) for quantile_prob_str in self.quantiles_probs_str))
+        self.quantiles = self.df.filter(self.quantiles_probs_str).values
```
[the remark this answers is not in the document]
limpbot: Done.

amlb/results.py
```diff
@@ -671,67 +671,104 @@ class TimeSeriesResult(RegressionResult):
...
+        self.quantiles_probs_str = list(column for column in self.df.columns if column.startswith('0.'))
+        if not all(x in self.df.columns for x in ['truth', 'predictions', 'item_id', 'seasonal_error']) or len(self.quantiles_probs_str) == 0:
+            msg=f'Missing columns for calculating time series metrics. Given columns are {list(self.df.columns)}.'
+            raise ValueError(msg)
+
+        self.truth = self.df['truth'].values if self.df is not None else None
+        self.pred_mean = self.df['predictions'].values if self.df is not None else None
+
+        self.seasonal_error = self.df['seasonal_error'].values
+        self.quantiles_probs = np.array(list(float(quantile_prob_str) for quantile_prob_str in self.quantiles_probs_str))
```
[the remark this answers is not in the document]
limpbot: Done.

amlb/results.py
```diff
@@ -671,67 +671,104 @@ class TimeSeriesResult(RegressionResult):
...
-        self.predictions = self.predictions.astype(float, copy=False)
-        self.quantiles = self.quantiles.astype(float, copy=False)
-        self.y_past_period_error = self.y_past_period_error.astype(float, copy=False)
+        self.quantiles_probs_str = list(column for column in self.df.columns if column.startswith('0.'))
+        if not all(x in self.df.columns for x in ['truth', 'predictions', 'item_id', 'seasonal_error']) or len(self.quantiles_probs_str) == 0:
+            msg=f'Missing columns for calculating time series metrics. Given columns are {list(self.df.columns)}.'
+            raise ValueError(msg)
+
+        self.truth = self.df['truth'].values if self.df is not None else None
+        self.pred_mean = self.df['predictions'].values if self.df is not None else None
```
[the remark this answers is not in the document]
limpbot: True, done.

amlb/results.py
```diff
@@ -671,67 +671,104 @@ class TimeSeriesResult(RegressionResult):
...
-        self.quantiles = self.df.iloc[:, 2:-1].values
-        self.quantiles_probs = np.array([float(q) for q in self.df.columns[2:-1]])
-        self.truth = self.truth.astype(float, copy=False)
-        self.predictions = self.predictions.astype(float, copy=False)
-        self.quantiles = self.quantiles.astype(float, copy=False)
-        self.y_past_period_error = self.y_past_period_error.astype(float, copy=False)
+        self.quantiles_probs_str = list(column for column in self.df.columns if column.startswith('0.'))
+        if not all(x in self.df.columns for x in ['truth', 'predictions', 'item_id', 'seasonal_error']) or len(self.quantiles_probs_str) == 0:
+            msg=f'Missing columns for calculating time series metrics. Given columns are {list(self.df.columns)}.'
+            raise ValueError(msg)
```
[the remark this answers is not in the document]
limpbot: Thank you for the pointer. Done.

amlb/results.py
```diff
@@ -671,67 +671,104 @@ class TimeSeriesResult(RegressionResult):
...
-        self.truth = self.df['truth'].values if self.df is not None else None #.iloc[:, 1].values if self.df is not None else None
-        self.predictions = self.df['predictions'].values if self.df is not None else None #.iloc[:, -2].values if self.df is not None else None
-        self.y_past_period_error = self.df['y_past_period_error'].values
-        self.quantiles = self.df.iloc[:, 2:-1].values
-        self.quantiles_probs = np.array([float(q) for q in self.df.columns[2:-1]])
-        self.truth = self.truth.astype(float, copy=False)
-        self.predictions = self.predictions.astype(float, copy=False)
-        self.quantiles = self.quantiles.astype(float, copy=False)
-        self.y_past_period_error = self.y_past_period_error.astype(float, copy=False)
+        self.quantiles_probs_str = list(column for column in self.df.columns if column.startswith('0.'))
```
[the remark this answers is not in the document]
limpbot: Done.

amlb/results.py
```diff
@@ -671,67 +671,104 @@ class TimeSeriesResult(RegressionResult):
...
+        self.seasonal_error = self.seasonal_error.astype(self.dtype, copy=False)
 
         self.target = Feature(0, 'target', 'real', is_target=True)
         self.type = DatasetType.timeseries
 
+    def itemwise_mean(self, values):
+        return np.array([np.mean(values[self.item_ids == unique_item_id]) for unique_item_id in self.unique_item_ids], dtype=self.dtype)
+
+    def itemwise_select_first(self, values):
+        return values.copy()[::self.prediction_length]
```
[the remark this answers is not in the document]
limpbot: This line is removed by now anyways, but it was there to ensure that potential mutations after returning the `np.array` would not affect the original `np.array`.

PGijsbers:
Unfortunately, I did not find the time to actually try and run the code. I am running into HDF5 issues building docker containers (MacOS is not supported for GluonTS), but I don't think it's specific to this PR. 

 I am still missing documentation on the dataset formats and the introduced dataset meta-data (the attributes  `forecast_horizon_in_steps`, `id_column`, `timestamp_column`, `seasonality`). While some of them are somewhat self-explanatory, a brief description of what they are and which data formats are allowed for them is important for other people to be able to use it with their datasets.
 
Otherwise, if @sebhrusen OKs this then I am good with it too.

With this I am signing off for year. Happy holidays and see you in January :)

PGijsbers:
Thanks you very much for the contribution 🙏 🎉

shchur:
Hi @PGijsbers and @sebhrusen, thank you for the comments on the PR! Are there any remaining items that need to be addressed before the PR can be merged?

Innixma:
@sebhrusen @PGijsbers Would greatly appreciate if we can understand next steps for this PR. We are happy to address any further review comments you have.

PGijsbers:
At last, I found the time to have another look! I tried to run the provided example containerized in docker with the command:

`python runbenchmark.py autogluonts:timeseries timeseries test -m docker`

But unfortunately it currently fails on install. I first applied the fixes from #495 to make sure the regular setup still worked, which it did (`python runbenchmark.py autogluon test test -m docker`). I also added 'timeseries' as a valid framework tag to `config.yaml` to ensure the right version could be found, and cleared any local docker images of `automlbenchmark/autogluon`. Then, running the command fails with:

```
#25 102.2 ModuleNotFoundError: No module named 'autogluon.timeseries'
```
which I think originates from an earlier failure to install dependencies:
```
#25 30.99 ERROR: Could not find a version that satisfies the requirement catboost<0.25,>=0.23.0; extra == "all" (from autogluon-tabular[all]) (from versions: none)
```

I also tried fixing the version to an old AutoGluon which matches the period this review was requested (`0.6.1`), and ultimately get a similar error message:
```
#25 24.03 ModuleNotFoundError: No module named 'autogluon'
```
likely because of
```
#25 16.52 ERROR: No matching distribution found for ray[tune]<2.1,>=2.0; extra == "all"
```

The main underlying issue would be the outdated Python version (a concern already raised by Nick in #511), but in principle I don't understand why the old version wouldn't be able to install with simply 3.7-compatible versions. Full output [here](https://gist.github.com/PGijsbers/3bff3389a8d31dc366757fce7745fc25).

We will raise the Python versions, but I believe this should still work without it (at least for versions compatible with Py3.7).

Innixma:
@PGijsbers Interesting, that is an odd error. I'd say we wouldn't be very concerned with supporting v0.6.x AG given that v0.7 is released, so my recommendation is first updating the python version and seeing if v0.7.x works. If not, I can try seeing about fixing.

Let me know if there is a particular reason v0.6.x support is important for AMLB here.

Looking at the specific error, this looks very strange because the catboost version that is being specified is very old. It almost looks like it is trying to install an old version of AutoGluon (older than v0.6.x), but I don't know why that would be the case.

Looking at our old releases, the version that matches that catboost version range requirement is AutoGluon 0.1.0:

https://github.com/autogluon/autogluon/blob/0.1.0/tabular/setup.py#L41

That is over 2 years old, so I don't know why it would be trying to install such an old version of AutoGluon...

PGijsbers:
While not everything in this PR was in #564, it is my understanding that this PR is superseded and there will be other separate PRs for e.g., adding TS support for other frameworks. Closing this PR, just ping if that's not correct.

PGijsbers:
Thank you all for your efforts and patience! And I am happy we finally got around to merging things with #564 🎉

PR #511, opened 2023-02-20
Switch to using Python 3.8 instead of Python 3.7

Innixma:
Python 3.7 is no longer supported by Python, and some frameworks such as AutoGluon only support Python 3.8+ in their latest releases. Further, many packages such as scikit-learn, pandas, numpy, etc. no longer support Python 3.7.

I'd recommend AMLB use Python 3.8 or Python 3.9 as the new default. (I've tested setting Python 3.9, works out of the box with AutoGluon).

### DIFF
--- a/resources/config.yaml
+++ b/resources/config.yaml
@@ -88,7 +88,7 @@ openml:                # configuration namespace for openML.
 
 versions:              # configuration namespace for versions enforcement (libraries versions are usually enforced in requirements.txt for the app and for each framework).
   pip:
-  python: 3.7          # the Python minor version that will be used by the application in containers and cloud instances, , also used as a based version for virtual environments created for each framework.
+  python: 3.8          # the Python minor version that will be used by the application in containers and cloud instances, , also used as a based version for virtual environments created for each framework.
 
 container: &container          # parent configuration namespace for container modes.
   force_branch: true           # set to true if image can only be built from a clean branch, with same tag as defined in `project_repository`.


PGijsbers:
I think 3.9 would be a better choice as the scientific python community is about to drop releases for 3.8 in a month as per [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html). A secondary concern would be to provide support for installations of older frameworks, while the Python versions can be independent in principle. I think it's now always fixed to be `versions.python` and there isn't a good mechanism for pinning Python versions based on framework versions.  Suggestions welcome.

Innixma:
> I think 3.9 would be a better choice as the scientific python community is about to drop releases for 3.8 in a month as per [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html). A secondary concern would be to provide support for installations of older frameworks, while the Python versions can be independent in principle, they are not in practice. I think it's now always fixed to be `versions.python` and there isn't a good mechanism for pinning Python versions based on framework versions. Suggestions welcome.

3.9 should be fine, I actually switched to using 3.9 and it appears to work for AutoGluon (AutoGluon supports 3.8-3.10 in latest v0.7 release). Let me know if you'd like me to edit the PR to be 3.9, or else I think you should be able to directly commit to this PR and update it on your end.

Innixma:
@PGijsbers Please feel free to close this PR in favor of #520

PR #522, opened 2023-04-15
Update mlnet

LittleLittleCloud:


LittleLittleCloud:
@PGijsbers Hi can you check why ci/cd build is cancelled all the time?

PGijsbers:
Not entirely sure. I just merged #520, maybe you can rebase your PR on top of the updated `master` and see if the error persists?

LittleLittleCloud:
@PGijsbers Is it possible to use a larger machine for building pipeline? The building pipeline fails at dresses-sales because of 143 while local run succeed all the time, which indicates that the training process is killed by system because of

LittleLittleCloud:
@PGijsbers I finally got some time to fix this PR. Could you review it.

frameworks/MLNet/setup.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 HERE=$(dirname "$0")
-MLNET='mlnet'
+MLNET_PACKAGE='mlnet-linux-x64'
```
PGijsbers, line 3: Would it be possible to adjust this so it also works on ARM (and preferably also macos)? For example by storing the result of:
```python
import sys
import platform
platform = {"linux": "linux", "darwin": "osx"}.get(sys.platform)
machine =  {"arm64": "arm64", "x86_64": "x64"}.get(platform.machine())
print(f"mlnet-{platform}-{machine}")
```

PGijsbers (changes requested):
Thanks! Looks good at first glance, but I have a minor suggestion to not restrict the installation unnecessarily. I'll try out your changes when I am back in office next week.

frameworks/MLNet/setup.sh
PGijsbers: That should be in the above script already :)

PR #452, opened 2022-02-01
Save results on each job completion

sebhrusen:
Results are currently saved when all jobs are completed, which is a bit annoying for so many reasons (what if process hangs or get killed before completion? can't easily access completed tasks even when running locally...)
This PR changes the default behaviour to now save the results incrementally after each job.

also addressing https://github.com/openml/automlbenchmark/issues/418

PR #526, opened 2023-05-30
[Open for review] Store results after each job completion

PGijsbers:
Update requirements files to account for #520.
Supersedes #452.

Has a few small patches for bugs, which probably should be looked into later (see comments).


amlb/utils/process.py
```diff
@@ -153,9 +172,20 @@ def read_pipe(pipe, timeout):
...
-                                                 process.stderr if process.stderr else 2], activity_timeout),
-                              ['', '']))
+    process_output = list(iter(
+        lambda: read_pipe(
+            [process.stdout if process.stdout else 1,
+             process.stderr if process.stderr else 2
+             ], activity_timeout
+    ), ['', '']))
+    if not process_output:
+        log.warning(
```
PGijsbers, line 182: When tracking process output, if no output is provided this can be an issue with logging. However, depending on the configuration and the integration script, it is also possible that no logging output is the expected scenario. Instead of always giving a warning like this patch does, we should try to inspect the logging configuration (and possibly even integration script) to check whether or not a warning should be give, and what the right warning is.

amlb/runners/aws.py
```diff
@@ -924,6 +924,11 @@ def download_file(obj, dest, dest_display_path=None):
         self._update_instance(instance_id, success=success)
         return success, error
 
+    def _results_summary(self, scoreboard=None):
+        log.info(
```
PGijsbers, line 928: It is technically possible to still provide a result summary, which would be nice. The old version also did not provide a summary either, though. So for now we just have this message (the old PR #452 would simply crash here, because it would format the results of an empty dataframe in the base class level).

PGijsbers:
This PR is now open for review. Anyone that wants to can provide a review, but please mostly focus on correctness. I "adopted" this PR from Seb, I am not exactly sure why certain decisions were made but I think merging will still be valuable.

PGijsbers:
closing/opening to trigger CI with the latest `master` changes

PR #532, opened 2023-06-12
Measure inference time

PGijsbers:
**help wanted:** I am posting this now since it would be nice to have some feedback. I need merge this very soon in order to run experiments on time.

---

This PR introduces improved measurements for inference time, there are two major changes:
 - inference time is now to be measured the same way across frameworks: everything from reading from disk to producing predictions
 - inference time is measured for different dataset sizes, created by sampling with replacement from the test set

A few notes:
 - [x] I think we should allow also a second measurement which measures the inference time based on a Python object input (e.g., dataframe), for frameworks that support that. It would be nice to also know what the inference time is without the overhead of loading the file (which for small batches probably dominates). (edtit: example for autogluon present)
 - [ ] The implementation is probably not ideal. To be quite honest, this touches parts of the benchmark I hadn’t worked with before so I had to adjust as I go. If this is acceptable, I would rather prefer to refactor it later instead.
 - [x] Currently we only use one subsample per inference batch size, we should generate multiple. We should also make the amount of measurements configurable.
 - [x] I would like to also store each individual measurement, not just the aggregate. 
   - [x] This way we can do things like calculate different aggregates (median/mean) and/or exclude some measurements (some models may have a cold-start, which should be excluded from the aggregate).
 - [x] ~I need to add a way to clean up the generated files.~ They are currently cleaned up by virtue of them being stored in the temporary directory (if the framework uses `run_in_venv`). It might be nicer to generate and delete them right when doing the inference measurements in the subprocess, but in that case we need to be able to generate the splits without the `OpenMLDataset` object. Future work.
 
Current implementation’s result file after running `python runbenchmark.py FRAMEWORK -t iris -f 0` for `constantpredictor`, `tpot`,  and `autogluon`:
```
id,task,framework,constraint,fold,type,result,metric,mode,version,params,app_version,utc,duration,training_duration,predict_duration,models_count,seed,info,acc,balacc,logloss,inference_10000_rows,inference_1000_rows,inference_1_rows,models_ensemble_count
openml.org/t/59,iris,constantpredictor,test,0,multiclass,-1.09861,neg_logloss,local,1.2.2,,"dev [measure_inference_time, 71e7a8b]",2023-06-12T09:58:32,0.2,0.0002,4e-05,1,1757794907,,0.333333,0.333333,1.09861,0.00123646,0.00113094,0.00232852,
openml.org/t/59,iris,TPOT,test,0,multiclass,-0.0131854,neg_logloss,local,0.12.0,,"dev [measure_inference_time, 71e7a8b]",2023-06-12T09:59:55,76.4,65.4,0.0002,368,1156006760,,1.0,1.0,0.0131854,0.0196556,0.00544026,0.0176911,
openml.org/t/59,iris,AutoGluon,test,0,multiclass,-0.746956,neg_logloss,local,0.7.0,,"dev [measure_inference_time, 71e7a8b]",2023-06-12T10:00:30,28.4,18.7,0.002,14,2011490481,,0.933333,0.933333,0.746956,0.0672353,0.0116372,0.00261562,2.0
```

Notes:
 - [Measurements for TPOT are inaccurate](https://github.com/openml/automlbenchmark/pull/532#discussion_r1227194852)
 - [We might run into issues for large datasets (many features)](https://github.com/openml/automlbenchmark/pull/532#discussion_r1227203246), for the experiments we will simply (re)run those datasets a more lenient job timeout (or none at all).

### DIFF
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -96,7 +96,8 @@ class OpenmlDataset(Dataset):
     def inference_subsample_files(self, fmt: str) -> list[Tuple[int, str]]:
         return [
             (n, str(self._inference_subsample(fmt=fmt, n=n)))
-            for n in [1, 1000, 10_000]
+            for n in rconfig().inference_time_measurements.batch_sizes
+            for _ in range(rconfig().inference_time_measurements.repeats)
         ]
 
     @profile(logger=log)
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -94,6 +94,14 @@ class OpenmlDataset(Dataset):
         return self._test
 
     def inference_subsample_files(self, fmt: str) -> list[Tuple[int, str]]:
+        """Generates n subsamples of size k from the test dataset in `fmt` data format.
+
+        We measure the inference time of the models for various batch sizes
+        (number of rows). We generate config.inference_time_measurements.repeats
+        subsamples for each of the config.inference_time_measurements.batch_sizes.
+        These subsamples are stored to file in the `fmt` format (parquet, arff, or csv).
+        The function returns a list of tuples of (batch size, file path).
+        """
         seed = rget().seed(self.fold)
         return [
             (n, str(self._inference_subsample(fmt=fmt, n=n, seed=seed + i)))
@@ -106,9 +114,6 @@ class OpenmlDataset(Dataset):
         """ Write subset of `n` samples from the test split to disk in `fmt` format """
         # Just a hack for now, the splitters all work specifically with openml tasks.
         # The important thing is that we split to disk and can load it later.
-        if fmt not in ["csv", "arff", "parquet"]:
-            msg = f"{fmt=}, but must be one of 'csv', 'arff', or 'parquet'."
-            raise ValueError(msg)
 
         # We should consider taking a stratified sample if n is large enough,
         # inference time might differ based on class
@@ -131,6 +136,9 @@ class OpenmlDataset(Dataset):
             )
         elif fmt == "parquet":
             subsample.to_parquet(subsample_path)
+        else:
+            msg = f"{fmt=}, but must be one of 'csv', 'arff', or 'parquet'."
+            raise ValueError(msg)
 
         return subsample_path
 
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -93,7 +93,7 @@ class OpenmlDataset(Dataset):
         self._ensure_split_created()
         return self._test
 
-    def inference_subsample_files(self, fmt: str) -> list[Tuple[int, str]]:
+    def inference_subsample_files(self, fmt: str, with_labels: bool = False) -> list[Tuple[int, str]]:
         """Generates n subsamples of size k from the test dataset in `fmt` data format.
 
         We measure the inference time of the models for various batch sizes
@@ -104,20 +104,21 @@ class OpenmlDataset(Dataset):
         """
         seed = rget().seed(self.fold)
         return [
-            (n, str(self._inference_subsample(fmt=fmt, n=n, seed=seed + i)))
+            (n, str(self._inference_subsample(fmt=fmt, n=n, seed=seed + i, with_labels=with_labels)))
             for n in rconfig().inference_time_measurements.batch_sizes
             for i, _ in enumerate(range(rconfig().inference_time_measurements.repeats))
         ]
 
     @profile(logger=log)
-    def _inference_subsample(self, fmt: str, n: int, seed: int = 0) -> pathlib.Path:
+    def _inference_subsample(self, fmt: str, n: int, seed: int = 0, with_labels: bool = False) -> pathlib.Path:
         """ Write subset of `n` samples from the test split to disk in `fmt` format """
         # Just a hack for now, the splitters all work specifically with openml tasks.
         # The important thing is that we split to disk and can load it later.
 
         # We should consider taking a stratified sample if n is large enough,
         # inference time might differ based on class
-        subsample = self._test.X.sample(
+        test = self._test.data if with_labels else self._test.X
+        subsample = test.sample(
             n=n,
             replace=True,
             random_state=seed,
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -94,14 +94,15 @@ class OpenmlDataset(Dataset):
         return self._test
 
     def inference_subsample_files(self, fmt: str) -> list[Tuple[int, str]]:
+        seed = rget().seed(self.fold)
         return [
-            (n, str(self._inference_subsample(fmt=fmt, n=n)))
+            (n, str(self._inference_subsample(fmt=fmt, n=n, seed=seed + i)))
             for n in rconfig().inference_time_measurements.batch_sizes
-            for _ in range(rconfig().inference_time_measurements.repeats)
+            for i, _ in enumerate(range(rconfig().inference_time_measurements.repeats))
         ]
 
     @profile(logger=log)
-    def _inference_subsample(self, fmt: str, n: int) -> pathlib.Path:
+    def _inference_subsample(self, fmt: str, n: int, seed: int = 0) -> pathlib.Path:
         """ Write subset of `n` samples from the test split to disk in `fmt` format """
         # Just a hack for now, the splitters all work specifically with openml tasks.
         # The important thing is that we split to disk and can load it later.
@@ -114,7 +115,7 @@ class OpenmlDataset(Dataset):
         subsample = self._test.X.sample(
             n=n,
             replace=True,
-            random_state=rget().seed(self.fold)
+            random_state=seed,
         )
 
         _, test_path = self._get_split_paths()
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -125,7 +125,7 @@ class OpenmlDataset(Dataset):
 
         _, test_path = self._get_split_paths()
         test_path = pathlib.Path(test_path)
-        subsample_path = test_path.parent / f"{test_path.stem}_{n}.{fmt}"
+        subsample_path = test_path.parent / f"{test_path.stem}_{n}_{seed}.{fmt}"
         if fmt == "csv":
             subsample.to_csv(subsample_path, header=True, index=False)
         elif fmt == "arff":
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -2,13 +2,14 @@
 **openml** module implements the abstractions defined in **data** module
 to expose `OpenML<https://www.openml.org>`_ datasets.
 """
+import pathlib
 from abc import abstractmethod
 import copy
 import functools
 import logging
 import os
 import re
-from typing import Generic, Tuple, TypeVar, Union
+from typing import Generic, Tuple, TypeVar
 
 import arff
 import pandas.api.types as pat
@@ -16,7 +17,7 @@ import openml as oml
 import xmltodict
 
 from ..data import AM, DF, Dataset, DatasetType, Datasplit, Feature
-from ..resources import config as rconfig
+from ..resources import config as rconfig, get as rget
 from ..utils import as_list, lazy_property, path_from_split, profile, split_path, unsparsify
 
 
@@ -91,6 +92,39 @@ class OpenmlDataset(Dataset):
     def test(self):
         self._ensure_split_created()
         return self._test
+
+    @profile(logger=log)
+    def _inference_subsample(self, fmt: str, n: int) -> pathlib.Path:
+        """ Write subset of `n` samples from the test split to disk in `fmt` format """
+        # Just a hack for now, the splitters all work specifically with openml tasks.
+        # The important thing is that we split to disk and can load it later.
+        if fmt not in ["csv", "arff", "parquet"]:
+            msg = f"{fmt=}, but must be one of 'csv', 'arff', or 'parquet'."
+            raise ValueError(msg)
+
+        # We should consider taking a stratified sample if n is large enough,
+        # inference time might differ based on class
+        subsample = self._test.X.sample(
+            n=n,
+            replace=True,
+            random_state=rget().seed(self.fold)
+        )
+
+        _, test_path = self._get_split_paths()
+        test_path = pathlib.Path(test_path)
+        subsample_path = test_path.parent / f"{test_path.stem}_{n}.{fmt}"
+        if fmt == "csv":
+            subsample.to_csv(subsample_path, header=True, index=False)
+        elif fmt == "arff":
+            ArffSplitter(self)._save_split(
+                subsample,
+                subsample_path,
+                name=f"{self._oml_dataset.name}_inference_{self.fold}_{n}"
+            )
+        elif fmt == "parquet":
+            subsample.to_parquet(subsample_path)
+
+        return subsample_path
 
     @lazy_property
     @profile(logger=log)
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -93,6 +93,12 @@ class OpenmlDataset(Dataset):
         self._ensure_split_created()
         return self._test
 
+    def inference_subsample_files(self, fmt: str) -> list[Tuple[int, str]]:
+        return [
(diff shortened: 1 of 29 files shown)


frameworks/AutoGluon/exec.py
```diff
@@ -68,14 +69,21 @@ def run(dataset, config):
     # Persist model in memory that is going to be predicting to get correct inference latency
     predictor.persist_models('best')
 
+    def inference_time_classification(path: str):
+        data = TabularDataset(path)
```
Innixma, line 73: `data = TabularDataset(path)` is not required, predict and predict_proba internally load as TabularDataset if you pass the string path as input.

This also allows to pass either the DataFrame or the path as input to the function, depending on what you want to measure.

frameworks/AutoGluon/exec.py
```diff
@@ -68,14 +69,21 @@ def run(dataset, config):
...
+        return None, predictor.predict_proba(data, as_multiclass=True)
+
+    def inference_time_regression(path: str):
+        data = TabularDataset(path)
+        return predictor.predict(data, as_pandas=False), None
+
+    infer = inference_time_classification if is_classification else inference_time_regression
```
Innixma, line 80: If we are loading the test data inside the function, then we can save memory by not loading the test data in line 68 (otherwise it is existing in memory twice, which isn't optimal). Instead do `test_data = TabularDataset(test_path)` after the inference times are calculated to minimize memory usage. (after line 86)

frameworks/AutoGluon/exec.py
```diff
@@ -68,14 +69,21 @@ def run(dataset, config):
...
     if is_classification:
-        with Timer() as predict:
-            probabilities = predictor.predict_proba(test_data, as_multiclass=True)
         predictions = probabilities.idxmax(axis=1).to_numpy()
-    else:
-        with Timer() as predict:
-            predictions = predictor.predict(test_data, as_pandas=False)
-        probabilities = None
+
+    inference_times = measure_inference_times(infer, dataset.inference_subsample_files)
```
Innixma, line 86: Seems a bit forceful to always call this function. Maybe better to do something akin to `if enable_inference_subsampling`?

frameworks/AutoGluon/exec.py
```diff
@@ -68,14 +69,21 @@ def run(dataset, config):
...
+        data = TabularDataset(path)
+        return None, predictor.predict_proba(data, as_multiclass=True)
+
+    def inference_time_regression(path: str):
+        data = TabularDataset(path)
+        return predictor.predict(data, as_pandas=False), None
+
+    infer = inference_time_classification if is_classification else inference_time_regression
+    with Timer() as predict:
+        predictions, probabilities = infer(test_data)
```
Innixma, line 82: Unrelated to the PR, but it might be good to consider having an alternative test data file with the label column removed, that is explicitly loaded for these predict calls across all frameworks. This would further sanity check that no system is using the labels to influence the predictions they make.

I think it isn't really a problem at present since the systems are pretty trustworthy, but for new systems it would be helpful to have that guardrail.

frameworks/AutoGluon/exec.py
Innixma, line 86: Unsure how realistic loading 1 row from a parquet file is for the real-time inference batch_size=1 scenario. I'd assume in most cases the data is part of a pipeline and thus is already expected to be in memory.

This is a minor point and I don't think it matters too deeply for the purposes of a paper.

frameworks/TPOT/exec.py
```diff
@@ -67,6 +70,18 @@ def run(dataset, config):
...
+    def infer(path):
+        data = pd.read_parquet(path)
+        if is_classification:
+            try:
+                return tpot.predict_proba(data)
+            except RuntimeError:
+                return tpot.predict(data)
+        return tpot.predict(data)
+
+    inference_times = measure_inference_times(infer, dataset.inference_subsample_files)
```
Innixma, line 83: Any consideration of loading the raw file and doing the preprocessing done by AMLB to make it compatible with frameworks like TPOT live? This would properly compute the true inference time, as currently TPOT skips the preprocessing overhead since it is given pre-processed data, unlike frameworks such as AutoGluon and H2O that take in the raw data and have to pre-process it as part of the inference call.

Specifically TPOT is given the encoded version of the data, while ignoring the time it takes to encode it:

```
    X_train, X_test = dataset.train.X_enc, dataset.test.X_enc
    y_train, y_test = unsparsify(dataset.train.y_enc, dataset.test.y_enc)
```

Beyond ignoring the time it take to encode it, it further benefits from the (in many cases) smaller data file artifact that would be able to load from disk faster than the raw file which is larger.

frameworks/shared/callee.py
```diff
@@ -86,3 +89,12 @@ def load_data(name, path, **_):
         kill_proc_tree(include_parent=False, timeout=5)
 
     json_dump(res, config.result_file, style='compact')
+
+def measure_inference_times(predict_fn: Callable[[str], Any], files: list[Tuple[int, str]]) -> dict[int, list[float]]:
+    inference_times = defaultdict(list)
+    for subsample_size, subsample_path in files:
+        for _ in range(10):
```
Innixma, line 96: Just a heads up: for `arcene` and `Mercedes_Benz_Greener_Manufacturing` you might run into very long run-times for this call when predicting 10,000 rows with 10 repeats depending on the framework.

The current logic of AMLB is to terminate if the process goes on for too long, so you might get some failures due to this on otherwise successful frameworks. This can be solved by disabling the termination logic or reducing the number of repeats.

amlb/results.py
```diff
@@ -444,6 +444,10 @@ def compute_score(self, result=None, meta_result=None):
         required_meta_res = ['training_duration', 'predict_duration', 'models_count']
         for m in required_meta_res:
             entry[m] = meta_result[m] if m in meta_result else nan
+
+        if inference_times := Namespace.get(meta_result, "inference_times"):
+            for n_samples, measured_times in Namespace.dict(inference_times).items():
+                entry[f"inference_{n_samples}_rows"] = np.mean(measured_times)
```
Innixma, line 450: small nit: `f"inference_rows_{n_samples}"` will be easier to parse downstream via the shared prefix `inference_rows_`.

An alternative name: `infer_batch_size_{n_samples}`

Innixma (commented):
Looks good! I added some minor comments and suggestions

mfeurer (commented):
Some comments on the general notes:
> I think we should allow also a second measurement which measures the inference time based on a Python object input (e.g., dataframe), for frameworks that support that. It would be nice to also know what the inference time is without the overhead of loading the file (which for small batches probably dominates).

Yes, this would be great.

> The implementation is probably not ideal. To be quite honest, this touches parts of the benchmark I hadn’t worked with before so I had to adjust as I go. If this is acceptable, I would rather prefer to refactor it later instead.

I cannot comment on this as I do not know the internals either.

> Currently we only use one subsample per inference batch size, we should generate multiple. We should also make the amount of measurements configurable.

Yes, this would be great. Especially for small batch sizes like 1 there might be a high variance.

> I would like to also store each individual measurement, not just the aggregate. This way we can do things like calculate different aggregates (median/mean) and/or exclude some measurements (some models may have a cold-start, which should be excluded from the aggregate).

That sounds useful.

> I need to add a way to clean up the generated files.

That as well.

Moreover, I was wondering whether you could also retrieve the times for 10 and 100 samples? Or is this something that doesn't appear in real-life (maybe @Innixma knows?)?

frameworks/AutoGluon/exec.py
PGijsbers: Opened an issue https://github.com/openml/automlbenchmark/issues/533

frameworks/AutoGluon/exec.py
PGijsbers: I don't think it's very realistic, likely the overhead of accessing the file dominates inference time for all but the most unwieldy models in this setting.  But for some frameworks it is the only way we can measure them (e.g., frameworks in other languages) unless we rely on self-reported numbers (if they provide any).

amlb/datasets/openml.py
```diff
@@ -92,6 +93,45 @@ def test(self):
...
+        if fmt == "csv":
+            subsample.to_csv(subsample_path, header=True, index=False)
+        elif fmt == "arff":
+            ArffSplitter(self)._save_split(
+                subsample,
+                subsample_path,
+                name=f"{self._oml_dataset.name}_inference_{self.fold}_{n}"
+            )
+        elif fmt == "parquet":
+            subsample.to_parquet(subsample_path)
```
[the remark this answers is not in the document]
PGijsbers: It was already guarded at the top of the function, but I moved it to `else` for clarity (and it's less error prone).

frameworks/TPOT/exec.py
PGijsbers: I had a look at this, but it seems quite convoluted to achieve this (as files are currently generated outside of the subprocess). Since TPOT is the only automl framework affected by this, and the remainder can work on raw parquet/arff/csv files, I think I will leave this for now and exclude `TPOT` from the comparison (an unfortunate choice to add it first to this PR).

frameworks/shared/callee.py
PGijsbers: Thanks for the heads up (again). I am not sure of a nice way to make this work automatically. On some level, it's still useful to have automatic termination (some frameworks won't stop otherwise), so I imagine we best increase it across the board. Perhaps I will rerun only those specific datasets with even more lenient (or absent) termination times.

frameworks/AutoGluon/exec.py
Innixma: Yeah, if you can add that for the Python frameworks, at least for batch size 1, that would be good. What we are really after is end-to-end latency for online inference scenarios where we can assume the data is already in memory (for example, being passed to a server endpoint with a warm-started model)

Innixma:
> Moreover, I was wondering whether you could also retrieve the times for 10 and 100 samples? Or is this something that doesn't appear in real-life (maybe @Innixma knows?)?

@mfeurer It does occur, but it is generally an easier scenario than batch 1 and large-batch.

10 and 100 samples will often have near identical total latency as 1 sample. It is only once you go beyond 100 (and often 1000), that you start to avoid fixed-cost overheads dominating the runtime.

I'd be ok both with and without those 10 & 100 measurements. I think 1 and 10000+ are the most important as they are the most challenging.

frameworks/AutoGluon/exec.py
PGijsbers: I updated the PR with an example of how we could do this for AutoGluon. It's getting a bit unwieldy, but I think it works (read: will prioritise that the merge, refactor later).

PGijsbers:
status update (and note to self):
 - processed most feedback
 - added an example for inference using a dataframe (see autogluon), plan to expand to python frameworks for single row inference speed
 - having issues getting the arff split files to work with the H2O integration. It gives an error that "no columns in common" even though manual checking seems to suggest that the files are fine.
 
 Once the last two points are resolved, I plan to merge and start running validation tests.

frameworks/AutoGluon/exec.py
```diff
@@ -64,18 +67,30 @@ def run(dataset, config):
...
+
+    def inference_time_regression(data: Union[str, pd.DataFrame]):
+        return predictor.predict(data, as_pandas=False), None
+
+    infer = inference_time_classification if is_classification else inference_time_regression
+    inference_times = {}
+    if config.measure_inference_time:
+        inference_times["file"] = measure_inference_times(infer, dataset.inference_subsample_files)
+        test_data = pd.read_parquet(dataset.test.path)
+        inference_times["df"] = measure_inference_times(
```
Innixma, line 84: (Suggestion, feel free to create a separate issue as this is mostly a QoL/robustness comment)

We should ensure `inference_times` is the same type structure for all frameworks. Currently AutoGluon has an additional layer of dictionary compared to `GAMA`.

I think `GAMA` should be updated to be `inference_times['file'] = ...`

We may want to have type checking asserts as part of the `return result(..., inference_times=inference_times)` logic that is common to all frameworks.

Something like:

```
if inference_times is not None:
    assert isinstance(inference_times, dict)
    if len(inference_times) > 0:
        for k, v in inference_times:
            assert isinstance(k, str)
            assert isinstance(v, dict)
            for batch_size, infer_time_list in v.items():
                assert isinstance(batch_size, int)
                assert isinstance(infer_time_list, list)
                for infer_time in infer_time_list:
                     assert isinstance(infer_time, float)
                     assert infer_time >= 0
```

frameworks/AutoGluon/exec.py
```diff
@@ -64,18 +67,30 @@ def run(dataset, config):
...
+            [(1, test_data.sample(1, random_state=i)) for i in range(100)],
+        )
+
+    test_data = TabularDataset(test_path)
+    with Timer() as predict:
+        predictions, probabilities = infer(test_data)
     if is_classification:
-        with Timer() as predict:
-            probabilities = predictor.predict_proba(test_data, as_multiclass=True)
         predictions = probabilities.idxmax(axis=1).to_numpy()
```
Innixma, line 93: Note: in v0.8, we now have a cleaner way to do this: https://auto.gluon.ai/dev/api/autogluon.tabular.TabularPredictor.get_pred_from_proba.html

We can probably keep as is for now. Switching will only matter if we end up running on metrics where decision threshold calibration matters (balanced accuracy, f1, etc.), as the current code in AMLB is forcibly using decision_threshold=0.5.

frameworks/AutoGluon/exec.py
```diff
@@ -64,18 +67,30 @@ def run(dataset, config):
...
+        inference_times["file"] = measure_inference_times(infer, dataset.inference_subsample_files)
+        test_data = pd.read_parquet(dataset.test.path)
+        inference_times["df"] = measure_inference_times(
+            infer,
+            [(1, test_data.sample(1, random_state=i)) for i in range(100)],
+        )
+
+    test_data = TabularDataset(test_path)
+    with Timer() as predict:
+        predictions, probabilities = infer(test_data)
```
Innixma, line 91: I thought we were going to include the file loading as part of the predict time? Or is this not the case?

Current code would not include loading time as part of the time.

frameworks/AutoGluon/exec.py
```diff
@@ -64,18 +67,30 @@ def run(dataset, config):
...
+    def inference_time_classification(data: Union[str, pd.DataFrame]):
+        return None, predictor.predict_proba(data, as_multiclass=True)
+
+    def inference_time_regression(data: Union[str, pd.DataFrame]):
+        return predictor.predict(data, as_pandas=False), None
+
+    infer = inference_time_classification if is_classification else inference_time_regression
+    inference_times = {}
+    if config.measure_inference_time:
+        inference_times["file"] = measure_inference_times(infer, dataset.inference_subsample_files)
```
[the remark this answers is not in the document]
PGijsbers: Generating them are part of `measure_inference_times` is what I wanted to do initially, but I am currently relying on the automated data loading that `amlb/datasets/openml.py::OpenMLDataset` provides (as well as the arff splitter), which is not available from within the `run_in_venv` call (hence doing that in the `__init__.py`). 
I could use generated file names, but just passing the file names doesn't seem too much of a problem to me.

frameworks/AutoGluon/exec.py
PGijsbers: Yes, I deliberately only updated `AutoGluon` because I felt updating one framework was enough to get an idea for feedback. If all seems well (enough), I will update the other frameworks.

I am not sure about the `assert` type checking. I hope to refactor amlb in the future to contain more type annotations, use mypy, etc. which should catch most of it.

frameworks/AutoGluon/exec.py
PGijsbers: We won't use those metrics for our next set of experiments, but PR to make things _also_ work for `0.8` is welcome.

frameworks/AutoGluon/exec.py
PGijsbers: It does for `measure_inference_times`, I'll make sure it does for `predict` too, where possible. thanks!

frameworks/AutoGluon/exec.py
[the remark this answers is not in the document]
Innixma: I see, that sounds reasonable for now

PR #534, opened 2023-06-14
Allow newer autosklearn versions to use the pandas data instead

PGijsbers:
Use pandas dataframes with `autosklearn>=0.15`, instead of the encoded numpy arrays.
This is preferred because it is more "auto" and because it makes inference time measurements more comparable across frameworks.

note: this is on top of the `measure_inference_time` PR #532.

@eddiebergman @mfeurer are there any other changes that should be made to account for the switch?


eddiebergman (approved):
Looks good to me, however probably better to have approval by @mfeurer first.

mfeurer (approved):

PR #543, opened 2023-06-20
Update AutoGluon `max_memory` from 0.1 to 0.4 in persist_models

Innixma:
Updated max_memory to 0.4 as when tested it resulted in no crashes. The 0.1 default is overly conservative. For batch_size=1 in inference, this gives ~2x faster inference speed on average as more frequently models are able to fit into memory using this raised limit.

### DIFF
--- a/frameworks/AutoGluon/exec.py
+++ b/frameworks/AutoGluon/exec.py
@@ -68,7 +68,7 @@ def run(dataset, config):
         )
 
     # Persist model in memory that is going to be predicting to get correct inference latency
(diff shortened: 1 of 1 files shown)


Innixma:
I've also opened a PR for AutoGluon to officially use 0.4 going forward as the default (although 0.8 still uses 0.1 as default)

https://github.com/autogluon/autogluon/pull/3338

PGijsbers:
Merging this now, but please revert the change (or make it conditional), when a new release with the changed default is available. Thanks!

PR #549, opened 2023-06-21
Add Naive AutoML integration

PGijsbers:
Adding NaiveAutoML 🎉

@fmohr @mwever  could you have a look at my usage of NaiveAutoML? I did the integration purely based of your docs and source code, so it is possible I misunderstood some of the usage.

Some questions:
 - I noticed NAML tracks memory usage, but is there a way to provide it with a memory constraint?
 - I wasn't able to get reproducible behavior for providing a target (`y`) the was a pandas series of str/categorical. One time it froze, another it would spew errors, so I couldn't open an issue. So for now within AMLB I decided to use the option for using an encoded target variable.
 - Should we set `execution_time`? A fixed timeout regardless of `timeout` or `dataset` size seems prone to failure. Or does it adapt if left at default? If it's currently fixed to 10 seconds, I propose to instead set it to 5~10% of the total time.
 
TODO:
 - run full-length `validation`:
   - [x] locally
   - [x] docker
   - [ ] aws
 - [x] add to CI grid
 - [x] add to `2023Q2` benchmark definitions

On merge:
 - [ ] open issue to add to website docs (opening an issue instead since I will probably rewrite the website somewhat first)

### DIFF
--- a/.github/workflows/run_all_frameworks.yml
+++ b/.github/workflows/run_all_frameworks.yml
@@ -52,7 +52,7 @@ jobs:
           skip_evaluation=0
           if [[ $is_common -eq 0 ]];
           then
-            FRAMEWORKS='["autogluon", "autosklearn", "gama", "h2oautoml", "mlplanweka", "mlr3automl", "tpot", "tunedrandomforest"]'
+            FRAMEWORKS='["autogluon", "autosklearn", "gama", "h2oautoml", "mlplanweka", "mlr3automl", "naiveautoml", "tpot", "tunedrandomforest"]'
             TASKS='["iris", "kc2", "cholesterol"]'
             BENCHMARK='["test"]'
           else
--- a/frameworks/NaiveAutoML/__init__.py
+++ b/frameworks/NaiveAutoML/__init__.py
@@ -1,0 +1,32 @@
+from amlb.benchmark import TaskConfig
+from amlb.data import Dataset
+from amlb.utils import call_script_in_same_dir
+
+
+def setup(*args, **kwargs):
+    call_script_in_same_dir(__file__, "setup.sh", *args, **kwargs)
+
+
+def run(dataset: Dataset, config: TaskConfig):
+    from frameworks.shared.caller import run_in_venv
+
+    data = dict(
+        target=dataset.target.name,
+        train=dict(
+            X=dataset.train.X,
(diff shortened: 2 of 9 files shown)


fmohr:
Hi @PGijsbers ,

wow thanks for working on this integration.

> I noticed NAML tracks memory usage, but is there a way to provide it with a memory constraint?

There is unfortunately no way to constraint the memory in NAML.

> I wasn't able to get reproducible behavior for providing a target (y) the was a pandas series of str/categorical. One time it froze, another it would spew errors, so I couldn't open an issue. So for now within AMLB I decided to use the option for using an encoded target variable.

This is new to me. I have used categorical target variables a lot (but maybe never as pandas Series objects). I mean it would still be great if you could create an issue for that, so I can have a look at it. Freezing does not sound good anyway.

> Should we set execution_time? A fixed timeout regardless of timeout or dataset size seems prone to failure. Or does it adapt if left at default? If it's currently fixed to 10 seconds, I propose to instead set it to 5~10% of the total time.

Currently its just a constant. So if we want something relative to total time, this needs to be computed previously when instantiating the NAML object. Your suggestion sounds reasonable to me, but this needs to be computed manually (currently). For upcoming releases, we plan to use [LCCV](https://github.com/fmohr/lccv) anyway  for candidate evaluation.

PGijsbers:
I am currently experiencing some issues, at some point in the amlb wrap-up process a `segfault` is thrown for some datasets. I can't reproduce it fully, and I haven't been able to set up any MWE outside of AMLB and those within AMLB take at least 20 minutes - extremely hard to debug. Despite it happening only after control is returned to amlb  (as far as I can tell), it seems unlikely that it's an amlb bug as none of the other frameworks are affected. I suspect it's triggered by garbage collection or unclosed processes somehow. Will investigate further.

PGijsbers:
I have not been able to track down the source, but I did manage to reproduce it outside of AMLB: https://github.com/fmohr/naiveautoml/issues/25

I tried to find the source within NAML, but I don't really have an idea and the flaky reproducibility combined with the time it takes to trigger the error makes it really tough. Considering the experiment deadline, I am going to go ahead and merge this since I am convinced that it's not related to the integration script per se. This way we can at least do a best-effort benchmark. Hopefully we can resolve the issue in the future.

PR #564, opened 2023-07-04
Time series forecasting improvements: metrics, dataset preprocessing & AutoGluon wrapper

shchur:
This PR contains several improvements to time series forecasting functionality in AMLB and supersedes #507.
- Refactor the wrapper for AutoGluon-TimeSeries
  - The code now works with AutoGluon v0.8.* - checked local, Docker and AWS.
  - Both AutoGluon-Tabular and AutoGluon-TimeSeries now correspond to a single framework `autogluon` with the same installation procedure. The correct predictor object is used depending on the `DatasetType`.
- Task definition
  - Now `quantile_levels` can be defined as a part of task definition, instead of always setting this argument to `[0.1, 0.2, ..., 0.8, 0.9]`.
  - More checks added to make sure that time series datasets are constructed correctly.
- Refactor metric definitions to make them consistent with the textbook definitions.
  - Always use the `predictions` column when computing point forecast metrics
  - Add new metric: scaled pinball loss
- Store auxiliary information needed to compute metrics (in-sample seasonal error & item IDs) as dataset attributes instead of delegating their computation to individual frameworks (as was done before).
  - Some scaled metrics (e.g., [mean absolute scaled error](https://en.wikipedia.org/wiki/Mean_absolute_scaled_error), [scaled pinball loss](https://www.kaggle.com/code/andrewcameronsmith/introduction-to-scaled-pinball-loss)) require information based on the historic time series values (before the forecast horizon begins). The `Result` object in AMLB does not have access to historic information, so we store this extra information as `optional_columns` in the predictions dataframe.
- Explain how to define a time series forecasting task in `HOWTO.md`.
----
Work left for future PRs
- Add more AutoML frameworks for forecasting
- Add more time series datasets


PGijsbers:
Thanks for updating the PR, sorry for our unresponsiveness on the old one. We currently have an open [fedot integration PR](https://github.com/openml/automlbenchmark/pull/563) and according to their [readme](https://github.com/aimclub/FEDOT) they also do "time series prediction problems". Would you be willing to talk with @nicl-nno and see if the two of you can add time series support for the FEDOT integration as well? That would be a good test as to whether the abstraction levels/supported features are sufficient to generalize this beyond just AutoGluonTS.

I am planning to have a closer look at the FEDOT PR tomorrow, but based on a first impression I expect to be able to merge it this week.

shchur:
Thanks! Sure, I will be happy to help with integrating the time series support for FEDOT.

The current PR does not really introduce new functionality and mostly refactors / fixes the existing code (to make it runnable), so I think it would still be useful, unless we plan to completely refactor all time series code in AMLB. 

In my opinion, it would make sense to work on the FEDOT time series PR after #563 and this PR are merged. Current abstractions are readily compatible with other libraries like StatsForecast, GluonTS, AutoPyTorch, so I don't expect there to be a lot of work needed for FEDOT. But if any changes to the core TS logic will be necessary for FEDOT, I'm happy to take care of them. Does that sound reasonable to you?

nicl-nno:
I agree with @shchur that is is better to separate the TS support to the new PR.

In case it's helpful: we have small self-developed benchmark for time series forecasting named [pytsbe](https://github.com/ITMO-NSS-team/pytsbe). 
It already contains [integration](https://github.com/ITMO-NSS-team/pytsbe/tree/main/pytsbe/models) with several AutoML solutions (FEDOT, LAMA, AutoTS, AutoGluon) and baselines. Maybe some code snippets will be useful for AMLB development.

shchur:
@PGijsbers I have just double-checked, the code works locally, on Docker, and on AWS, so the PR is ready for review.

PGijsbers:
> I agree with @shchur that is is better to separate the TS support for FEDOT to the new PR.

Sure, that sounds fine. I'll have a look at both this week (hopefully starting tomorrow). Thanks!

amlb/benchmark.py
```diff
@@ -404,6 +404,9 @@ def __init__(self, name, openml_task_id, test_server, fold, metrics, seed,
         self.git_info = git_info
         self.measure_inference_time = measure_inference_time
         self.ext = ns()  # used if frameworks require extra config points
+        if quantile_levels is None:
+            quantile_levels = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]
```
shchur, line 408: Is it possible to store the default `quantile_levels` in `config.yaml` instead of here?

PGijsbers: From my understanding it seems it would be fine under `benchmarks.defaults`, or do you think a different spot is more appropriate?

PGijsbers (changes requested):
I said it [last time](https://github.com/openml/automlbenchmark/pull/507#issuecomment-1362954560) as well, but we need good documentation on how to construct a time series task. In particular what the different properties mean (e.g., `forecast_horizon_in_steps`, `seasonality`, etc.) and what the data should look like (test set contains the training set).

shchur:
Thank you for the review and sorry for missing the relevant discussion on the previous PR!

This PR only contains minimal changes necessary to make the code runnable & correct the metric definitions. I have not made any changes to the dataset schema - but I think that the current schema is not ideal and could be substantially improved. I would like to incorporate these changes before updating the documentation. Below is a short overview of the current time series dataset schema and a possible modification.

If these changes look good to you, I will incorporate them to the PR & document them in `HOWTO.md`.

----
### Current dataset schema (master branch / this PR)

At the bare minimum, each time series task must be defined by

- `forecast_horizon_in_steps` - a positive integer that specifies how many steps into the future need to be forecast for each time series (= number of future time series values that need to be predicted)
- two CSV files
  - `train.csv` with 3 columns (`item_id` - unique ID of each time series, `timestamp` of the observation, `target` time series value), with contents such as

    | item_id | timestamp | target |
    |--------|--------|--------|
    | A | 2020-01-01 | 2.0 | 
    | A | 2020-01-02 | 4.0 | 
    | B | 2022-05-01 | 1.0 | 
    | B | 2020-05-02 | 3.0 | 
    | B | 2020-05-03 | 1.0 | 

  - `test.csv` with contents such as

    | item_id | timestamp | target |
    |--------|--------|--------|
    | A | 2020-01-01 | 2.0 | 
    | A | 2020-01-02 | 4.0 | 
    | A | 2020-01-03 | 5.0 | 
    | B | 2022-05-01 | 1.0 | 
    | B | 2020-05-02 | 3.0 | 
    | B | 2020-05-03 | 1.0 | 
    | B | 2020-05-04 | 0.0 | 

(in this example the dataset contains 2 time series `['A', 'B']` and `forecast_horizon_in_steps = 1`)

The `train.csv` and `test.csv` files must satisfy the following conditions
- The `timestamp`s for each individual time series are regularly sampled (e.g., daily, monthly, 30 min, frequency).
- For each time series in `train.csv`, `test.csv` contains all past values + exactly `forecast_horizon_in_steps` future values corresponding to the forecast horizon.

This dataset schema has a major limitation: the `forecast_horizon_in_steps` is "baked in" into the `train.csv`/`test.csv` files. If we wanted to evaluate on the same dataset with a different value for `forecast_horizon_in_steps`, we would need to re-generate the `train.csv` file.

To address this limitation, I propose the following modification to the dataset schema.

----
### Proposed changes (to be implemented in this PR or another PR, if you approve the design)

We only store a single file `test.csv` that is identical to the test file in the current schema.

The train set (called `train.csv` before) is generated automatically by AMLB by removing the last `forecast_horizon_in_steps` values from each time series in the test set.

If the specified `forecast_horizon_in_steps` is >= the length of the shortest time series in `test.csv`, an exception is raised. 

This also makes it easy to evaluate over multiple folds for time series data: for each fold `i` in `[0, 1, 2, 3, ...]`, we define
- the test set by removing `i * forecast_horizon_in_steps` from each original time series in `test.csv`
- the train set by removing `(i+1) * forecast_horizon_in_steps` from each original time series in `test.csv`
----

Please let me know if this design looks reasonable to you.

amlb/benchmark.py
shchur: Moved the default values to `benchmark.defaults` section of `config.yaml`.

frameworks/AutoGluon/__init__.py
```diff
@@ -40,22 +40,21 @@ def run_autogluon_timeseries(dataset: Dataset, config: TaskConfig):
         dataset.timestamp_column = None
     if not hasattr(dataset, 'id_column'):
         dataset.id_column = None
-    if not hasattr(dataset, 'forecast_range_in_steps'):
-        raise AttributeError("Unspecified `forecast_range_in_steps`.")
+    if not hasattr(dataset, 'forecast_horizon_in_steps'):
```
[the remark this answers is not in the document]
shchur: Good point, removed the assertion

PGijsbers:
> This also makes it easy to evaluate over multiple folds for time series data: for each fold i in [0, 1, 2, 3, ...], we define
>  * the test set by removing i * forecast_horizon_in_steps from each original time series in test.csv
>  * the train set by removing (i+1) * forecast_horizon_in_steps from each original time series in test.csv

I think that all makes sense, but it seems the description is off to me. Given `forecast_horizon_in_steps=K` then you want to have:

| rows | fold 0 | fold 1 | fold 2 | ... | fold n|
| :-- | :-- | :-- | :-- | :-- | :-- |
| 1..K | train | train| train | ... | train |
| K..2K | test | train| train | ... | train |
| 2..3K |   | test| train | ... | train |
| 3..4K |   |  | test | ... | train |
| ... |   |  |   |  |   |
| (n-1)K...nK |   |  |   | ... | test |

Right? The way I read your description the `test` set actually grows by `forecast_horizon_in_steps` each fold. And I image it is useful to specify a minimum train set greater than `forecast_horizon_in_steps` (e.g., training data is one year, and the folds consist of subsequent days (e.g., a forecast horizon of 3 days).

amlb/datasets/file.py
```diff
@@ -138,37 +135,45 @@ def _extract_train_test_paths(self, dataset, fold=None, name=None):
...
-            msg = f"Error: Not all sequences of train and test set have same sequence length difference."
-            raise ValueError(msg)
-        if dataset.forecast_range_in_steps > forecast_range_in_steps_mean_diff_train_test:
-            msg = f"Error: Forecast range {dataset.forecast_range_in_steps} longer than difference between test and train sequence lengths {forecast_range_in_steps_mean_diff_train_test}."
-            raise ValueError(msg)
-        if dataset.forecast_range_in_steps > forecast_range_in_steps_max_min_train_test:
-            msg = f"Error: Forecast range {dataset.forecast_range_in_steps} longer than minimum sequence length + 1, {forecast_range_in_steps_max_min_train_test}."
-            raise ValueError(msg)
+            log.warning("Warning: For timeseries task, setting undefined `id_column` to `item_id`.")
+            dataset_config['id_column'] = 'item_id'
```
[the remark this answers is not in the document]
shchur: Every dataset should contain `timestamp_column` and `id_column` - these are necessary to understand which rows in the CSV file correspond to each individual time series & sorting the values chronologically. I would expect that for most datasets these columns have the same default names (`timestamp` and `item_id`, respectively), so we don't need to copy the same code across all task definitions.

How about we change this code as follows?
```python
if dataset_config['id_column'] is None:
    if 'item_id' not in dataset.train.data:
        raise AssertionError('Task definition for timeseries must include `id_column` since the default ID column "item_id" is missing from the data.')
    else:
        log.warning("Warning: For timeseries task, setting undefined `id_column` to `item_id`.")
        dataset_config['id_column'] = 'item_id'
```

Alternatively, we could provide the default names for these columns somewhere else in the code.

shchur:
It seems that we are describing the same thing, but I was numbering the folds in reverse order 😅 

rows | fold 0 | fold 1 | ... | fold (n-1) | fold n
-- | -- | -- | -- | -- | --
1..K | train | train | ... | train | train
K..2K | train | train | ... | train | test
2..3K | train  | test | ... | test | 
... |   |   |     |  
(n-2)K...(n-1)K | train  |  test   | |
(n-1)K...nK | test  |    | |

This means, fold 0 aways corresponds to evaluation on the data with the most recent `K` timestamps for each time series (which recovers the setting used in most forecasting competitions).

This way if the user decides to evaluate on folds `[0, 1, 2, ..., n-1]`, then the `n` folds with the largest amount of training data will be used.

amlb/datasets/file.py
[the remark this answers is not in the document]
PGijsbers: Sorry for being unclear - it seems reasonable enough to expect some kind of identifier column and some timestamp column, just maybe not that the `id_column` is called `item_id`. But as long as an exception is raised of `"item_id"` is not present, I don't see an objection to it being the default. In any case you can drop the inner 'else` and I would also add an outer else to check if the provided value (if there is one) is actually present in the data:

```python
if dataset_config['id_column'] is None:
    if 'item_id' not in dataset.train.data:
        raise ValueError('Task definition for timeseries must include `id_column` since the default ID column "item_id" is missing from the data.')
    log.warning("Warning: For timeseries task, setting undefined `id_column` to `item_id`.")
    dataset_config['id_column'] = 'item_id'
elif dataset_config['id_column'] not in dataset.train.data:
    raise ValueError(f"`id_column` set to '{dataset_config["id_column"]}' but is missing from the data.")
```

I considered
```python
dataset_config['id_column']  =  dataset_config['id_column']  or "item_id"
if dataset_config['id_column'] not in dataset.train.data:
    raise ValueError(f"`id_column` set to '{dataset_config["id_column"]}' but is missing from the data.")
```
but the extra warning message on setting a default and more precise error messages seem useful enough for the code to be more verbose.

shchur:
@PGijsbers I have moved all time series related logic to a new class `TimeSeriesDataset` that inherits from `FileDataset` and added instructions on how to add new time series tasks to `HOWTO.md`. Please let me know if this looks good you.

docs/HOWTO.md
```diff
@@ -277,6 +281,62 @@ Then the datasets can be declared in the benchmark definition file as follow:
...
+    id_column: IdColumn  # defaults to 'item_id'
+    timestamp_column: TimestampColumn  # defaults to 'timestamp'
+    target: TargetColumn
+  folds: 1
+```
+The configuration of a time series task **must** include the following parameters:
+- `path` - a local or remote (e.g., S3) path to the CSV file with time series data.
+- `freq` - a [pandas-compatible frequency string](https://pandas.pydata.org/docs/user_guide/timeseries.html#offset-aliases) that denotes the frequency of the time series. For example, `D` for daily, `H` for hourly, or `15min` for 15-minute frequency.
+- `forecast_horizon_in_steps` - a positive integer denoting how many future time series values need to be predicted.
+- `seasonality` - a positive integer denoting the seasonal period of the data that is used for computing metrics like [mean absolute scaled error](https://en.wikipedia.org/wiki/Mean_absolute_scaled_error#Seasonal_time_series).
```
PGijsbers, line 328: I assume the unit is number of steps?

docs/HOWTO.md
```diff
@@ -277,6 +281,62 @@ Then the datasets can be declared in the benchmark definition file as follow:
...
+
+| IdColumn |	TimestampColumn |	TargetColumn |
+|---------|-----------|--------|
+| A       |	2020-01-01|	2.0    |
+| A       |	2020-01-02|	1.0    |
+| A       |	2020-01-03|	5.0    |
+| B       |	2019-01-01|	8.0    |
+| B       |	2019-01-01|	2.0    |
+| B       |	2019-01-01|	1.0    |
+| B       |	2019-01-01|	9.0    |
```
PGijsbers, line 306:
```suggestion
| B       |	2019-01-02|	2.0    |
| B       |	2019-01-03|	1.0    |
| B       |	2019-01-04|	9.0    |
```
I assume repeated timestamps are not allowed and this was a copy-paste error.

docs/HOWTO.md
```diff
@@ -277,6 +281,62 @@ Then the datasets can be declared in the benchmark definition file as follow:
...
+
+**Note**:
+- AMLB automatically generates the train & test splits from the raw data depending on the chosen `forecast_horizon_in_steps` and `fold` parameters. For each time series `x` in the datasets, the folds are constructed as:
+  - In fold `0`, use `x[:-forecast_horizon_in_steps]` as the train set and `x[-forecast_horizon_in_steps:]` as the test set (forecast horizon).
+  - In fold `1`, use `x[:-2*forecast_horizon_in_steps]` as the train set and `x[-2*forecast_horizon_in_steps:-forecast_horizon_in_steps]` as the test set.
+  - In fold `n`, use `x[:-(n+1)*forecast_horizon_in_steps]` as the train set and `x[-(n+1)*forecast_horizon_in_steps:-n*forecast_horizon_in_steps]` as the test set.
+- Time series may have different lengths or have different starting timestamps, but need to have the same frequency
+- The shortest time series in the dataset must have length of at least `folds * forecast_horizon_in_step + 1`
+- All time series must have regular timestamp index
+
```
PGijsbers, line 339: Adding the Table from our discussion would reduce the strain on the reader and make it clear at a glance.

amlb/datasets/file.py
```diff
@@ -346,26 +306,102 @@ def release(self, properties=None):
...
 
+        # Store repeated item_id & in-sample seasonal error for each time step in the forecast horizon - needed later for metrics like MASE.
+        # We need to store this information here because Result object has no access to past time series values.
+        self.repeated_item_id = self.test.data[self.id_column].cat.codes.to_numpy()
+        self.repeated_abs_seasonal_error = self.compute_seasonal_error()
+
+    def save_train_and_test_splits(self, full_data, fold, save_dir):
+        full_data = full_data.sort_values(by=[self.id_column, self.timestamp_column])
+        shortest_ts_length = full_data.groupby(self.id_column).size().min()
+        min_expected_ts_length = (fold + 1) * self.forecast_horizon_in_steps
```
PGijsbers, line 365: This is different from the documentation:
> The shortest time series in the dataset must have length of at least `folds * forecast_horizon_in_step + 1`

amlb/results.py
```diff
@@ -684,71 +684,108 @@ def r2(self):
...
-    @metric(higher_is_better=False)
-    def mase(self):
-        """Mean Absolute Scaled Error"""
-        return float(np.nanmean(np.abs(self.truth/self.y_past_period_error - self.predictions/self.y_past_period_error)))
+        if (~np.isfinite(self.predictions)).any() or (~np.isfinite(self.quantile_predictions)).any():
+            raise ValueError('Predictions contain NaN or Inf values')
+
+        _, unique_item_ids_counts = np.unique(self.item_ids, return_counts=True)
+        if len(set(unique_item_ids_counts)) != 1:
+            raise ValueError(f'Error: Predicted sequences have different lengths {unique_item_ids_counts}.')
```
PGijsbers, line 714: Isn't the more correct test that each predicted sequence should be of length `truth`?

PGijsbers (changes requested):
I think we are almost there :)

amlb/datasets/file.py
PGijsbers: Gotcha, my bad!

amlb/datasets/file.py
```diff
@@ -33,32 +34,25 @@ def __init__(self, cache_dir=None):
     def load(self, dataset, fold=0):
         dataset = dataset if isinstance(dataset, ns) else ns(path=dataset)
         log.debug("Loading dataset %s", dataset)
+        target = dataset['target']
+        type_ = dataset['type']
+        features = dataset['features']
+
+        if DatasetType[type_] == DatasetType.timeseries:
```
PGijsbers, line 41:
```suggestion
        if type_ and DatasetType[type_] == DatasetType.timeseries:
```
`type_` may be `None` which gives a `KeyError` which made the unit tests fail.
Could you also add a unit test in `tests/unit/amlb/datasets/file` that loads a timeseries dataset and some tests that assert the validation logic?

PGijsbers (changes requested):
Hi, sorry for not catching this earlier, but unit tests fail and we should actually also have a few tests added to ensure loading time series datasets doesn't break in the future.

amlb/datasets/file.py
shchur: Thanks! Done

PGijsbers: Sorry for causing so many "round trips" :/

shchur: @PGijsbers I've added tests and force-added a small dataset sample to `tests/unit/amlb/datasets/file/resources/m4_hourly_subset.csv`. Do you think there are any other scenarios we should cover with tests?

PGijsbers: I think we should at least cover the scenario where the dataset's `timestamp` and `id_column` are non-default raise the right errors if not specified, and do load when they are specified as well as test that the loader correctly determines when a dataset has insufficient samples to evaluate the required amount of folds.

shchur: @PGijsbers Done 👍

PGijsbers:
We are currently experiencing some issues with the OpenML server. However, I can confirm that tests pass locally (except for three which explicitly don't use a cache). I'll wait a little longer to see if OpenML issues resolve, but if they do not I'll likely merge tomorrow (and if they do, then I'll run the tests and assuming all passes, merge it too :)).

shchur:
Thank you @PGijsbers, I appreciate your help and support with this PR!

docs/HOWTO.md
```diff
@@ -277,6 +281,69 @@ Then the datasets can be declared in the benchmark definition file as follow:
...
+  1..K | train | train | ... | train | train
+  K..2K | train | train | ... | train | test
+  2..3K | train  | train | ... | test |
+  ... |   |   |     |  
+  (n-2)K...(n-1)K | train  |  test   | |
+  (n-1)K...nK | test  |    | |
+
+- Time series may have different lengths or have different starting timestamps, but need to have the same frequency
+- The shortest time series in the dataset must have length of at least `folds * forecast_horizon_in_step + 1`
+- All time series must have regular timestamp index
```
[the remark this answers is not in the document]
shchur: For example, it is not allowed to have a dataset where one time series has hourly frequency and another has minutely frequency, as in

| item_id | timestamp | target |
|--------|--------|--------|
| A | 2000-01-01 01:00 | ... |
| A | 2000-01-01 02:00 | ... |
| A | 2000-01-01 03:00 | ... |
| B | 2000-01-01 01:01 | ... |
| B | 2000-01-01 01:02 | ... |
| B | 2000-01-01 01:03 | ... |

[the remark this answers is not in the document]
shchur: I've also added a comment to the other PR regarding the column names https://github.com/PGijsbers/automlbenchmark/commit/9f6e95000f1412001e9e2041184196e3edab0cad#r122106033

PGijsbers:
Merged! Thank you @shchur (and also the people of the original PR) for being patient with us :)

PGijsbers:
Just a heads up, I will be out of office in August so I may not respond in that period :)

Innixma:
Thanks @PGijsbers for all the help provided in the reviews!! Excited to have this merged in.

PR #572, opened 2023-07-20
Update README.md

Innixma:
- Fix minor typo in README
- Enforce `item_id` to always be categorical in time series tasks

### DIFF
--- a/amlb/datasets/file.py
+++ b/amlb/datasets/file.py
@@ -357,7 +357,7 @@ class TimeSeriesDataset(FileDataset):
 
         # Store repeated item_id & in-sample seasonal error for each time step in the forecast horizon - needed later for metrics like MASE.
         # We need to store this information here because Result object has no access to past time series values.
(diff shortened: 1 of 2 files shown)


PGijsbers:
Good catch, but shouldn't all instances actually now just be `autogluon`? The exec script has been updated to switch based on dataset type. If that's correct, please update the other instances and then I'll merge :)

Innixma:
Good point, @shchur could you confirm?

shchur:
@Innixma yes, I can confirm. I've opened a PR with changes to `Innixma:patch-1` with updated `README.md` that also includes a small bugfix to ensure that the `item_id` is always parsed as a category. Feel free to move those changes to this PR.

Innixma:
@PGijsbers Updated the PR with @shchur's changes. Should be ready for review/merge.

PGijsbers (approved):

PGijsbers:
Thanks! :)

Issue #573, opened 2023-07-21
503 Error when downloading dataset


eddiebergman:
I'm getting an issue with 503 errors from the openml servers when using the pinned requirement of `0.13.1`, where it can't download the tasks from openml

eddiebergman:
Forcing an update to `0.14.1` doesn't work either due to renamed variables.

PGijsbers:
OpenML is experiencing some server issues, and while they last `openml==0.13.1` will not be able to download new datasets. I applied a patch to the benchmark (#579) that makes the benchmark compatible with newer versions of openml which can download data despite the server issues. This makes openml fall back on loading arff files, which in some cases are different and thus **may lead to different results**. Be deliberate about whether upgrading or waiting is the right choice for you! If you do decide to upgrade simply update your openml python installation: `python -m pip install --upgrade openml`. You should now be able to run the benchmark again on new datasets.

suzhoum:
Thanks for working on the fix! I'm currently testing on the merged fix, on both my dev machine and in docker container that I built that's dependent on AMLB. It works fine with on my dev machine, but when running in docker container, I encountered a series of git errors after the benchmark run has started, which I didn't see before. Is there anything that might have led to the error?

```
Running benchmark `AutoGluon:stable` on `small` framework in `local` mode.
Loading frameworks definitions from ['/app/ag_bench_runs/tabular/ag_bench_test/automlbenchmark/resources/frameworks.yaml'].
Loading benchmark constraint definitions from ['/app/ag_bench_runs/tabular/ag_bench_test/automlbenchmark/resources/constraints.yaml'].
Loading benchmark definitions from /app/ag_bench_runs/tabular/ag_bench_test/automlbenchmark/resources/benchmarks/small.yaml.
fatal: not a git repository (or any of the parent directories): .git
```

PGijsbers:
Does the script stop after that? If not, then I think this is expected behaviour (though something we might want to change):
when building the docker image [files are copied](https://github.com/openml/automlbenchmark/blob/76edcf3ad7e97ce4af6e4cad73c66f8f79ed5b4d/amlb/runners/docker.py#L161C2-L161C2) into the image. The `.git` files are not copied into the repository. This means that the version of the benchmark that exists in a container  is not a git repository. When the benchmark is running inside docker it runs as if it is doing a local run, so it tries to record any information about the git environment - in this case it can't fine any and reports that. In the end, the git information of the data present on the host machine should be recorded for the experiment.
[...]

PR #574, opened 2023-07-21
deps(openml): update to 0.14.1

eddiebergman:
Please see #573

Not sure where else this might need to be updated or how to test this properly, this was just a local fix that got it to work.

### DIFF
--- a/amlb/datasets/openml.py
+++ b/amlb/datasets/openml.py
@@ -39,7 +39,7 @@ class OpenmlLoader:
     def __init__(self, api_key, cache_dir=None):
         oml.config.apikey = api_key
         if cache_dir:
-            oml.config.set_cache_directory(cache_dir)
+            oml.config.set_root_cache_directory(cache_dir)
 
         if oml.config.retry_policy != "robot":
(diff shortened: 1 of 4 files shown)


PGijsbers:
I am a bit hesitant to bump the python version, especially since the ARFF files possibly have different types than the parquet files. I think I would prefer a solution where the code works with either, and not bump the requirements.

E.g., through something like:
```python
try:
  set_openml_cache = openml.config.set_cache_directory
except AttributeError:
  set_openml_cache = openml.config.set_root_cache_directory
...
set_openml_cache(...)
```

That would allow individuals to bump their openml python versions and for it to work out of the box, but they would need to be explicit about this. We can pin your issue with an explanation about this.

Innixma:
Thanks @eddiebergman , this worked to fix the problem on my end! I understand Pieter's reasoning for waiting on the merge though. That makes sense.

PGijsbers:
Closing this because I added the fix in #579 in a way that it stays compatible with openml==0.13.1.
Reasoning for not bumping openml given above.

PR #575, opened 2023-07-21
Fix errors for the remainder of `amlb` module.

PGijsbers:
Opted out of one exception: there is a special case for the recovery script where `run_mode` is "script". This makes some fields optional, which then introduces some ~10 errors because in the normal workflow they are _not_ optional and this is assumed in the code. I think it is probably possibly to refactor the recovery benchmark mode into something that is not actually a `Benchmark` as it doesn't really seem to use the internals that much in fetch mode. For reconnect it later acts as `script`, so there is probably some room to simplify the workflow and address the underlying issue.


amlb/runners/singularity.py
```diff
@@ -85,13 +85,11 @@ def _start_container(self, script_params=""):
         for d in [in_dir, out_dir, custom_dir]:
             touch(d, as_dir=True)
         script_extra_params = "--session="  # in combination with `self.output_dirs.session` usage below to prevent creation of 2 sessions locally
-        inst_name = self.sid
```
PGijsbers, line 88: @franchuterivera I removed some variables that didn't seem to be used. From `singularity run` docs I also didn't see a way to set a name for a container, so I think this is likely just left over from a copy-paste. Do you think the modifications I made here are correct?

PGijsbers: @eddiebergman I am not sure if Francisco is still reachable through Github, but since you likely also use Singularity perhaps you can chime in. It would be much appreciated.

eddiebergman: I unfortunatly don't use singularity, perhaps @neeratyoy or @mfeurer can chime in?

amlb/runners/aws.py
```diff
@@ -1211,68 +1213,6 @@ def _ec2_startup_script(self, instance_key, script_params="", timeout_secs=-1):
             timeout=timeout_secs if timeout_secs > 0 else rconfig().aws.max_timeout_seconds,
         )
 
-    def _ec2_startup_script_bash(self, instance_key, script_params="", timeout_secs=-1):
```
PGijsbers, line 1214: this wasn't used

amlb/runners/aws.py
```diff
@@ -1290,10 +1230,6 @@ def run(self, save_scores=False):
         super().run(save_scores)
         self._upload_results()
 
-    def run_one(self, task_name: str, fold, save_scores=False):
```
PGijsbers, line 1293: this wasn't used

amlb/runners/singularity.py
PGijsbers: to reviewers: you can ignore this conversation if you're not familiar.

Neeratyoy: hope not too late
I do work with Singularity but will not claim I understand the whole setup completely
though I have mainly been building containers and already had code to pull and run it
in our current project, we pull the container using its name and source 
and subsequently _start_ the singularity instance by name [here](https://github.com/automl/HPOBench/blob/master/hpobench/container/client_abstract_benchmark.py#L298)

PGijsbers: Thanks for having a look! The question here specifically is whether the `inst_name` is likely a left-over from copy-pasting docker code and would have no function for a `singularity run` command. It's currently not included in the command, and I can not see a way in the singularity docs to set a "name" for an instance. (not sure if using the right terminology for singularity, but the _image_ from which a container is made is still specified, of course).

PR #600, opened 2023-10-19
Fix minor bugs in time series forecasting

shchur:
Fix several minor bugs in the time series benchmark:
- ensure that `item_id` column is parsed as a string. If the column is parsed as an integer, it may lead to a different sorting order, which in turn introduces mistakes to seasonal error calculation.
- make sure that AutoGluon respects the random seed provided in the config
- add `s3fs` & `fsspec` dependencies that are necessary to enable pandas to load datasets from S3

### DIFF
--- a/amlb/datasets/file.py
+++ b/amlb/datasets/file.py
@@ -343,6 +343,8 @@ class TimeSeriesDataset(FileDataset):
         self.id_column = config['id_column']
         self.timestamp_column = config['timestamp_column']
 
+        # Ensure that id_column is parsed as string to avoid incorrect sorting
+        full_data[self.id_column] = full_data[self.id_column].astype(str)
         full_data[self.timestamp_column] = pd.to_datetime(full_data[self.timestamp_column])
         if config['name'] is not None:
             file_name = config['name']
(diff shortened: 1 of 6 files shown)


amlb/datasets/file.py
```diff
@@ -343,6 +343,8 @@ def __init__(self, path, fold, target, features, cache_dir, config):
         self.id_column = config['id_column']
         self.timestamp_column = config['timestamp_column']
 
+        # Ensure that id_column is parsed as string to avoid incorrect sorting
+        full_data[self.id_column] = full_data[self.id_column].astype(str)
```
PGijsbers, line 347: Sorry, I have been swamped lately. Hope to have a closer look soon (also the other PR).
In the current unit tests, we test for
```python
assert pat.is_categorical_dtype(ds._dtypes[ds.id_column])
```
[here](https://github.com/openml/automlbenchmark/blob/6872b63f3ebc6d7fe38ebc7376ba27c13088c78e/tests/unit/amlb/datasets/file/test_file_dataloader.py#L295C5-L295C62)

Shouldn't that test be affected?

shchur: Thanks, that's a good catch! I've updated the test.

PR #604, opened 2024-01-08
Update AutoGluon to 1.0 API

Innixma:
Made a few updates to align AMLB with AutoGluon 1.0's API.

Note that all changes are backwards compatible. For AutoGluon 1.0 exclusive logic, I added if/else checks so it continues to work with older versions: Further, the mainline AMLB code works with AutoGluon 1.0 already as old methods were deprecated but not removed, it will just log deprecation warnings.

Changes related to AutoGluon 1.0:

1. `persist_models` -> `persist`
2. `persist` max_memory=0.4 -> removed as it is now the default
3. `predictor.predict_from_proba` to allow for decision threshold calibration (not enabled by default)
4. Remove `silent=True` to `leaderboard` call, as it is now silent by default.

General improvements unrelated to AutoGluon 1.0:

1. Allowed `config.max_runtime_seconds` to be None when `good_quality` or `high_quality` preset is specified and avoid crashing.
2. `set -e`: Ensure install script crashes when an error occurs rather than continuing (previously could lead to corrupted installs sneaking by and being ran on datasets, now they should error properly)
3. Switch repo location from `awslabs/autogluon` to `autogluon/autogluon` (it was migrated last year, `awslabs/autogluon` redirects to `autogluon/autogluon`)

### DIFF
--- a/frameworks/AutoGluon/exec.py
+++ b/frameworks/AutoGluon/exec.py
@@ -51,7 +51,7 @@ def run(dataset, config):
     time_limit = config.max_runtime_seconds
     presets = training_params.get("presets", [])
     presets = presets if isinstance(presets, list) else [presets]
-    if (preset_with_refit_full := (set(presets) & {"good_quality", "high_quality"})) & (time_limit is not None):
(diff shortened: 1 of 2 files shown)


PGijsbers (approved):
Sorry for the delay

PR #611, opened 2024-03-06
Add time series forecasting support

### LINKED ISSUE #610: Time series forecasting support for FEDOT
As a demonstration of the concept for time-series benchmarking #494, more autoML frameworks capable of predicting time series should be provided with benchmarking support. 

This issue is dedicated to implementing time-series benchmarking support for the [FEDOT](https://github.com/aimclub/FEDOT) framework.

Lopa10ko:
WIP

closes #610


frameworks/FEDOT/exec_ts.py
```diff
@@ -0,0 +1,151 @@
...
+    id_column = dataset.id_column
+
+    log.info('Predicting on the test set.')
+    training_duration, predict_duration = 0, 0
+    models_count = 0
+    truth_only = test_df[dataset.target].values
+    predictions = []
+
+    for label in train_df[id_column].unique():
+        train_sub_df = train_df[train_df[id_column] == label].drop(columns=[id_column], axis=1)
```
shchur, line 45: This can be very inefficient for large dataframes with many time series (scales as O(N^2)). A better option would be to use groupby, e.g.
```python
for label, ts in train_df.groupby(id_column, sort=False):
    train_series = ts[dataset.target].to_numpy()
```

frameworks/FEDOT/exec_ts.py
```diff
@@ -0,0 +1,151 @@
...
+        train_series = np.array(train_sub_df[dataset.target])
+        train_input = InputData(
+            idx=train_sub_df.index.to_numpy(),
```
shchur, line 48: This code uses the range index `[0, 1, 2, ..., n-1]` instead of the timestamps of the time series. Should this be `train_sub_df[timestamp_column]` instead, or does FEDOT ignore the timestamps?

frameworks/FEDOT/exec_ts.py
```diff
@@ -0,0 +1,151 @@
...
+            features=train_series,
+            target=train_series,
+            task=task,
+            data_type=DataTypesEnum.ts
+        )
+
+        test_sub_df = test_df[test_df[id_column] == label].drop(columns=[id_column], axis=1)
+        test_series = np.array(test_sub_df[dataset.target])
```
shchur, line 56: This contains the future values of the time series, I doubt that these values need to be fed to the model at predictions time.

For example, if training data contains time steps `[1, 2, 3, ..., T]` and the goal is to predict `[T+1, ..., T+H]`, then based on your current code
- `train_series` contains timesteps `[1, 2, 3, ..., T]`
- `test_series` contains timesteps `[T+1, ..., T+H]`

My guess is that we need to pass `train_series` as input to both `fit()` and `predict()`

frameworks/FEDOT/exec_ts.py
```diff
@@ -0,0 +1,151 @@
...
+            data_type=DataTypesEnum.ts
+        )
+
+        fedot = Fedot(
+            problem=TaskTypesEnum.ts_forecasting.value,
+            task_params=task.task_params,
+            timeout=runtime_min,
+            metric=scoring_metric,
+            seed=config.seed,
+            max_pipeline_fit_time=runtime_min / 10,
```
shchur, line 71: Why is `/ 10` necessary here?

frameworks/FEDOT/exec_ts.py
```diff
@@ -0,0 +1,151 @@
...
+        models_count += fedot.current_pipeline.length
+
+    save_artifacts(fedot, config)
+    return result(output_file=config.output_predictions_file,
+                  predictions=np.hstack(predictions),
+                  truth=truth_only,
+                  target_is_encoded=False,
+                  models_count=models_count,
+                  training_duration=training_duration,
+                  predict_duration=predict_duration)
```
shchur, line 98: Nit: I think it would be cleaner to remove the line `training_duration, predict_duration = 0, 0` above and just return `training_duration=training.duration, predict_duration=predict.duration` here.

frameworks/FEDOT/exec_ts.py
```diff
@@ -0,0 +1,151 @@
...
+                log.info('Pipeline crashed. Using no-op forecasting')
+                prediction = np.full(len(test_series), train_series[-1])
+
+        predict_duration += predict.duration
+
+        predictions.append(prediction)
+        models_count += fedot.current_pipeline.length
+
+    save_artifacts(fedot, config)
+    return result(output_file=config.output_predictions_file,
```
shchur, line 92: It's necessary to return `dataset.repeated_item_id` and `dataset.repeated_abs_seasonal_error` as `optional_columns` in the result for MASE computation to work correctly (see https://github.com/openml/automlbenchmark/blob/master/frameworks/AutoGluon/exec_ts.py#L63C1-L67C1). 
This is a rather ugly hack that is necessary to make history-dependent metrics like MASE compatible with the AMLB results API.

frameworks/FEDOT/exec_ts.py
Lopa10ko: since we train the FEDOT model for each series (for each label), `training.duration` value isn't cumulative and will only reflect the time spent on the last iteration

frameworks/FEDOT/exec_ts.py
Lopa10ko: generally speaking, this is a small safety measure to ensure that the training time of one pipeline is exactly within the total timeout. the classification and regression #563 uses the same empirical approach. it

frameworks/FEDOT/exec_ts.py
```diff
@@ -0,0 +1,151 @@
...
+def run(dataset, config):
+    log.info("\n**** FEDOT ****\n")
+
+    scoring_metric = get_fedot_metrics(config)
+
+    training_params = {"preset": "best_quality", "n_jobs": config.cores}
+    training_params.update({k: v for k, v in config.framework_params.items() if not k.startswith('_')})
+    n_jobs = training_params["n_jobs"]
+
+    log.info('Running FEDOT with a maximum time of %ss on %s cores, optimizing %s.',
```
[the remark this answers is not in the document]
Lopa10ko: done d583cb2

frameworks/FEDOT/exec_ts.py
shchur: My bad, I didn't realize this was happening inside the loop over individual series

frameworks/FEDOT/exec_ts.py
Lopa10ko: FEDOT ignores timestamps, so that idx should be fine

frameworks/FEDOT/exec_ts.py
Lopa10ko: thanks, tried that: imo using zip may be inappropriate here, so I would stick with a janky bad-asymptotic solution to ensure that the labels are matching
e6f19e7

frameworks/FEDOT/exec_ts.py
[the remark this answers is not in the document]
Lopa10ko: d583cb2

frameworks/FEDOT/exec_ts.py
Lopa10ko: in essence `predict` and `forecast` are the same thing, but we could pass `test_input` with `features` from `train_series` and `target` from `test_series` to get the prediction horizon in a `predict` 

changed to a clear `forecast` method
a357726

frameworks/FEDOT/exec_ts.py
Lopa10ko: c400961

Lopa10ko:
@shchu, thanks for the initial review :)
I intend to continue exploring the benchmarks and testing some things locally in order to ensure that all FEDOT-related functionality is working correctly. This may take some time, so when I'm finished, I will undraft this pull request.

Lopa10ko:
@shchur sorry it took ages to finally return to this PR.
would be great to have a further review

shchur:
Hi @Lopa10ko @PGijsbers, I will try to have a look at the PR today

frameworks/FEDOT/exec_ts.py
shchur: Would it make sense to split the time limit evenly across the series? Right now it seems that 10% of the total time limit is given to each series, which may lead to overruns if >10 series are available.

frameworks/FEDOT/exec_ts.py
```diff
@@ -0,0 +1,149 @@
...
+            task=task,
+            data_type=DataTypesEnum.ts
+        )
+
+        test_sub_df = test_df[test_df[id_column] == label].drop(columns=[id_column], axis=1)
+        horizon = len(test_sub_df[dataset.target])
+
+        fedot = Fedot(
+            problem=TaskTypesEnum.ts_forecasting.value,
+            task_params=task.task_params,
```
shchur, line 60: As far as I understand, Fedot currently only supports point forecasting but AMLB may also include probabilistic forecasting tasks (see https://github.com/openml/automlbenchmark/blob/master/amlb/results.py#L767-L792). Probably it would make sense to raise an exception if someone tries to evaluate FEDOT on such a probabilistic forecasting task.

shchur (approved):
Two small comments that probably need to be addressed, but otherwise looks good to me.

frameworks/FEDOT/exec_ts.py
Lopa10ko: is there a way to distinguish a probabilistic forecasting task based on the benchmark run config?

the `get_fedot_metrics` function already emits logs in case of unsupported metrics (like `mql`, `wql`, and `sql`)

frameworks/FEDOT/exec_ts.py
Lopa10ko: [20dac85](https://github.com/openml/automlbenchmark/pull/611/commits/20dac85c24e5f8dd4ab746eed0b91642bfa3d609)

frameworks/FEDOT/exec_ts.py
shchur: I think that filtering by the mql, wql, sql metrics is the simplest way to accomplish this.

Another option is to repeat the point forecast for each of the quantile levels and raise a warning.

Lopa10ko: [db10d89](https://github.com/openml/automlbenchmark/pull/611/commits/db10d890c698753cf3edafcf12b81691a1d29309)

Lopa10ko:
@shchur thanks for a review, 
@PGijsbers if you have any suggestions for improvements, I would be happy to incorporate them

PGijsbers:
In principle it looks fine. Does FEDOT support provided memory constraints? If so, I would appreciate it if you add that to the integration script (`config.max_mem_size_mb`).

Lopa10ko:
> In principle it looks fine. Does FEDOT support provided memory constraints? If so, I would appreciate it if you add that to the integration script (`config.max_mem_size_mb`).

@PGijsbers sorry, FEDOT does't have the capability to handle memory limitations.

PGijsbers:
Sorry for the delay, I thought this was merged.

Issue #356, opened 2021-07-22
Turn off telemetry of MLNet 


PGijsbers:
By default MLNet has telemetry enabled ([source](https://docs.microsoft.com/en-us/dotnet/machine-learning/resources/ml-net-cli-telemetry#opt-out-of-data-collection)). We want this to be opt-in rather than opt-out for the benchmark, especially since we don't expect all users to extensively read the documentation for each framework.

mcnojo:
I've added the necessary line to opt out by setting the env variable and it runs against the MLNet tests without issue, but is there any desire to add a toggle to the CLI for opting back in?  If so, how should I proceed with this/what resources can give me a better sense of the code?  I've been reading the AutoML functions for a few hours without much progress.

PGijsbers:
Where do you set the environment variable? Through python in `exec.py`? If so, then using a script parameter would be best. These can be set through the [configuration](https://github.com/openml/automlbenchmark/blob/b90cd9b30ff9c9048ae3240ecdece55a0a3da189/resources/frameworks.yaml#L28), and from the `exec.py` are provided through [config.framework_params](https://github.com/openml/automlbenchmark/blob/b90cd9b30ff9c9048ae3240ecdece55a0a3da189/frameworks/AutoGluon/exec.py#L122). This is hidden in the very last [info](https://openml.github.io/automlbenchmark/docs/extending/framework/) block on the framework page - we should make it more visible, and in the future separate out framework parameters from script parameters :)
So, in that case you can make setting the environment variable conditional on the presence of a parameter in the framework configuration. These are still not exposed to the CLI (though invoking the definition which has telemetry ON is available through the CLI, of course).

mcnojo:
Very helpful context, thank you!  The environment variable was originally set in exec.py.

After reading the above, I've changed this to set a framework parameter so that exec.py will pull from the config and set the environment variable conditionally.  

I've opened a pull request [here](https://github.com/openml/automlbenchmark/pull/619#issue-2259391697).

PGijsbers:
Closed by #619

PR #619, opened 2024-04-23
Issue #356, Default Telemetry (Off)

mcnojo:
Added the required framework parameter for MLNet Telemetry, and set the environment variable for MLNet in exec.py to then opt out by default. 

This is the [original issue.](https://github.com/openml/automlbenchmark/issues/356#issue-950480518)

Cheers,
-Noah

### DIFF
--- a/frameworks/MLNet/exec.py
+++ b/frameworks/MLNet/exec.py
@@ -35,6 +35,10 @@ def run(dataset: Dataset, config: TaskConfig):
     train_time_in_seconds = config.max_runtime_seconds
     sub_command = config.type
 
+    # Opt-out of Telemetry via framework parameter
+    MLDOTNET_CLI_TELEMETRY_OPTOUT = config.framework_params.get('telemetry_disabled', 'False')
+    os.environ['MLDOTNET_CLI_TELEMETRY_OPTOUT'] = MLDOTNET_CLI_TELEMETRY_OPTOUT
+
     # set up MODELBUILDER_AUTOML
     MODELBUILDER_AUTOML = config.framework_params.get('automl_type', 'NNI')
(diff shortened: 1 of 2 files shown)


frameworks/MLNet/exec.py
```diff
@@ -35,6 +35,10 @@ def run(dataset: Dataset, config: TaskConfig):
     train_time_in_seconds = config.max_runtime_seconds
     sub_command = config.type
 
+    # Opt-out of Telemetry via framework parameter
+    MLDOTNET_CLI_TELEMETRY_OPTOUT = config.framework_params.get('telemetry_disabled', 'False')
```
PGijsbers, line 39:
```suggestion
    MLDOTNET_CLI_TELEMETRY_OPTOUT = config.framework_params.get('_telemetry_disabled', 'False')
```

resources/frameworks.yaml
```diff
@@ -145,6 +145,8 @@ MLNet:
   version: 'latest'
   description: |
     MLNET.CLI is a automated machine learning tool implemented by ml.net.
+  params:
+    telemetry_disabled: ‘True'
```
PGijsbers, line 149:
```suggestion
    _telemetry_disabled: ‘True'
```

PGijsbers (changes requested):
Thanks! I do think I prefer the parameter to be defined with a leading underscore to denote it as a script parameter (it's our convention). Intuitively, I expect the framework parameters without leading underscore to be forwarded directly to the automl system itself. I think you can argue both ways here, but I am leaning this way :)
Otherwise, LGTM 🚀

PGijsbers:
Thanks again!

Issue #408, opened 2021-10-02
Flaml does not support tag 'stable' or 'latest'


eddiebergman:
Hi,

Do you plan to support these tags for flaml? If not, is there a particular reason?

---
While setting up some infrastructure for automating bench marking, I noticed the following two commands fail:
```bash
python runbenchmark.py flaml:stable
python runbenchmark.py flaml:latest
```

Turns out flaml is not included in `resources/frameworks_{stable,latest}.yaml`.

PGijsbers:
There is no reason that they are not added as far as I am aware, and I see no issue with adding them now (as far as I can tell it's even already supported in [their setup script](https://github.com/openml/automlbenchmark/blob/master/frameworks/flaml/setup.sh)).

PR #626, opened 2024-06-13
Fix FLAML installation problems, support version >=2

### LINKED ISSUE #623: 'NoResultError' raised by FLAML Framework During Benchmark Execution
Hello,

I have just installed AutoML Benchmarck in an environment running Python 3.9.17. However, when I execute the command `python runbenchmark.py flaml`, I encounter the following error:

NoResultError: `best_iteration` is only defined when early stopping is used.

Is there a workaround to resolve this issue?

Thank you.

[error_traceback.txt](https://github.com/openml/automlbenchmark/files/15469087/error_traceback.txt)

PGijsbers:
FLAML was fixed to <2, and would install an incompatible xgboost version. This patch makes it so that newer versions of FLAML can be installed. Additionally, when installing an older version a compatible version of xgboost is installed.
Finally, it adds a definition of flaml:latest in line with the other framework definitions.

Closes #623, #408.

@qingyun-wu as far as I can tell, no changes to the `exec.py` script are necessary. The simple test cases pass.


PGijsbers:
Merging this since I would like the issues to be resolved and at this point it is not clear if we will get the framework authors to chime in.

PR #629, opened 2024-06-25
Add CI to periodically check whether frameworks install and run

PGijsbers:
Sometimes frameworks break because either;
 - a new framework releases is no longer compatible with the integration script, or
 - dependencies of the framework are not pinned, and a newer non-compatible dependency is released

In these cases, we want to automatically be notified.
This PR sets up a basic workflow to accomplish this.
In the future I want to extract the action and release is as a stand-alone action on the marketplace so that AutoML framework authors can also use it in their CI.

### DIFF
--- a/amlb/benchmark.py
+++ b/amlb/benchmark.py
@@ -22,6 +22,7 @@ import sys
 
 import pandas as pd
 
+from .frameworks.definitions import load_framework_definition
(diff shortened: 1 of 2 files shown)


PGijsbers:
Hi Eddie,
If you find a little bit of time to do a sanity check for this, I'd love to hear your thoughts. If not, a heads up about it is appreciated.

PR #638, opened 2024-09-16
Early stopping for AG and inference time enabled for FEDOT

israel-cj:


PGijsbers:
This PR contains a lot of extra changes and files than just allowing early stopping for some frameworks. I'll close this and set up a draft with the minimal changes in a way that I think makes sense for the benchmark. Then we can discuss.

PR #639, opened 2024-09-17
Add generic callback configuration logic

PGijsbers:
You can configure callbacks for AutoGluon through configuration, e.g.: 

```yaml
AutoGluonES:
  extends: AutoGluon
  params:
    callbacks:
      EarlyStoppingEnsembleCallback:
        patience: 5
```

Should probably also be added to the time series script (might be worth at some point trying to see what can be stored in a shared module?)?


Innixma (approved):
LGTM! Thanks for adding. Might be nice if there is some documentation on how to properly specify the callbacks in the yaml file, otherwise it isn't trivial to reverse engineer the proper formatting.

PR #643, opened 2024-10-29
Update AutoGluon to use uv during setup.sh

Innixma:
- Update AutoGluon to use [uv](https://github.com/astral-sh/uv) instead of pip for installation in setup.sh This aligns with mainline AutoGluon now switching to `uv` for its install instructions.

- The switch to `uv` reduces AutoGluon installation time from ~10 minutes to ~40 seconds on EC2. This also avoids [a bug in pip where it fails to install AutoGluon](https://github.com/autogluon/autogluon/issues/4515).

- Removed setuptools and wheel installs, as they aren't needed by `uv` and there is a bug in torch 1.4.x that causes an exception if setuptools is installed at the same time.

- Will now install from source in `non-editable` mode. Previously it would install in `editable` mode, which could cause bugs if running AMLB locally alongside a pre-existing local AutoGluon install. This is now avoided by ensuring that the venv directly uses the code pulled in `setup.sh`.

### DIFF
--- a/frameworks/AutoGluon/setup.sh
+++ b/frameworks/AutoGluon/setup.sh
@@ -3,6 +3,9 @@
 # exit when any command fails
 set -e
 
+# Uncomment for debugging installation
+# set -x
+
 HERE=$(dirname "$0")
 VERSION=${1:-"stable"}
 REPO=${2:-"https://github.com/autogluon/autogluon.git"}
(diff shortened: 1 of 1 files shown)


PGijsbers (approved):
Sorry for the delay. I wanted to try it out locally to understand where `uv` installs its packages if created from a virtual environment to ensure the normal cleanup instructions are still valid. It behaves as I expected so we're good to go :)

PR #644, opened 2024-10-30
Add jobhistory argument to runbenchmark.py

Innixma:
Add jobhistory argument to runbenchmark.py

This allows the user to specify the specific path to the prior results file when resuming a run, rather than it being tied to the expected output file path on the local machine (which isn't the case when running via AWS mode and not continuing to use the original host instance)

For example, from s3:

```
python runbenchmark.py AutoGluon:stable test test --resume --jobhistory s3://automl-benchmark-ag/aggregated/ec2/2024_10_25/results.csv
```

If the user does not specify `jobhistory`, then the logic remains identical.

I also did some minor code cleanup in `runbenchmarks.py` to reduce code dupe.

### DIFF
--- a/amlb/benchmark.py
+++ b/amlb/benchmark.py
@@ -21,6 +21,8 @@ import signal
 import sys
 from typing import List, Union
 
+import pandas as pd
+
 from .job import Job, JobError, SimpleJobRunner, MultiThreadingJobRunner
 from .datasets import DataLoader, DataSourceType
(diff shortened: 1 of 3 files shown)


PGijsbers (commented):
Overall looks good to me. Some minor suggestions.

Innixma:
@PGijsbers updated, thanks!

Issue #642, opened 2024-10-05
Defaults to improve framework integration setup


PGijsbers:
I started looking at our framework integration requirements (in preparation for external integrations), and the setup is a bit cumbersome. Some proposals: 

### 1. Default for absent `setup` in `__init__.py`
A function `setup` is required to be defined in `__init__.py`, in _every_ integration this has the same implementation:
 ```python
def setup(*args, **kwargs):
       call_script_in_same_dir(__file__, "setup.sh", *args, **kwargs)
```
we could make it the default behavior, removing the need to add this to `__init__.py`.

### 2. Exit on error `setup.sh`
The setup scripts can be finicky and produce a lot of output. The automl benchmark can't (always) check whether an installation is successful. Errors easily get lost in logs. For all these reasons, it can be easy to miss a framework failed to install successfully. Even though there is (almost?) never a reason you should be OK with the setup script failing and continuing to try and run an evaluation. For that reason, many scripts already include a `set -e`. I think the default should be for the caller to set this mode (integration scripts can always undo it with `set +e`). Especially useful for people working on new integrations that might overlook this otherwise.

### 3. Shared setup should be run before framework specific`setup.sh`
Currently, it is the burden of the integration script to call the shared installation. The main reason for this is, I think, so they can specify whether or not a virtual environment should be created (and have PY and PIP set accordingly). I think whether or not a virtual environment should be created should become a configuration in the `frameworks.yaml` file, and default to yes. With the option being put in the configuration, AMLB can now setup the shared environment and _then_ call the framework-specific setup script. Further reducing the boilerplate, and also making it easier since that "boilerplate" still required knowing where your framework integration lived relative to the setup scripts (by sourcing another file).

---
Interested to hear your thoughts. Do you see any obstacles? Better ways? Additional changes?

(@eddiebergman, @Innixma)

Innixma:
> I think whether or not a virtual environment should be created should become a configuration in the frameworks.yaml file

+1, this would be great to simplify debugging and development. Easy ways to run directly with the calling environment and skip creating a new one will make things lighter when someone wants to run a debugger through the entire logic.

Overall LGTM

Issue #279, opened 2021-04-06
define abstractions for framework runner


sebhrusen:
The goal is to provide an abstraction and default implementation(s) for most common scenarios.
This would also allow frameworks to support several versions easily.
Finally, and more structured framework runner will simplify the integration effort and standardize support for extra features like the `_save_artifacts` param.

1st suggestion (incomplete, and will change):
```python
class FrameworkRunner:
    def __init__(self, config, dataset): pass
    def prepare_data(self): pass
    def fit(self, …): pass
    def predict(self, …): pass
    def get_result(self): pass
    def save_artifacts(self): pass
```

PGijsbers:
Based on our discussion, we should include some type of recovery mode.

PGijsbers:
With this refactor, it will also be easier to use the various "checkpoints" to store partial results and/or have more dynamic time cut-offs. For example, the one hour time limit could be more strictly enforced for just the `fit` call, while being (much) more lenient in phases after `fit`, as compared to having a single large budget for all phases combined. This will avoid both scenarios where EC2 instances live needlessly long because they get hung in a `fit` call and those were results are incompletely merely because the `predict` part took longer.

PR #649, opened 2024-11-17
Add hack that allows specifying the user directory as git

PGijsbers:
Minimal hack to easily define frameworks and configurations remotely (e.g., on a GitHub repository). 
The repository must have all the files you would expect for adding a custom framework locally, like [this one](https://github.com/pgijsbers/amlb-template). With the hack we just identify if the custom directory is a git repository, and if so clone it first, then set the local directory as the user directory.

```commandline
python runbenchmark.py ExplainableBoostingMachine -u https://github.com/pgijsbers/amlb-template.git
```

@Innixma @eddiebergman Where do you think we should go from here? I think a few things might be nice to have:

 - specifying where repositories should be cloned to through a configuration option (instead of "downloads").
 - automatically pulling the latest changes from the remote
 - allowing branches/commits to be specified
 - some metadata file to communicate compatibility with different AMLB versions (not quite necessary yet, but I hope to revise some of the requirements in the future as e.g., #642, #279, some of which _require_ changes to the setup scripts). Would be better to error out early.
 - specify multiple user directories (e.g., you might have a framework integration in one, and some definitions in another)
 - add information to the logs and results file on which framework integration was used.
 - externalize most (if not all?) frameworks from this repository (#571)
 - modifying the `frameworks.yaml` to allow us to specify integration scripts of frameworks, so that you can  `python runbenchmark.py FRAMEWORK` and it can then automatically fetch the integration for `FRAMEWORK` if it is one of the "trusted" predefined ones in the shipped `frameworks.yaml`?
 
Are there any additions or concerns that come to mind? Love to hear your thoughts.

### DIFF
--- a/runbenchmark.py
+++ b/runbenchmark.py
@@ -4,6 +4,7 @@ import os
 import re
 import shutil
 import sys
+from pathlib import Path
 
 # prevent asap other modules from defining the root logger using basicConfig
 import amlb.logger
@@ -11,7 +12,7 @@ import amlb.logger
 import openml
 
 import amlb
-from amlb.utils import Namespace as ns, config_load, datetime_iso, str2bool, str_sanitize, zip_path
+from amlb.utils import Namespace as ns, config_load, datetime_iso, str2bool, str_sanitize, zip_path, run_cmd
 from amlb import log, AutoMLError
 from amlb.defaults import default_dirs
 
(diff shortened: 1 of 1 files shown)


eddiebergman:
Heyo,

I definitely mentioned [1] as something that would be a nice feature to have for benchmarking _other_ peoples algorithms, as they intend their setup to be.

Aside from some of the nice features, I'm on board with most of them. I think one of the priorities would be the specifying the commit-sha, optionally even having a recommended one in the framework's own definition (i.e. "if benchmarking, and you want the _latest_ use this version for now while we work on main"). As for the hack of setting it to the user-directory, that kind of works but ultimately the goal of having the `.amlb-setup` in a repo that other can re-use is just to have the framework definition as well as setup. The actual datasets and runtime configurations would not be needed and would be defined by the user actually benchmarking the tool with the `.amlb-setup` (or whatever it would be called).

As for the slower setup times, I'm not that concerned as usually:
1. Once the framework is setup once, it usually does not need to be setup again, or at least rarely.
2. The setup time for downloading the datasets usually dominates.

PGijsbers:
Thanks for your thoughts! While specifying specific repo versions with commit hashes remains useful/important, I would say the

> "if benchmarking, and you want the latest use this version for now while we work on main"

is what a stable branch or otherwise releases should be for. In practice, I don't think people would want to update commit hashes every time they decide something is stable.

For the subdirectory, _other_ people could also have two repositories: one with the framework, one with the integration. Though I guess that also requires some additional work to ensure the repositories are in sync by e.g., add it in CI or introducing some versioning constraints in the integration repo.

> ther can re-use is just to have the framework definition as well as setup

Yes, so either the framework integration/definition would be either its own option, or we could allow for arbitrary many user dirs to be defined, where all configs files are simply merged.

PR #655, opened 2024-11-26
Add code coverage to CI

PGijsbers:
Note that this only covers (some) of the unit tests ~, and none of the test matrix that runs the complete benchmark workflows~.


PGijsbers:
There may be some weird interaction where `codecov` is not available if the cache is used.

PR #659, opened 2024-11-28
Remove from_config and only keep from_configs

PGijsbers:
Working towards a moving target at #658, I am splitting off individual cleaned up commits for easier reviews.

This PR removes the `from_config` function since it is interchangable with the more flexible `from_configs`.
I could detect no use for the `runscores` script, and it didn't even work, so I took the liberty to also remove that.


PGijsbers (commented):
Few actual changes, mostly sanity check and also wondering if you used `runscores` in any way.

PGijsbers:
@eddiebergman I was doing the "round-robin" reviewer assignment, but it seemed like Nick might be on vacation. So to keep things moving I'm assigning you instead. Hope that's alright with you.

eddiebergman:
Looks good to me, thanks for highlighting the actual changes. Not familiar with this functionality really, but makes sense from a high-level overview.

One recommendation is to have a blanket ruff formatting PR to make future changes easier to review :)

eddiebergman (approved):

PGijsbers:
Thanks for checking!

PR #665, opened 2024-12-04
Ruff formatting

PGijsbers:
Here's the ruff formatting change you suggested.
I also made some changes for the yaml checker, but otherwise didn't touch it.
Most notably I did not resolve any of the pre-commit mypy/ruff linter errors.


PGijsbers:
Nevermind, will need a closer look tomorrow as the formatting broke the codebase.

eddiebergman:
I can't really review this as you can imagine. However I did take a look for `ruff` configuration and didn't find anything in the `pyproject.toml`.

I'm a fan of having defaults on, but there are likely exceptions you'd like to make w.r.t. to applying this an older codebase, also worth at least specifying the python version to ruff so it can take that into account with respect to its rule set from [`pyupgrade`](https://docs.astral.sh/ruff/rules/#pyupgrade-up). You can use [here](https://github.com/automl/amltk/blob/5221d80fb4fec92f2d6f2fc73fce4eb16139c09f/pyproject.toml#L106) as a reference if you'd like.

PGijsbers:
Thanks. I wasn't expecting you to review the formatting changes themselves, rather to see if there were other points of attention, like the ones you pointed out :) Sorry for not being clear about that, and thanks for your comment.

Issue #566, opened 2023-07-18
Improving the developer experience - what do we need?


PGijsbers:
I think it is time we invest into the developer experience before adding more features. To open the issue I am sharing my thoughts on where we should put our effort, but I welcome feedback and suggestions! I do not propose to halt all effort in other directions,  but I think contributors that want to focus on the longevity of the benchmark should ideally focus their efforts here first. My hope is that by addressing these issues, future features and frameworks will be easier to add and quicker to review.

In my opinion, this project is missing several aspects that I would expect from a modern Python project:

 - [ ] Type hints
 - [ ] pre-commit for auto-formatting and linting (e.g., `ruff`, `black`)
 - [ ] Good unit test coverage. Fast isolated tests for things like data ingestion, configuration parsing, and result processing. Due to the stochasticity and time consuming process of installing framework tools, full integration tests are probably left for CI by default.
 - [ ] Clear contribution guidelines
 - [ ] Step-by-step tutorials for different ways to extend the benchmark (meaning benchmark, constraint, framework)

With that in place (or at least the first three), I think the following refactorings should be highest priority:

 - [ ] Refactor the use of `Namespace`s everywhere. To me it makes it much harder to see what information functions expect from their signatures, and at times code has to be carefully analyzed (or checked at runtime) to determine the exact content of arguments. I propose to replace (most) `Namespace`s by dataclasses or pydantic models, whenever dynamic assignment isn't necessary. Might need to use something like [Pydantic's TypedDict](https://docs.pydantic.dev/latest/usage/types/dicts_mapping/#typeddict) to support merging out of the box.
 - [ ] Break down large functions and modules that contain multiple classes and responsibilities (e.g., `amlb/runners/aws.py` or `amlb/datasets/file.py` into multiple submodules to make the code easier to navigate.
 - [ ] Prefer immutable to mutable.
 - [ ] https://github.com/openml/automlbenchmark/issues/279 which also allows things like recording intermediate results. E.g., train->save artifacts->test->inference time, would more reliably provide results while still attempting to measure inference time.

PGijsbers:
In the same line, we could reduce developer overhead by externalizing the framework integrations: https://github.com/openml/automlbenchmark/discussions/571

This hopefully has the benefit that framework authors claim more ownership of their integration and ensure it stays updated themselves.

mfeurer:
I fully agree on everything you write and the plan you laid out here! This will make AMLB easier to understand, use, and contribute to.

[...]

PR #667, opened 2024-12-08
Separate out framework def as dataclass

PGijsbers:
This PR defines a Framework dataclass in favor of using a Namespace. The main advantages are type checks, IDE support, and validation. No real functional differences.

Because namespaces were used with "attribute access" (the `.` notation), switching out a namespace for Framework instances doesn't require very invasive changes.
Unfortunately, CI is a bit broken currently (AutoGluon used to be a reliable indicator until v1.2 broke that). But the unit tests pass, Random Forest passes, and AutoGluon-cholesterol passes, so I feel fairly confident about the change.

Part of #658 on a way to a better dev experience (#566).

@mfeurer It's been a long time since you agreed to do occasional PR reviews for AMLB. I understand if things changed and you are no longer available. If so, no hard feelings - just leave a message.

### DIFF
--- a/amlb/frameworks/definitions.py
+++ b/amlb/frameworks/definitions.py
@@ -1,10 +1,16 @@
+from __future__ import annotations
+
 import copy
 import itertools
 import logging
 import os
-from typing import List, Optional, Union
+from dataclasses import dataclass, field
+from typing import List, Optional, Union, TYPE_CHECKING
 
 from amlb.utils import Namespace, config_load, str_sanitize
+
+if TYPE_CHECKING:
+    from amlb import Resources
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


amlb/frameworks/definitions.py
```diff
@@ -233,3 +239,45 @@ def _remove_frameworks_with_unknown_parent(frameworks: Namespace):
...
+    setup_args: list[str] = field(default_factory=list)
+    # more optionals
+    params: dict = field(default_factory=dict)
+    refs: list = field(default_factory=list)
+    description: str | None = None
+    project: str | None = None
+
+    def __post_init__(self):
+        if isinstance(self.image, dict):
+            self.image = Image(**self.image)
```
PGijsbers, line 273: This helps where framework namespaces were initiated based on nested dicts (loaded from configuration files).

mfeurer (approved):

PR #668, opened 2024-12-11
Simplify and test container image names

PGijsbers:
Originally there was a lot of duplication between the general image name generation and the one that was "re-implemented" in the singularity runner. This refactoring simplifies that and re-uses the common implementation of container in the singularity class. Moreover, this simplification does away with the ugly `kwargs` that were added just to support custom implementations in the inheriting classes (as_docker was only ever used in calls to `_container_image_name` as opposed to `image_name`).


eddiebergman (approved):
LGTM

Issue #664, opened 2024-12-04
bug: Error with logging


eddiebergman:
I'm sporadically encountering an error in logging which boils down to this line, in which it states that `ValueError: I/O operation on closed file`. It's not conclusive and I can't make a reproducing script. Just rerunning the failed jobs seems to work ~99% of the time.

https://github.com/openml/automlbenchmark/blob/aded7c7ada67fb2b7871cbd2b461a0fb93ec8baa/amlb/logger.py#L98

I'm not sure where, but somehow this error cascades into a recursion error as the error keeps trying to get logged (perhaps in the overwritten `print`?).

The initial trigger seems to come from inside the `.framework_module.run()`, which then cascades several times. I can say for certain that

https://github.com/openml/automlbenchmark/blob/aded7c7ada67fb2b7871cbd2b461a0fb93ec8baa/amlb/benchmark.py#L602-L613

This is a trimmed down version of the `exec.py` file:

```python
from __future__ import annotations

# Yup, don't ask, this is required, and in this order
# https://github.com/pypa/setuptools/issues/3044
import setuptools  # noqa: F401
import pip  # noqa: F401

import logging

from typing import TYPE_CHECKING
from frameworks.shared.callee import call_run, result
import pandas as pd
import time

log = logging.getLogger(__name__)
log.setLevel(logging.INFO)

if TYPE_CHECKING:
    from amlb.benchmark import TaskConfig
    from amlb.data import Dataset

def run(dataset: Dataset, config: TaskConfig):
    log.info("\n**** OMITTED ****\n")

    is_classification = config.type == "classification"

    # Helpful to debug locally
    print("cat indices:", dataset.categorical_indices)
    print("feature names:", dataset.train.X.columns)

    if is_classification:
        model = OMITTED
    else:
        model = OMMITED

    X_train, X_test = dataset.train.X, dataset.test.X
    y_train, y_test = dataset.train.y, dataset.test.y
    # Turn into series if needs be
    if isinstance(y_train, pd.DataFrame):
        y_train = y_train.squeeze()
    if isinstance(y_test, pd.DataFrame):
        y_test = y_test.squeeze()

    now = time.monotonic()
    model.fit(X_train, y_train)
    train_duration_seconds = time.monotonic() - now

    log.info(f"Finished fit in {train_duration_seconds}s.")

    # 2x forwards
    probabilities = model.predict_proba(X_test) if is_classification else None

    now = time.monotonic()
    predictions = model.predict(X_test)
[...]

PR #672, opened 2024-12-13
Check if stream is already closed

PGijsbers:
Relates to #664 

Actually - perhaps the use of `io.StringIO` is a bit overkill here? How many print statements without newline are we expecting realistically? We could also eat the overhead and use regular string concatenation (still wouldn't solve root cause).

### DIFF
--- a/amlb/logger.py
+++ b/amlb/logger.py
@@ -103,7 +103,7 @@ def setup(
             nonlocal buffer
             buf_type = "err" if file is sys.stderr else "out"
             buf = buffer[buf_type]
-            if buf is None:
+            if buf is None or buf.closed:
                 buf = buffer[buf_type] = io.StringIO()
             line = sep.join(map(str, [*args]))
             buf.write(line)  # "end" newline always added by logger
--- a/resources/config.yaml
+++ b/resources/config.yaml
@@ -22,7 +22,9 @@ seed: auto              # default global seed (used if not set in task definitio
 token_separator: '.'    # set to '_' for backwards compatibility.
                         # This separator is used to generate directory structure and files,
                         # however the '_' separator makes the parsing of those names more difficult as it's also used in framework names, task names...
-archive: ['logs']       # list of output folders that should be archived by default.
+archive: []       # list of output folders that should be archived by default.
+                  # Note that when used in a parallel setting where the same framework is invoked multiple times in
+                  # the same second, archiving directories may lead to errors due to race conditions (#672).
 
 setup:                   # configuration namespace for the framework setup phase.
   live_output: true      # set to true to stream the output of setup commands, if false they are only printed when setup is complete.


LennartPurucker:
I just ran a bigger benchmark with this PR, but the bug/error still occurs for ~10% of runs with this fix.
But I got a longer error message now, the same as shared by Eddie. 

I am not quite able to follow the issue's conclusions, but I can try to log into two files next.
As I have a unified separate log file from the cluster management anyhow, this is not a downside to me.

PGijsbers:
Thanks for letting me know! It's a bit messy as it's an ongoing investigation, but the current assumption is still that:
 - logging from two different processes can lead to an error while logging
 - when this error occurs, Python internals simply `print` something.
 - since we patch the `print` statement to write to a log instead, and this gives an error
 - we are now in an infinite loop until max stack depth is exceeded
 
The only problem is that I could not reproduce the race condition of the first step to confirm this. But if your next experiment with two logs does seem to mitigate the issue, then I think it's a fair assumption that we understand the root cause.

PGijsbers:
And just to verify, your new stack trace still includes `ValueError: I/O operation on closed file`?

LennartPurucker:
> And just to verify, your new stack trace still includes ValueError: I/O operation on closed file?

Mhm, no, not as far as I can tell. The starting point currently is 

`OSError: [Errno 116] Stale file handle`

PGijsbers:
Thanks, so it's not exactly the same then :) because the error I asked about should be the one that this branch prevents (but since there are other root causes, it does not solve the eventual termination of the script). If you can share a stack trace I would appreciate that.

LennartPurucker:
Ah okay, it is the same one as reported by Eddie in #664:


```
--- Logging error ---
Job `local.openml_t_190392.4h8c.madeline.0.AutoGluonOursTS14022025` failed with error: maximum recursion depth exceeded
  File "/usr/lib/python3.10/logging/__init__.py", line 1104, in emit
    self.flush()
  File "/usr/lib/python3.10/logging/__init__.py", line 1084, in flush
    self.stream.flush()
OSError: [Errno 116] Stale file handle

During handling of the above exception, another exception occurred:

[... repeated in various forms and orders ...]

  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/amlb/benchmark.py", line 581, in run
    meta_result = self.benchmark.framework_module.run(self._dataset, task_config)
  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/frameworks/AutoGluon/__init__.py", line 16, in run
    return run_autogluon_tabular(dataset, config)
  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/frameworks/AutoGluon/__init__.py", line 33, in run_autogluon_tabular
  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/frameworks/shared/caller.py", line 134, in run_in_venv
    output, err = run_cmd(cmd, *args,
  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/amlb/utils/process.py", line 255, in run_cmd
    completed = run_subprocess(str_cmd if params.shell else full_cmd,
  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/amlb/utils/process.py", line 85, in run_subprocess
    stdout, stderr = communicate(process, input, timeout=timeout)
  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/amlb/utils/process.py", line 80, in communicate
    return (communicate_fn(process, input=input, timeout=timeout) if communicate_fn
  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/amlb/utils/process.py", line 249, in communicate
    res = h(*args, **kwargs)
  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/amlb/utils/process.py", line 179, in live_output_unix
    process_output = list(iter(
  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/amlb/utils/process.py", line 180, in <lambda>
    lambda: read_pipe(
  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/amlb/utils/process.py", line 173, in read_pipe
    print(re.sub(r'\n$', '', line, count=1))
  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/amlb/logger.py", line 102, in new_print
    print_logger.log(level, buf.getvalue())
  File "/usr/lib/python3.10/logging/__init__.py", line 1547, in log
    self._log(level, msg, args, **kwargs)
  File "/usr/lib/python3.10/logging/__init__.py", line 1624, in _log
    self.handle(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 1634, in handle
    self.callHandlers(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 1696, in callHandlers
    hdlr.handle(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 968, in handle
    self.emit(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 1218, in emit
    StreamHandler.emit(self, record)
  File "/usr/lib/python3.10/logging/__init__.py", line 1108, in emit
    self.handleError(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 1022, in handleError
    traceback.print_exception(t, v, tb, None, sys.stderr)
  File "/usr/lib/python3.10/traceback.py", line 121, in print_exception
    print(line, file=file, end="")
  File "/work/dlclarge1/purucker-amlb/amlb-fork-2/amlb/logger.py", line 102, in new_print
    print_logger.log(level, buf.getvalue())


[... repeated in various forms and orders ...]
```

LennartPurucker:
I added a second file handler to the logging with the following code:

```python
    if log_file:
        # create file handler
        app_handler = logging.FileHandler(log_file, mode='a')
        app_handler.setLevel(app_level)
        app_handler.setFormatter(file_formatter)
        app_logger.addHandler(app_handler)

        log_dir, log_filename = os.path.split(log_file)
        name, ext = os.path.splitext(log_filename)
        second_log_file = os.path.join(log_dir, f"{name}_framework{ext}")
        framework_handler = logging.FileHandler(second_log_file, mode='a')
        framework_handler.setLevel(app_level)
        framework_handler.setFormatter(file_formatter)
        frameworks_logger.addHandler(framework_handler)
```

As far as I can tell, the bug still occurs in my latest runs with the same error as above. Is this the correct way to log to two files?

Would I need to change the `console = logging.StreamHandler()` or `root_handler = logging.FileHandler(root_file, mode='a')` as well? Otherwise, the worst workaround would be to set `print_to_log=False`, I guess.

PGijsbers:
Looking at this again, I think my previous hypothesis about writing from two processes is wrong. From the setup code you may think that the subprocess and the main process both log to the same file, but reality is that the subprocess never initializes the logger this way, so the `frameworks` logger is really only for `frameworks.shared.caller`. In practice, the [default python logging is used which writes to stdout](https://github.com/openml/automlbenchmark/blob/5399d802015693727285cc06a480e26128fef039/frameworks/shared/utils.py#L21) as a byproduct of importing the utils module. This stream output is captured in the communication function of the main process and written to log file from there.

It would be very helpful if you could disable `print_to_log` and hopefully the produced error message will be more meaningful :/

LennartPurucker:
Running my next batch with `print_to_log=False` and see what happens (and if I can find the error in my log output)

PGijsbers:
Interested to hear the results. I don't really see the advantage of `print_to_log`, it seems "necessary" only because the communicate function uses `print` instead of writing to a logger. I might try rewriting the logging setup altogether over the next few days, provided I can find the time.

LennartPurucker:
Good news, `print_to_log=False` prevents AMLB from crashing; that is, the logging error occurs, but this does not kill the run anymore.

Bad news, this does not explain the root cause either. I attached an entire log file, but the problem still seems to be a `Stale file handle`, so the log file does not exist anymore?  The logging error seems to happen before the run starts, but only because the order might be incorrect. 
[slurm-13929043_63.txt](https://github.com/user-attachments/files/18866804/slurm-13929043_63.txt)

I noticed that log files on the disk are, at some point, zipped by AMLB. Moreover, after zipping, the log directory is emptied. Could it be that the code that zips the log files is somehow triggered for a run with the same session ID before all runs have finished? In the local mode, my runs often use the same session ID (and log dir) because they are started in parallel.

For example, I have a run like `autogluondefault14022025.openml_t_2073.4h8c.local.20250215T125210`, saved in a folder with the same name. In the `./predictions/task_name/` folder, I have the predictions for three folds. At the same time, the `./logs` folder is empty, but the `logs.zip` only contains two log files.

PGijsbers:
> Could it be that the code that zips the log files is somehow triggered for a run with the same session ID before all runs have finished?

I don't know from the top of my head, but that would certainly explain a stale file handle. I'll look into that next

PGijsbers:
I think it is exactly what you say it is. The session id is not unique enough to allow separate parallel invocations of the same framework on the same benchmark with the same constraint to be unique. The `runbenchmark.py` script assumes it is though, since it cleans up the log directory after zipping in [here](https://github.com/openml/automlbenchmark/blob/5399d802015693727285cc06a480e26128fef039/runbenchmark.py#L395). 

It is possible to generate this with something like instantiating the same thing 20 times, though not reliably for me:
```
 python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip &python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip &python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip &python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip &python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip &python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip &python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip & python runbenchmark.py flaml -s skip &python runbenchmark.py flaml -s skip &
```

Simply commenting out the line linked above probably should solve the issue. Though you have logs eating disk space.

Possible solutions:

 - Detect when a benchmark is started with a session ID for which a directory is already created, and then:
    - Augmented to session ID, or
    - Error out, or
    - Work with unique subdirectories
 - Make session IDs unique by design, such as using UUIDs. 
 - Make session IDs semi-unique by design, such as introducing task and fold. Still has errors if the exact same experiment is run twice in parallel. Though here too you could detect it and error out.
 - Rely on the user to define unique session ids (not a fan).
 
 Any fix that changes the way session IDs are generated or artifacts are stored will affect scripts which currently rely on that structure, though. 
That said, it is also very inconvenient that with the current structure multiple processes write to the same file. It makes it much harder to understand individual experiments. Also open to other suggestions.

LennartPurucker:
Nice! 

So, besides fixing this, there is another problem with session IDs/filesystem usage: I would like to somehow map an entry in the .csv file to the dir where I can find its log/predictions/meta files. Right now, I need to check all files to determine this mapping. 

UUIDs would help with that the most. That is, use a UUID for the session ID, with the UUID then also being stored as part of the `results.csv`. 

This would break all existing scripts, indeed. So, I am not sure we want to go this route. Commenting out the lines above might be the best idea, as, IMO, the log files are, anyway, not that large/expensive right now, and we could provide a post-hoc script to zip all log files. 

I will comment out the code for now in my fork.

PGijsbers:
I changed the default to not archive log files. I feel that's the most reasonable solution for now.

LennartPurucker:
Agree! Great!

LennartPurucker:
I just noticed that there might be one more (smallish) problem: there is a race condition for https://github.com/openml/automlbenchmark/blob/master/amlb/benchmark.py#L473, which can make the "local" results broken. The lock for the global results makes sure the results file is correct. 

Unless the lock breaks due to Linux (as it just did for me today :D)

PGijsbers:
Breaks how? Can you create a new issue? Thanks!

PR #679, opened 2024-12-14
Remove TODOs from the code base, use issue tracker instead

PGijsbers:


amlb/runners/aws.py
```diff
@@ -1621,52 +1618,3 @@ def _ec2_startup_script(self, instance_key, script_params="", timeout_secs=-1):
             if timeout_secs > 0
             else rconfig().aws.max_timeout_seconds,
         )
-
-
-class AWSRemoteBenchmark(Benchmark):
```
PGijsbers, line 1626: I think the idea is to create an AWS benchmark which saves results during execution to avoid losing data when a task errors when you run multiple tasks? Not entirely sure what Seb meant. Running multiple tasks is assumed to be sequential (in parallel we can't guarantee fair resource usage), which means also extra cleanup between runs and so on. There hasn't been a request for this in the 6 years since it was written, i'm not 100% sure what is meant, so this is not converted to an issue,

amlb/runners/container.py
```diff
@@ -136,7 +135,6 @@ def _run():
                     else rconfig().seed,
                 )
             )
-            # TODO: would be nice to reload generated scores and return them
```
PGijsbers, line 139: While that would follow the other job results, results currently get printed and saved to logs fine? So I am not 100% sure what the added value is. Might be missing something, didn't have a close look.

PR #683, opened 2024-12-22
Enh/error on timeout

### LINKED ISSUE #681: Program continues after setup fails
In docker mode, I have encountered scenarios where building the image failed, and the benchmark tries to continue invoking the run command. I suspect that it is because `config.yaml:setup.activity_timeout` seconds elapse during setup without any output, the setup is aborted and this is not handled properly, but it needs some more investigation.

Desired outcome: benchmark stops providing a clear error message that the experiment is aborted due to inactivity in the setup.

PGijsbers:
Closes #681

If a process "completes" due to exceeding the activity timeout, it should raise an error, and not just continue.

### DIFF
--- a/amlb/utils/process.py
+++ b/amlb/utils/process.py
@@ -50,15 +50,7 @@ def file_lock(path, timeout=-1):
 
 
 class StaleProcessError(subprocess.TimeoutExpired):
-    def __init__(self, cmd, timeout, output=None, stderr=None, setting=None):
-        super().__init__(cmd, timeout, output, stderr)
-        self.setting = setting
-
-    def __str__(self):
-        return (
-            f"Process '{self.cmd}' was aborted after producing no output for {self.timeout} seconds. "
-            f"If the process is expected to take more time, please raise the '{self.setting}' limit."
-        )
+    pass
 
 
 def run_subprocess(
@@ -119,14 +111,11 @@ def run_subprocess(
 
         retcode = process.poll()
         if retcode is None:
-            # Process still lives: communication stopped because of activity timeout
+            # Process still lives => communication stopped because of activity timeout
             process.kill()
             process.wait()
             raise StaleProcessError(
-                process.args,
-                float("nan"),
-                output=stdout,
-                stderr=stderr,
+                process.args, float("nan"), output=stdout, stderr=stderr
             )
 
         if check and retcode:
--- a/amlb/utils/process.py
+++ b/amlb/utils/process.py
@@ -109,9 +109,13 @@ def run_subprocess(
             process.kill()
             raise
 
+        # Communication is aborted if either
+        #  - the process has completed, or
+        #  - the process does not produce any output in the specified timeout
+        # `communicate` can't distinguish those cases. If the process has a
+        # return code then it stopped already, otherwise we stop it now.
         retcode = process.poll()
         if retcode is None:
-            # Process still lives => communication stopped because of activity timeout
             process.kill()
             process.wait()
             raise StaleProcessError(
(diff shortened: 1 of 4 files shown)


amlb/utils/process.py
```diff
@@ -104,7 +116,19 @@ def communicate(process, input=None, timeout=None):
         except:  # also handles kb interrupts
             process.kill()
             raise
+
         retcode = process.poll()
+        if retcode is None:
```
PGijsbers, line 121: As far as I can tell, this is a reliable way to tell the process is still running. And if the process is still running at this point I think the only reason can be that the communicate function returned early, which should only happen with an activity timeout.

amlb/utils/process.py
```diff
@@ -186,6 +210,9 @@ def read_pipe(pipe, timeout):
         pipes = as_list(pipe)
         # wait until a pipe is ready for reading, non-Windows only.
         ready, *_ = select.select(pipes, [], [], timeout)
+        # if a pipe is not ready it could be timeout or it could be end of process
+        # so at this point we do not know. Only after the communicate function is over do we know.
+        # i.e., if the process is still running it does not have a retcode.
```
PGijsbers, line 215: Preferably I would have raised the activity timeout from the function which uses it. But at this stage we can unfortunately not detect whether the error should be raised.

runbenchmark.py
```diff
@@ -358,9 +360,19 @@
             args.mode,
         )
 
-    bench.setup(amlb.SetupMode[args.setup])
-    if args.setup != "only":
-        res = bench.run(args.task, args.fold)
+    try:
+        bench.setup(amlb.SetupMode[args.setup])
+    except StaleProcessError as e:
```
PGijsbers, line 365: I'll move to exception notes instead and/or revise this structure. The reason I did it this way is to communicate more clearly to the user with a final message what went wrong and how to solve it. I want to generally make errors easier to parse, as there are some issues opened that are completely solvable from the traceback, but users can't/don't try to parse those.

amlb/utils/process.py
shchur: Hi @PGijsbers, I am not very familiar with this section of the code base but the logic looks reasonable to me. 

Do I understand correctly that after the previous try/except block has been completed, the process should not be running, and this code block ensures that this is the case?

shchur (approved):

amlb/utils/process.py
PGijsbers: The `stdout, stderr = communicate(process, input, timeout=timeout)` line is blocking until no stderr and stdout is detected. A lack of output means that either there was an activity timeout (nothing written during the specified interval), or the process stopped. The communication logic doesn't detect which. 
That's why at this level, after communication has stopped, we can do an extra check (this one), to see if the process is alive. If the process is still alive we can infer that the communication was stopped because a timeout was detected. In that case, the subprocess wouldn't just stop by itself, so we send the kill signal and raise an error as a normal return from this function would indicate the process finished successfully.

PGijsbers: And also thanks so much for doing the review anyway :) I understand generally speaking people aren't too familiar with the internals (to be frank, I also had to brush up on some of it as Seb wrote this), I just appreciate a sanity check. Both for correctness and to avoid me refactoring/solving things in a way that only I will be able to understand later.

shchur: Thanks for the explanation, that makes sense. I would maybe add a small comment indicating the two possible reasons why `retcode is None` - at least for me that helpful to understand the purpose of this block.

PR #685, opened 2024-12-31
Error when frameworks installation script fails

PGijsbers:
If we detect an error during setup, we want to stop the benchmark process immediately, instead of continuing.
Currently, the failed process goes unnoticed and often results in unexpected errors like missing dependencies during the execution of the `exec.py` script.

### DIFF
--- a/amlb/utils/process.py
+++ b/amlb/utils/process.py
@@ -245,7 +245,6 @@ def run_cmd(cmd, *args, **kwargs):
         activity_timeout=None,
         log_level=logging.INFO,
         monitor=None,
-        exit_immediately=False,
     )
     for k, v in params:
         kk = "_" + k + "_"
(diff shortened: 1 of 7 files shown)


amlb/utils/process.py
```diff
@@ -252,7 +252,7 @@ def run_cmd(cmd, *args, **kwargs):
             params[k] = kwargs[kk]
             del kwargs[kk]
     cmd_args = as_cmd_args(*args, **kwargs)
-    full_cmd = flatten([cmd]) + cmd_args
+    full_cmd = [os.environ["SHELL"], "-e"] + flatten([cmd]) + cmd_args
```
PGijsbers, line 255: This feels stupid, but it's the only way I could find to make the `-e` setting be in effect by default for the invoked command (which we want for the setup).

eddiebergman (commented):
It seems it would be easier to detect this at the level of the subprocess call, rather than injecting a `-e`, this in effect offloads to the responsibility to the creator of the script to exit with the correct status code.

This would require a template that includes and documents a `-e` call in there, so people are aware of it if not super familiar with writing bash scripts. People writing a script for Windows, of which I imagine there are not too many, would have to do so in the way expected of a `.bat` file or whatever it is.

eddiebergman:
It's not too much of an issue in the end, just need to ensure that using `set -e` twice isn't an issue.

I'm usually for offloading these issues to the user, namely just because it's hidden behaviour and might end up with their own manual running being different than what the benchmark does. For example, the `git clone <src> <dest>` will error if `<dest>` exists, but this is not always an issue for the setup, as you might still want to checkout a certain version.

PGijsbers:
Related to #681

PR #686, opened 2025-01-02
Tiny bug : oml_id must be an int in load_openml_task

SubhadityaMukherjee:
Check if oml_id is a valid integer before trying to load the task. This sometimes does not happen by default. (Only get_suite supports string ids.)

I am not sure if this behavior is different in the new OpenML API.

### DIFF
--- a/amlb/benchmarks/file.py
+++ b/amlb/benchmarks/file.py
@@ -35,4 +35,9 @@ def load_file_benchmark(
     log.info("Loading benchmark definitions from %s.", benchmark_file)
     tasks = config_load(benchmark_file)
     benchmark_name, _ = os.path.splitext(os.path.basename(benchmark_file))
+    for task in tasks:
+        if task["openml_task_id"] is not None and not isinstance(
+            task["openml_task_id"], int
+        ):
+            raise TypeError("OpenML task id for task {task.name!r} must be integer.")
     return benchmark_name, benchmark_file, tasks
--- a/amlb/benchmarks/openml.py
+++ b/amlb/benchmarks/openml.py
@@ -66,10 +66,15 @@ def load_openml_tasks_from_suite(domain: str, oml_id: str) -> list[Namespace]:
     return tasks
 
 
-def load_openml_task(domain: str, oml_id: str) -> list[Namespace]:
+def load_openml_task(domain: str, oml_id: int) -> list[Namespace]:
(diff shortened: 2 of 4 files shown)


amlb/benchmarks/openml.py
```diff
@@ -20,7 +21,13 @@ def is_openml_benchmark(benchmark: str) -> bool:
...
 def load_oml_benchmark(benchmark: str) -> tuple[str, str | None, list[Namespace]]:
     """Loads benchmark defined by openml suite or task, from openml/s/X or openml/t/Y."""
-    domain, oml_type, oml_id = benchmark.split("/")
+    domain, oml_type, oml_id_str = benchmark.split("/")
+    try:
+        oml_id = int(oml_id_str)
+    except ValueError:
+        raise ValueError(
+            f"Could not convert OpenML id {oml_id_str!r} in {benchmark!r} to integer."
+        )
```
PGijsbers, line 30: The reason it was sometimes `str` and sometimes not was because when specified as an openml benchmark (openml/{s,t}/{id} it was passed as str. But when loaded from a file the ids were conventionally defined as integers.

amlb/benchmarks/file.py
```diff
@@ -35,4 +35,9 @@ def load_file_benchmark(
     log.info("Loading benchmark definitions from %s.", benchmark_file)
     tasks = config_load(benchmark_file)
     benchmark_name, _ = os.path.splitext(os.path.basename(benchmark_file))
+    for task in tasks:
+        if task["openml_task_id"] is not None and not isinstance(
+            task["openml_task_id"], int
+        ):
+            raise TypeError("OpenML task id for task {task.name!r} must be integer.")
```
PGijsbers, line 42: While tasks defined in a yaml file _should_ be integer. Nothing stops a user from defining a str instead. So here we can give an early warning.

PGijsbers (approved):
Thanks for the pointer. However, I wanted to understand _why_ sometimes there was an integer and sometimes a string. Modified/moved the logic a little bit to catch these mistakes earlier in the process to give a faster and more informative error message.

btw. you are using a "too new" version of OpenML. The one defined in `requirements.txt` still supports identifiers passed as a str, which is why this wasn't raised before :) but it's good future proofing anyway, I want to upgrade someday.

PR #688, opened 2025-02-11
noninteractive and sudo fix

SubhadityaMukherjee:
Tiny changes since the previous version did not run successfully. 
sudo wasnt needed for the build
and the noninteractive frontend was not exported correctly.


PGijsbers (approved):
thanks!

PR #692, opened 2025-02-27
Load data should download data to disk

PGijsbers:
This should fix the issue with arff files not being downloaded in time to generate data splits.


SubhadityaMukherjee (approved):
This works as expected :)
I cleared my cache and tested it again.

Issue #661, opened 2024-11-29
AutoGluon v1.2 fails


PGijsbers:
When AutoGluon resorts to pickle files to transfer data between benchmark and framework process, the benchmark fails because there is an incompatible numpy version installed in the framework's virtual environment.

The basic autogluon tests on #656 succeeded, but failed on #657. The latter obviously should have no effect on the functioning of the integration, but now there is a [`No module named numpy._core`](https://github.com/openml/automlbenchmark/actions/runs/12057168971/job/33621215442#step:7:890) error message. However, from the logs we see that the former used AutoGluon 1.1.1, whereas the latter has AutoGluon 1.2.

PGijsbers:
It looks like AutoGluon ends up installing a more modern version of numpy ([logs](https://github.com/PGijsbers/automlbenchmark/actions/runs/12282659124/job/34274368429#step:7:415)). Though from what I understood the introduction of `_core` was in 2.0? Anyway, the shared requirements are violated during the installation of AG which results in this error.

PGijsbers:
I think the long-term solution is probably to avoid using pickles and instead use parquet for everything (except we have to figure out what to do with sparse data).

deadsoul44:
I was able to run a task with AutoGluon 1.2 on 2024-11-30. But I get this error now. This is probably due to the new release of numpy.

deadsoul44:
Tried numpy 2.0.0, still fails with the same message.

Innixma:
Thanks for highlighting! This is actually also a bug in AutoGluon to an extent, because CatBoost doesn't support numpy 2.x yet. The problem is that we are installing scikit-learn-intelex as a separate install step in setup.sh: https://github.com/openml/automlbenchmark/blob/master/frameworks/AutoGluon/setup.sh#L49

Because the latest skex uses numpy 2, it force installs numpy 2, which breaks CatBoost. So this would harm AutoGluon's predictive quality quite a bit. It might have even harmed my latest benchmarks, I didn't look closely because we only added numpy 2.0 support very close to the release. To fix this we would want to include the `skex` install in the same line as the rest of the install, rather than doing it after. This would then align with the user install experience of skex.

Once I'm back from NeurIPS I'll try to take a look at this and see if I can fix.
[...]

PR #696, opened 2025-03-02
Drastically reduce the amount of frameworks used in common check

PGijsbers:
To improve the consistency of CI, for now we focus on just three reliable AutoML frameworks.


PGijsbers:
Also closes #661

